PR #552, opened 2014-12-26
warn upon pin reuse; closes #548

### LINKED ISSUE #548: question: Pin reuse
If I do this:

```js
var rgb = new Led.RGB(9, 10 11);
```

Then use the `rgb` object as in the [rainbow example](https://github.com/rwaldron/johnny-five/blob/master/eg/led-rainbow.js), everything works great.

However, if I do this:

```js
var rgb = new Led.RGB(9, 10, 11);
var red = new Pin(9), green = new Pin(10), blue = new Pin(11);
```

Then, the exact same "rainbow" code does nothing; the LED never turns on. 

If I had to hazard a guess, instantiating the `Pin`s after the `Led.RGB` instance causes the `Pin`s to be put in the incorrect mode--but, of course, not sure, so that's why I'm asking (my other guess would be that this is a `firmata` problem).

Any workarounds for this?  I'm just experimenting here.

* * *

Also, doing this:

```js
var rgb = new Led.RGB(9, 10, 11);
var red = new Pin({addr: 9, type: 'analog'});
```

causes the following exception:

```
    this.pins[this.analogPins[pin]].report = value;
                                           ^
>> TypeError: Cannot set property 'report' of undefined
    at Board.reportAnalogPin (/Volumes/samara/projects/boneskull/sparkfun/node_modules/johnny-five/node_modules/firmata/lib/firmata.js:1214:44)
    at Board.analogRead (/Volumes/samara/projects/boneskull/sparkfun/node_modules/johnny-five/node_modules/firmata/lib/firmata.js:567:8)
    at new Pin (/Volumes/samara/projects/boneskull/sparkfun/node_modules/johnny-five/lib/pin.js:122:25)
```

This looks like "user error", but it should probably be more friendly.

boneskull:
- `Board#register` is now an object, properly indexed by address.
- `Board#register` is of type `Object<String,Array>`, to potentially store multiple objects against the same address
- `Board#serialize` will retain the same behavior if only one object is used for an address; otherwise it calls `reduce()` for each

I couldn't figure out where to put tests for this.


rwaldron:
Alternate approach here: https://github.com/rwaldron/johnny-five/commit/1b1530e225246650fdc366ea3e744908cac5bf07

PR #578, opened 2015-01-14
Fix Servo.to

deathbearbrown:
Issue:
When calling Servo.to with the same value for degrees twice in a row, the servo wouldn't fail.  Instead it went to 0 degrees.

### DIFF
--- a/test/servo.js
+++ b/test/servo.js
@@ -180,6 +180,29 @@ exports["Servo"] = {
     test.done();
   },
 
+  degreeChange: function(test) {
+    test.expect(2);
+
+    this.servo = new Servo({
+      pin: 11,
+      board: board
+    });
+    // We are spying on setInterval instead of servoWrite because the following set of
+    // parameters will result with setInterval will be called with an interval of infinity
+    var spy = sinon.spy(global, "setInterval");
+
+    this.servo.to(180);
+
+    this.servo.to(180, 1000, 100);
+
+    this.clock.tick(1010);
+
+    test.equal(spy.callCount, 0);
+    test.ok(this.servoWrite.callCount,1);
+
+    test.done();
+  },
+
   fps: function(test) {
     test.expect(1);
 
--- a/lib/servo.js
+++ b/lib/servo.js
@@ -136,7 +136,7 @@ function Servo(opts) {
   this.offset = opts.offset || 0;
   this.mode = this.io.MODES.SERVO;
   this.interval = null;
-  this.value = 0;
+  this.value = null;
 
   /**
    * Used for adding special controllers (i.e. PCA9685)
@@ -281,6 +281,11 @@ Servo.prototype.to = function(degrees, time, rate) {
     last = this.last && this.last.degrees || 0;
     distance = Math.abs(last - degrees);
     percent = 0;
+
+    if (distance === 0) {
+      process.nextTick(this.emit.bind(this, "move:complete"));
+      return this;
+    }
 
     // If steps are limited by Servo resolution
     if (distance < rate) {


rwaldron:
LGTM, when @dtex is +1 this can land

dtex:
I only looked at this briefly, but two things come to mind:

First,
I think this is duplicating something that is supposed to be happening already in the servowrite methods.

````
...
// If same degrees, emit "move:complete" and return immediately.
if (this.last && this.last.degrees === degrees) {
  //process.nextTick(this.emit.bind(this, "move:complete"));
  this.emit("move:complete");
  return this;
}
...
````
Maybe I screwed up and those are not working properly?

Second,
```this.value``` is kinda weird in that it shows the last requested value, not necessarily the current value of the servo. I'm not sure if that's a problem though. It kinda seems like it would still work even if a user had passed a long duration on previous request of the same or a different value (edge case yes, but I'm just thinking out loud).

There is also ```this.last.target``` and ```this.last.degrees``` the latter of which is a duplicate of ```this.value```. ```this.value``` currently shows the the last requested position when it should probably be the current position (this is for servo.to() calls with a duration parameter).

@rwaldron , can you tell us where this.value might be used? I don't see it in the wiki, it predates me and I didn't address it when I worked on Servo(). If it's for users, I'm thinking it should be a getter for ```this.last.target``` and we should remove the duplicate value.

**Comment by deathbearbrown** [comment]
Hi Donovan!
If it helps troubleshoot this only happens when I pass in more than just the degree argument. 

I was playing with my robot arm today and it almost smacked me in the head.

**Comment by dtex** [comment]
> I was playing with my robot arm today and it almost smacked me in the head.

Note to self: check license to ensure indemnification re: getting smacked in the head.

@deathbearbrown I copied over just your new test/servo.js file and even without your new lib/servo.js it passed, so the duplicate value catcher in servowrite was already working. I think something else is going on.

When you say it went to 0 degrees were you spying on the value, or is that just where the arm appeared to go? If the latter, I bet it was actually ```nan```. Now, to figure out how that might have happened...

Would you mind removing your added lines in lib/servo.js and testing again? I'm sure I'm overlooking something, but right now I'm befuddled and don't understand how the new code made a difference.

**Comment by deathbearbrown** [comment]
Spying on the value and that's where it went! When I get into the office I will make you a vine.

**Comment by dtex** [comment]
Okay. I'll try to recreate on my end as well.

File: test/servo.js [hunk 1]
```diff
@@ -180,6 +180,27 @@ exports["Servo"] = {
...
+      pin: 11,
+      board: board
+    });
+
+    this.servo.to(0);
+
+    this.servo.to(180, 1000, 100);
+    this.servo.to(180, 1000, 100);
+
+    this.clock.tick(1010);
```
dtex: This test won't catch the error. Both Servo.to() calls fire on the same tick when the distance is still 180. The test needs to wait for the first one to complete before calling the second, or calling servo.to(180) and then servo.to(180, 1000)

deathbearbrown: Oh ok, I'll fix that.

rwaldron:
> I think this is duplicating something that is supposed to be happening already in the servowrite methods.

The problem is that this is even getting to the `servoWrite` calls. If a program calls `servo.to(180); servo.to(180, 500); servo.to(180, 1000, 100);` then only the first of those should ever get a message through to `servoWrite`. One thing we discovered as well is that `this.value` is initialized to `0`, which is wrong (and also [my fault](https://github.com/rwaldron/johnny-five/commit/c6cc6ae95ce1329a9fcb19f46ca441b023ab8cca#diff-1)), since a newly initialized servo can't possibly know that the actual mechanism is at 0°—so `null` makes more sense. Once that's fixed, then the check can be further simplified.

test/servo.js
rwaldron: > The test needs to wait for the first one to complete before calling the second, 

In this case there is no need to wait, since any `servo.to(X)` call that follows, where `X` is the same value, should just: 

1. queue an async `this.emit("move:complete")`
2. return immediately

rwaldron:
> this.value currently shows the the last requested position when it should probably be the current position (this is for servo.to() calls with a duration parameter).

`this.value` is supposed to be "what the user last set the device to"

dtex:
The reason I defer the check to servoWrite is that not all controllers are limited to 8-bit resolution like the Arduino. The checks will be different.

deathbearbrown:
OK so this was fun. The test is fixed :sparkles: 
see this comment:
>We are spying on setInterval instead of servoWrite because the following set of
>parameters will result with setInterval will be called with an interval of infinity

dtex:
LGTM :shipit: 

Thanks for your patience and sleuthing.

rwaldron:
@dtex awesome, thanks!

deathbearbrown:
woooo!

PR #581, opened 2015-01-15
Test for "move:complete"

deathbearbrown:
Test for calling emit "move:complete" in the case where you call the same degree value twice.
Sorry the last PR got merged faster than I could add this. 

For testing this line:
https://github.com/rwaldron/johnny-five/blob/master/lib/servo.js#L286

### DIFF
--- a/test/servo.js
+++ b/test/servo.js
@@ -180,6 +180,23 @@ exports["Servo"] = {
     test.done();
   },
 
+  completeMoveEmmitted: function(test) {
+    test.expect(1);
+
+    this.servo = new Servo({
+      pin: 11,
+      board: board
+    });
+
+    this.servo.to(180);
+    this.servo.to(180, 1000, 100);
+    this.servo.on("move:complete", function() {
+      test.ok(this.servoWrite.callCount, 1);
+      test.done();
+    }.bind(this));
+  },
+
+
   degreeChange: function(test) {
     test.expect(2);
 
@@ -198,7 +215,7 @@ exports["Servo"] = {
     this.clock.tick(1010);
 
     test.equal(spy.callCount, 0);
-    test.ok(this.servoWrite.callCount,1);
+    test.ok(this.servoWrite.callCount, 1);
 
     test.done();
   },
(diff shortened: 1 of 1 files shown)


dtex:
This test fails in an unhelpful way. It will eventually time out and return ```Fatal error: Uncaught, unspecified "error" event.``` I wonder if there is a better way?

rwaldron:
```
$ grunt
Running "jshint:files" (jshint) task
>> 292 files lint free.

Running "jscs:files" (jscs) task
Missing space before opening curly brace for block expressions at test/servo.js :
   191 |    this.servo.to(180);
   192 |    this.servo.to(180, 1000, 100);
   193 |    this.servo.on("move:complete", function(){
----------------------------------------------------^
   194 |      test.ok(this.servoWrite.callCount,1);
   195 |      test.done();
>> 1 code style errors found!
Warning: Task "jscs:files" failed. Use --force to continue.

Aborted due to warnings.
```

rwaldron:
@dtex when the event isn't emitted? I've been trying to track that error event down and I can't find it. I also can't figure out how to intercept the timeout.

deathbearbrown:
I can't figure out how to test this right.  

@rwaldron suggested that I should wrap the event in process.nextTick(), but if I remove it & put a console.log() inside the .on('move:Complete') in the test, it never gets in there.

rwaldron:
This is maybe a dumb question: were you using "move:complete" or "move:Complete"?

deathbearbrown:
Not a dumb question. It's just 'move:complete'  I made a typo.

rwaldron:
Ok cool (that's what I figured, but never hurts to ask)

rwaldron:
@dtex neither of us can figure out how to intercept that error. The only consequence of accepting this patch without a nicer error is that it's hard to figure out what happened if `process.nextTick(this.emit.bind(this, "move:complete"));` is ever removed—which it shouldn't be anyway. I'm willing to accept this consequence.

dtex:
Agreed. It should never happen and the test will fail if it does which is the most important thing. Plus it'll be pretty easy to track down if it ever does come up.

PR #582, opened 2015-01-16
Update LED docs based upon feedback from the latest Nodebots NYC.

reconbot:
- Remove LED-Digital docs as they are too similar to other docs
 - Remove the LED command line argument for choosing a pin, it added complexity to a simple example. About half of our newbies had little to none cli experience.
 - Update the LED-on-off diagram to include a resistor as to not destroy LEDs


rwaldron:
> Remove LED-Digital docs as they are too similar to other docs

Cool. 

> Remove the LED command line argument for choosing a pin, it added complexity to a simple example. About half of our newbies had little to none cli experience.

Instead of changing this example, because I use this all the time (at least once a day) for testing, can you make a new one called "led-blink.js" or similar?



> Update the LED-on-off diagram to include a resistor as to not destroy LEDs

I rarely see this happen in practice. 

The resistor example is unnecessary and probably "looks too complicated" for new users. I would accept a diagram that included _both_ the original and the resisted circuit.

PR #583, opened 2015-01-16
Add grunt test-docs to ensure docs are up to date

reconbot:
- Not added to the default task as it relies on your changes to be committed and causes the docs to be generated. This is fine on our CI but not ok in local development.
 - Add to circle as a before_script because of the previous point
 - ir-distance.md was out of date

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,16 +119,17 @@
   "devDependencies": {
     "async": "~0.9.0",
     "copy-paste": "~0.3.0",
-    "optimist": "~0.6.1",
-    "keypress": "latest",
-    "sinon": "~1.10.2",
     "grunt": "~0.4.5",
     "grunt-cli": "~0.1.13",
     "grunt-contrib-jshint": "~0.10.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -119,16 +119,17 @@
...
     "grunt-cli": "~0.1.13",
     "grunt-contrib-jshint": "~0.10.0",
     "grunt-contrib-nodeunit": "~0.4.1",
     "grunt-contrib-watch": "~0.6.1",
     "grunt-jsbeautifier": "~0.2.2",
-    "grunt-jscs": "~0.8.0"
+    "grunt-jscs": "~0.8.0",
+    "keypress": "latest",
+    "optimist": "~0.6.1",
+    "shelljs": "^0.3.0",
```
rwaldron: ^ -> ~

rwaldron:
Mostly irrelevant, but I hate that this is called "docs" and lives in a directory called "docs" because they are not "docs". This is my fault :(

PR #586, opened 2015-01-20
Adding PCA9685 controller to Led

frxnz:
Adds support for the [Adafruit PWM Driver](https://www.adafruit.com/products/815) breakout and shield.

Also refactored `Led` to follow the Controller model.


test/led.js
```diff
@@ -487,6 +489,297 @@ exports["Led - PWM (Analog)"] = {
...
+
+  brightness: function(test) {
+    test.expect(2);
+
+    this.led.off();
+    this.led.brightness(255);
+    test.ok(this.spy.calledWith(64, [6, 0, 0, 4095, 15]));
+
+    // this.led.brightness(100);
+    // test.ok(this.spy.calledWith(64, [6, 0, 0, 4095 * 100 / 255, 15]));
```
frxnz: I was having trouble getting this to pass. The values being passed are correct, but the test fails. I'm wondering if there's a rounding issue.

frxnz: Derp, nevermind. The last element of the array is wrong.

eg/led-PCA9685.js
```diff
@@ -0,0 +1,24 @@
...
+five.Board().on("ready", function() {
+
+  var ledPin = process.argv[2] || 0;
+  var led = new five.Led({
+    address: 0x40,
+    pin: ledPin,
+    controller: "PCA9685"
+  });
+
+  // address: The address of the shield.
```
frxnz: This all needs to be cleaned up

frxnz:
@rwaldron Do I need to regenerate docs or does that get done on merge?

lib/led.js
```diff
@@ -23,27 +143,36 @@ function Led(opts) {
...
+  }
+  else {
+    controller = opts.controller;
+  }
+
+  if (controller == null) {
+    controller = Controllers["Standard"];
+  }
+
+  Object.defineProperties(this, controller);
```
dtex: This seems unnecessary since opts alway exists either as a number or an object. If opts does not exist we don't have a pin number so the instance is doomed anyway.

Are their reasons for doing it this way that I don't see?

It's handled like this in Servo and Motor:
````
if (this.opts.controller) {
  controller = typeof this.opts.controller === "string" ?
    Controllers[this.opts.controller] : this.opts.controller;

  Object.defineProperties(this, controller);
}
````

Is that something we should change?

**Comment by dtex** [comment]
I can't speak to LED in general, but the Controller implementation looks spot on (save my one question above). @Resseguie and @rwaldron know this class much better than I so I'm sure their input will be forthcoming.

**Comment by dtex** [comment]
Oh, and the commits should be squashed. There is so much going on with J5 that having new functionality spread out across multiple commits makes it hard to keep track.

File: lib/led.js (hunk 3 above)
Comment (frxnz) [inline lib/led.js]: @dtex `Led` can be initialized with no parameters, since there is a `defaultLed` hardcoded to pin 13. I can definitely see the reasoning behind this, since Arduino has the LED on pin 13, and it makes the "hello world" dead simple.

Comment (dtex) [inline lib/led.js]: See, I definitely can't speak to LED :-)

Would it be simpler have a default opts object and then decorate that with the user defined opts? That would simplify this and the defaultLed stuff.

Comment (frxnz) [inline lib/led.js]: Yep, that makes a lot of sense. There might still need to be some type checking, but I think having a default object would be more intuitive/readable.

File: eg/led-PCA9685.js (hunk 2 above)
Comment (rwaldron) [inline eg/led-PCA9685.js]: Since `0x40` is the default, you can remove the property from this constructor opts

File: eg/led-rgb-PCA9685.js [hunk 4]
```diff
@@ -0,0 +1,28 @@
...
+
+
+five.Board().on("ready", function() {
+
+  // Initialize the RGB LED
+  var led = new five.Led.RGB({
+    pins: [0, 1, 2],
+    address: 0x40,
+    controller: "PCA9685"
+  });
```
rwaldron: Oh rad, I like this alot :)

lib/led.js
```diff
@@ -1,9 +1,129 @@
...
+        this.pwmRange = opts.pwmRange || [0, 4095];
+
+        if (!this.board.Drivers[opts.address]) {
+          this.io.sendI2CConfig();
+          this.board.Drivers[opts.address] = {
+            initialized: false
+          };
+
+          // Reset
+          this.io.sendI2CWriteRequest(this.address, [this.COMMANDS.PCA9685_MODE1, 0x0]);
```
rwaldron: Update these i2c calls to the new API: https://github.com/jgautier/firmata#i2c (these should be `i2cWriteReg(address, reg, value)`)

lib/led.js
```diff
@@ -1,9 +1,129 @@
...
+      value: function() {
+
+        var on, off;
+        var state = priv.get(this);
+        var value = state.isAnode ? 255 - Board.constrain(value, 0, 255) : state.value;
+
+        on = 0;
+        off = this.pwmRange[1] * value / 255;
+
+        this.io.sendI2CWriteRequest(this.address, [this.COMMANDS.LED0_ON_L + 4 * (this.pin), on, on >> 8, off, off >> 8]);
```
rwaldron: `i2cWrite(address, [...bytes])`

lib/led.js
```diff
@@ -1,9 +1,129 @@
...
+      }
+    }
+  },
+  Standard: {
```
rwaldron: "Standard" => "DEFAULT"

lib/led.js
```diff
@@ -1,9 +1,129 @@
...
+          this.pin = typeof opts.pin === "undefined" ? defaultLed : opts.pin;
+          this.mode = this.io.MODES[
+            (opts.type && opts.type.toUpperCase()) ||
+            (this.board.pins.isPwm(this.pin) ? "PWM" : "OUTPUT")
+          ];
+        }
+
+      }
+    },
+    ledWrite: {
```
rwaldron: just call this `write`

test/led.js
```diff
@@ -361,14 +361,16 @@ exports["Led - PWM (Analog)"] = {
...
 
     this.led.mode = 1;
     this.led.fade();
+    this.clock.tick(1001);
     test.equal(this.led.mode, 3);
 
     this.led.strobe();
-    // pre 0.8.13, this would change to 1
+    this.clock.tick(1001);
     test.equal(this.led.mode, 3);
```
rwaldron: I don't see why these tests needed clock ticks? They don't test anything that is time-bound, the mode will be updated synchronously.

lib/led.js
rwaldron: None of this is necessary if you move the relevant parts of this: 

```js
  if (!opts || !opts.hasOwnProperty("controller")) {
    controller = Controllers["Standard"];
  }
  else if (typeof opts.controller === "string") {
    controller = Controllers[opts.controller];
  }
  else {
    controller = opts.controller;
  }

  if (controller == null) {
    controller = Controllers["Standard"];
  }

  Object.defineProperties(this, controller);
```

To some point after this: 

```js
  Board.Component.call(
  );
```
 

If you look in accelerometer.js, it's correct: 

```js
  var controller = null;
  var err = null;
  
  // ...
  
  Board.Component.call(
  );

  if (opts.controller && typeof opts.controller === "string") {
    controller = Controllers[opts.controller.toUpperCase()];
  } else {
    controller = opts.controller;
  }

  if (controller == null) {
    controller = Controllers["ANALOG"];
  }

  Object.defineProperties(this, controller);
```

rwaldron:
Needs a breadboard diagram to accompany `eg/led-PCA9685.js` and `eg/led-rgb-PCA9685.js`. Adafruit has fritzing components for the breakout, which is sufficient

test/led.js
frxnz: @rwaldron Since the PWM check and `this.mode` reset is now in `ledWrite` (since it's controller-specific), the mode doesn't get reset until the first write has fired.

That said the default delay is actually 10ms, so they should be `this.clock.tick(10);`

rwaldron: I get it now. I'm not sure I like it, but I'm not going to block on that either.

frxnz: I hear you, I was thinking "this isn't the right way to do this" as I was
doing it. I'll give it some more thought and see if I can come up with a
better solution.

rwaldron: Cool, I appreciate that. 

I'm looking forward to this support landing :)

frxnz:
- [x] Updated i2c calls to the new API
- [x] Since 0x40 is the default, you can remove the property from this constructor opts
- [x] "Standard" => "DEFAULT"
- [x] `ledWrite` => `write`
- [x] `this.mode` is read only, eliminates need for repeated PWM checking
- [x] Added breadboard diagram to docs

rwaldron:
I'm soooooo excited for this. I just want to play around with it first, then I'll land it :)

rwaldron:
Landed https://github.com/rwaldron/johnny-five/commit/a662c55aedbc81dbb8d938cea0970056d4928c69

PR #588, opened 2015-01-23
Move debounce function to constructor

### LINKED ISSUE #587: Simultaneous button events get dropped
If two buttons are pressed or released simultaneously, one of the two button's events will not fire. It is easily reproduced with imp-io because we send pin states in batches but it happens with Firmata on the Uno as well... It's just harder to reproduce. With my ninja like reflexes I'm able to make it happen about 10% of the time on the Uno.

The code to reproduce is pretty straightforward
````
var five = require("johnny-five");

var board = new five.Board();

  var b1 = new five.Button(3);
  var b2 = new five.Button(11);
  b1.state = b2.state = "off";

  b1.on("press", function() {
    b1.state = "on";
    report("b1");
  });

  b1.on("up", function() {
    b1.state = "off";
    report("b1");
  });

  b2.on("press", function() {
    b2.state = "on";
    report("b2");
  });

  b2.on("up", function() {
    b2.state = "off";
    report("b2");
  });

  report = function(trigger) {
    console.log(trigger, b1.state, b2.state);
  }
});
````

Then just press and/or release the two buttons at EXACTLY the same time.

### PR DESCRIPTION
**Pull request by dtex**
Fixes #587 

Prevents “cross-debouncing” of events across button instances.

**Don't merge yet.** - I don't have hardware with me to test.

### DIFF
--- a/lib/button.js
+++ b/lib/button.js
@@ -9,15 +9,7 @@ var priv = new Map(),
   aliases = {
     down: ["down", "press", "tap", "impact", "hit"],
     up: ["up", "release"]
-  },
-  // Create a 5 ms debounce boundary on event triggers
-  // this avoids button events firing on
-  // press noise and false positives
-  trigger = __.debounce(function(key) {
-    aliases[key].forEach(function(type) {
-      this.emit(type, null);
-    }, this);
-  }, 7);
+  };
 
 
 /**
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/button.js [hunk 1]
```diff
@@ -89,6 +81,15 @@ function Button(opts) {
   // Button instance properties
   this.holdtime = opts && opts.holdtime || 500;
 
+  // Create a 5 ms debounce boundary on event triggers
+  // this avoids button events firing on
+  // press noise and false positives
+  this.trigger = __.debounce(function(key) {
```
rwaldron: `this.trigger` => `var trigger`

Issue #564, opened 2014-12-30
Accelerometer: support MMA7361


rwaldron:
http://www.amazon.com/s/ref=nb_sb_noss?url=search-alias%3Daps&field-keywords=mma7361

BrianGenisio:
I have one on order.  I'll knock it out when it gets here.

BrianGenisio:
This accelerometer is not like the others :)

This unit requires a calibration step to determine the zero-values.  Unlike other accelerometers, the X,Y,Z don't necessarily have the same zero values.  The [Arduino code](https://github.com/jeroendoggen/Arduino-MMA7361-library/blob/master/AcceleroMMA7361/AcceleroMMA7361.cpp) relies on the calibrated centers of the separate axis, and I am certainly observing that they are different on my device.

Because of this, we can't use the same pattern that we have with other analog accelerometers, where we have a `zeroV` which is pre-determined and used for all three axis.  

I have a few questions/brainstorms, looking for opinions:

1. We'll need a `zeroV[]` array now.  Do we just do this in this controller?  Or do we push this down to the generic layer, and update the other controllers to accommodate?  We could allow `Number` or `Array` and use the appropriate logic based on what the controller exposes... this is probably the best approach.

2. We need to include a `calibrate` function on `Accelerometer`.  I know @rwaldron doesn't like modifying the shape for a single controller, so this means that `calibrate()` needs to be part of accelerometer now?  If so, I recommend using the same pattern as the Reflectance array.  This includes the functions `calibrate()`, `calibrateUntil(predicate)`, `loadCalibration(calibrationData)`, and `calibration` property exposed.  For the units that don't require calibration, calling `calibrate()` would overwrite the defaults.  This is probably a good thing in case some units are off from center.  I'd document in the wiki that this particular device requires a calibration.  Does this seem reasonable?

3. This unit also has a `sleep` pin which can be set to low to deactivate the device.  Do we expose the functionality for them to `enable()` and `disable()` the accelerometer?  And what about having a consistent shape?  Most of the other devices don't have this capability, but it could be a no-op or an exception in that case.

BrianGenisio:
See submitted PR.  I took a stab at answering these questions. 

1. I am supporting `Number` or `Array` for the `zeroV` values.

[...]

PR #592, opened 2015-01-25
Mma7361

BrianGenisio:
This PR adds [support for the MMA-7361](https://github.com/rwaldron/johnny-five/issues/564).  This device is different from others in that the three axis have different center values which can vary from device, and it has a "sleep" pin that when set to low, puts the device in sleep mode.

- Support the MMA-7361 with optional `sleepPin`
- Allow `zeroV` to be either an `Array` or `Number`. Available to any device type.
- Support an `autoCalibrate` boolean which will use the first several reads to find the center values.  Available to any device type.
- Export the `zeroV` property on the device so that it can be read and remembered for later (used in init)
- Add an `enable()` and `disable()` function that will turn the device on or off (stop responding to data events).  If the MMA-7361 provides a `sleepPin`, set the pin low to put the device to sleep.

**TODO**: ~~Fritzing diagram~~, ~~unit tests~~, and Wiki documentation updates (after it lands)


lib/accelerometer.js
```diff
@@ -242,16 +305,38 @@ function Accelerometer(opts) {
     this.toGravity = opts.toGravity || function(raw) { return raw; };
   }
 
+  if (!this.onEnable) {
+    this.onEnable = function(on) {};
```
rwaldron: Make this `enable` with a `disable` counterpart.

rwaldron: Also, you can fake "enable"/"disable" for those that don't have hardware support by just returning from the dataHandler if state.enabled === false

BrianGenisio: One step ahead of you on that one :)

BrianGenisio:
I can't find a Fritzing component for this.  Is there a suggested approach to create a fritzing diagram for something like this that doesn't have a component?

rwaldron:
I make the art (svg) in iDraw and then use an existing component (which is dumb, but that's how Fritzing works) to create a new component

BrianGenisio:
Awesome.  I'm done with this PR, pending review and secondary validation.

rwaldron:
There's a merge conflict in programs.json

eg/accelerometer-mma7361.js
```diff
@@ -0,0 +1,36 @@
...
+  // This device also benefits from a calibration step.  You can autoCalibrate
+  // by placing the device level on startup.  You can also read the calibrated
+  // centers by reading the accelerometer.zeroV array after calibration.  Subsequent
+  // initializations, you can omit the autoCalibrate and set the zeroV array
+  // in the options instead
+
+  var accelerometer = new five.Accelerometer({
+    controller: "MMA7361",
+    pins: ["A0", "A1", "A2"],
+    sleepPin: 13,
```
rwaldron: I just realized this will be missed by the `Board.Component` checks for "currently occupied pins" :\

rwaldron:
Landed https://github.com/rwaldron/johnny-five/commit/fb02c8953a04fedb0712c3a839d7b366fa901dd5

PR #596, opened 2015-01-27
add feature to send to daisy-chained shift registers

boneskull:
- `ShiftRegister.prototype.send` now accepts *n* parameters.
  - Each parameter is shifted out in series.
  - If multiple parameters are given, the `value` of the `ShiftRegister` is an array.
- added test

I was not able to figure out a way to use daisy-chained shift registers with **johnny-five** without this change.  At least, every C/C++ example I've seen would flip the latch, do multiple sends, then flip the latch again.

If you would like, I can add an example to `eg/`.

Here's one now:

I made a 20-sided-die roller with two seven-segment displays, and two 74HC595's.  Here's a bit of code that simulates a "roll" by displaying random digits with decreasing speed, then settling on a number from 1-20, when a button is pressed:

```js
function run() {
  // shortcut for ic.send(0, 0)
  this.clear();
  this.btn.on('press', function () {
    // _.fill(n, m) will create an Array of length n having all values m
    var delays = _.fill(10, 16)
      .concat(_.fill(8, 32))
      .concat(_.fill(6, 64))
      .concat(_.fill(4, 128))
      .concat(_.fill(2, 256))
      .concat(512);
    this.clear();

    Promise.bind(this)
      .return(delays)
      .each(function (delay) {
        // digits is a multidim array; each item is a pair (one for each 7-seg) of 
        // numeric values corresponding to a combination of segments.  
        return Promise.delay(this.ic.send.apply(this.ic, digits[this.randomDigit()]), delay);
      });
  }.bind(this));
}```

### CODE REVIEW

**Comment by rwaldron** [comment]
- [ ] Add an example to `eg/`, include a breadboard diagram (diagram file is named the same as example program)
  - [Fritzing](http://fritzing.org/home/)
  - Please use "curvy wires and legs": `Fritzing > Preferences > Breadboard View > [✓] Curvy wires and legs`
- [ ] Add an entry in `programs.json` under the `Shift Register` topic
- [ ] Update https://github.com/rwaldron/johnny-five/wiki/ShiftRegister

**Comment by boneskull** [comment]
Sounds good.  I was about to fritz this anyway.

**Comment by rwaldron** [comment]
Awesome!

**Comment by rwaldron** [comment]
Also, commit the .fzz file along with the png—this way other can use your work as a starting point :)

**Comment by boneskull** [comment]
@rwaldron I'm having trouble getting Fritzing to do what I want.  For example, the mini-breadboards I'm using are not the same (they have too large of a border), and the 7-segments are not the same, which is causing me to come up with something pretty ugly.  

If you want to skip the Fritz for this, I'm happy to skip it.

**Comment by rwaldron** [comment]
Can you take a few pictures of your hardware and link them here? I might be able to help with the fritzing part

**Comment by boneskull** [comment]
oh jeez, this is gonna be embarrassing

**Comment by rwaldron** [comment]
Nothing to be embarrassed about, we do this stuff as a team and help each other out :)

**Comment by boneskull** [comment]
Images [here](https://flic.kr/s/aHsk7rA4mX).

Video [here](http://youtu.be/u5mjDYqbeHU).

**Comment by BrianGenisio** [comment]
LOL.  That's WAY nicer than anything I'd post.  #impressed

**Comment by boneskull** [comment]
@BrianGenisio Ha, thanks.  I didn't really need the full-size breadboard, but I didn't figure that out until later.

**Comment by boneskull** [comment]
@rwaldron To note, the button on the full-size breadboard can be ignored; it's supposed to reset the 74HC595's, but isn't necessary for the example.

**Comment by rwaldron** [comment]
@boneskull can you use this as a starting point: [seven-segment.fzz](https://dl.dropboxusercontent.com/u/3531958/seven-segment.fzz)?

**Comment by boneskull** [comment]
Ok, I have this, but it's kind of...verbose.  Though, I suppose the circuit is as well.  

seven-segment-fritz

Fritzing is completely unable to auto-route this, but I'm assuming you still want a reasonable schematic?

**Comment by boneskull** [comment]
@rwaldron I've added examples and all that jazz.  Also #597 is rebased on top of this.

**Comment by boneskull** [comment]
Is anyone interested in accepting a fourth pin in the constructor?  `reset`

**Comment by rwaldron** [comment]
@boneskull looking awesome! I just checked off the completed items in the list above—can you make sure to update the ShiftRegister docs?


> Is anyone interested in accepting a fourth pin in the constructor? reset

If you're willing to do the work, go for it :)

**Comment by boneskull** [comment]
@rwaldron Done.  I'll open a new PR for the reset stuff

**Comment by rwaldron** [comment]
Sounds good.

File: test/shiftregister.js [hunk 1]
```diff
@@ -76,6 +76,11 @@ exports["ShiftRegister"] = {
     test.ok(spy.getCall(51).calledWith(4, 1)); // latch, high
     test.equals(this.shiftRegister.value, 16);
 
+    this.shiftRegister.send(0x01, 0x01);
+    test.ok(spy.getCall(52).calledWith(4, 0));
+    test.ok(shiftOutSpy.calledWith(2, 3, true, 1));
+    test.ok(spy.getCall(101).calledWith(4, 1)); // latch, high
+    test.deepEqual(this.shiftRegister.value, [1, 1]);
```
rwaldron: One last request here: split these out into a new test called "sendMany" (or something like that)

boneskull:
@rwaldron sorry abt that, it's done now.

PR #597, opened 2015-01-27
add clear() function to ShiftRegister

boneskull:
- by replacing the value or values of the `ShiftRegister` with `0`'s, we effectively clear it
- added constructor option `size`, which determines whether the initial value is `0` or an array of `0`'s
- util function `numericFill()` fills an array of length `size` with `0`'s, in the most straightforward non-ES6 way that I'm aware of
- added test

**This PR requires a059050, which is PR #596.**


lib/shiftregister.js
```diff
@@ -2,6 +2,10 @@ var Board = require("../lib/board.js");
 
 var priv = new Map();
 
+function numericFill(size) {
+  return new Array(size + 1).join("0").split("").map(parseFloat);
+}
```
rwaldron: This can be removed, it's unnecessary

rwaldron:
> non-ES6 way that I'm aware of

Why? This project is ES6 friendly and es6-shim is available throughout :)

lib/shiftregister.js
```diff
@@ -17,7 +21,11 @@ function ShiftRegister(opts) {
     latch: opts.pins.latch
   };
 
-  priv.set(this, { value: 0 });
+  this.size = opts.size || 1;
+
+  priv.set(this, {
+    value: this.size > 1 ? numericFill(this.size) : 0
```
rwaldron: `Array(10).fill(0)`

lib/shiftregister.js
```diff
@@ -28,14 +36,34 @@ function ShiftRegister(opts) {
...
 };
 
+/**
+ * Clear the shift register by replacing each value with a 0.
+ * @type {ShiftRegister}
+ */
+ShiftRegister.prototype.clear = function () {
+  var value = priv.get(this).value;
+  if (Array.isArray(value)) {
+    return this.send.apply(this, numericFill(value.length));
```
rwaldron: `numericFill(value.length)` => `Array(value.length).fill(0)`

boneskull:
Was not aware of that; I'll use es6-shim instead

lib/shiftregister.js
boneskull: apparently it's even easier to just `new UInt8Array(value.length)`

rwaldron: Works for me. Since it's used to clear, you can just declare it once in the module scope and reuse it for all calls to `clear()`

rwaldron:
@boneskull any updates to push?

boneskull:
For some reason GitHub didn't recognize that I pushed a new changeset, but I made the `fill`-related changes you requested.

boneskull:
@rwaldron this is now rebased on top of #596

PR #619, opened 2015-02-12
Update stepper-driver.md

deshawnbw:
This calls out that AdvancedFirmata should be used.

### DIFF
--- a/docs/stepper-driver.md
+++ b/docs/stepper-driver.md
@@ -1,6 +1,8 @@
 <!--remove-start-->
 # Stepper - Driver
 
+In order to use the Stepper class, your board must be flashed with AdvancedFirmata, which is available here: https://github.com/soundanalogous/AdvancedFirmata
+
 Run with:
 ```bash
 node eg/stepper-driver.js


rwaldron:
@deshawnbw Thanks for the contribution—unfortunately, you'll need to actually move the location of this message to `eg/stepper-driver.js`. The markdown files are all generated so your change will be lost on the next regen :\

deshawnbw:
was a little too easy to edit via github, I'll put that in as a separate pull request then, thanks!

rwaldron:
You can still edit `eg/stepper-driver.js` via github and I can do the regen when I land it

soundanalogous:
You can also use the stepper motor with [ConfigurableFirmata](https://github.com/firmata/arduino/releases/tag/v2.6.2) however that requires the user to update the Firmata library within the Arduino application whereas AdvancedFirmata works with the version of Firmata that ships with Arduino. ConfigurableFirmata is also harder to use (since you need to currently manually configure it, but [work is in progress to automate this](https://github.com/firmata/firmata-builder)).

rwaldron:
@soundanalogous thanks for the reminder and updated here https://github.com/rwaldron/johnny-five/commit/b05e95a45a717c78e7230f1acf6684a9a1c1b47b

PR #620, opened 2015-02-13
Update stepper-driver.js

deshawnbw:
Making sure its obvious from the example code that in order for this to work you will need AdvancedFirmata uploaded to the board.


rwaldron:
The travis failure is expected and ok :)

Issue #626, opened 2015-02-19
Migrate from colors.js to chalk


rwaldron:
https://github.com/sindresorhus/chalk

julianduque:
yes! @rwaldron can you assign this to me? I would love to work on it :)

rwaldron:
@julianduque it's all yours

scottgonzalez:
This is done.

PR #637, opened 2015-02-24
Migrate from colors.js to chalk

julianduque:
Related: #626 

Since chalk doesn't support the `orange` color I used the `reverse` for failure messages but I'm open to suggestion on what color to use here, this also fixes a couple of examples using colors and remove some unused colors logic from one example (`eg/magnetometer-north.js`)


julianduque:
@rwaldron Piezo tests are failing in a non deterministic way (on travis and local) but aren't related to this change

rwaldron:
Sorry I missed this!

PR #640, opened 2015-03-01
Add isAnode to LED state. Fixes gh-629

### LINKED ISSUE #629: RGB LED: isAnode: true not working
When I declare an RGB LED in the following way: 

```
	var led1 = new j5.Led.RGB({
		isAnode: true,
		pins: {
			red: 2,
			green: 3,
			blue: 4
		}
	});
```

isAnode is not sticking. Debugging shows that there is no `isAnode` property on `state` when I call `write()`. 

Currently investigating a fix, will report back when I figure out what's causing it.

frxnz:
* [x] Add `isAnode` to Led and Led.RGB state
* [x] Fix `analogWrite` value bug in PCA9685 controller
* [x] Add test for `isAnode`
* [x] Update docs (Led.RGB has no method `pulse`)

Fixes #629 

Confirmed working with `PCA9685` and `DEFAULT` controllers

### DIFF
--- a/docs/led-rgb-anode.md
+++ b/docs/led-rgb-anode.md
@@ -21,7 +21,12 @@ board.on("ready", function() {
     isAnode: true
   });
 
-  anode.pulse();
+  anode.strobe();
+
+  this.repl.inject({
+    anode: anode
+  });
+
 });
 
 ```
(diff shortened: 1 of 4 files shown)


docs/led-rgb-anode.md
```diff
@@ -21,7 +21,12 @@ board.on("ready", function() {
     isAnode: true
   });
 
-  anode.pulse();
+  anode.strobe();
```
rwaldron: Use "blink" instead

frxnz:
> Use "blink" instead

Done

PR #657, opened 2015-03-19
Update the led pulse example

kisPocok:
The guide image shows port 13. Let's use that.

### DIFF
--- a/docs/led-pulse.md
+++ b/docs/led-pulse.md
@@ -15,7 +15,7 @@ board.on("ready", function() {
 
   // Create a standard `led` component
   // on a valid pwm pin
-  var led = new five.Led(9);
+  var led = new five.Led(13);
 
   led.pulse();
 


dtex:
Hi @kisPocok ,

the MarkDown files in /docs are generated from the js files in /eg. Make your edit to eg/led-pulse.js and then run ```grunt examples```.

kisPocok:
Yeah, I got it after the test failed. I will update it soon, please revert/refuse this changes.

dtex:
New PR #660

Resseguie:
It actually needs to be 9 because 13 isn't PWM. It's the diagram that needs to be updated.

dtex:
We should re-open #660 instead. This one was never merged.

I'm a one man 3 Stooges act right now. Excuse me while I go poke myself in the eye.

kisPocok:
Sorry for the trouble. I didn't know that, thanks for the heads up!

Resseguie:
No problem. The examples have been created somewhat haphazard along the way so it's always good to have extra sets of eyes. I'd like to go through and check them all for consistency, style, etc.

rwaldron:
>:|


Start here and move forward in time: https://github.com/rwaldron/johnny-five/commits/master?page=5

I reviewed every single example and remade most of them, the problem is that the task is time consuming and mind numbing and it's often easy to reuse the same diagram for similar examples. 

Yes, another set of eyes.

Resseguie:
Oh good, it looks like you've done the hard work already! Cool. I haven't reviewed them since that update. I'll make a note to do a sanity check pass-through.

rwaldron:
Yes, please do. More eyes are better. Also, some examples may be missing diagrams, these need to be created

PR #661, opened 2015-03-20
Revert "Update LED Pulse example"

dtex:
Reverts rwaldron/johnny-five#660 The code was correct. The Fritzing diagram is what was wrong. Pin 13 on Uno is not PWM capable.


rwaldron:
Nice, good catch

dtex:
@Resseguie caught my mistake. The png had pin 13, the code had pin 9, and the fzz had pin 11. Woulda had a PR in to get everything on pin 11 already, but I'm a dumbass and made #659 on my master.

PR #664, opened 2015-03-22
Change isInverted property to invert

### LINKED ISSUE #643: Button, Servo (and others?): isInverted or invert?
Problem: 

- `Button` has a constructor options property called `invert`
- `Servo` has a constructor options property called `isInverted`

At this point I think we need to just make it so that _either_ will work on both, then drop one of them for 1.0

dtex:
Fixes #643. Updated all examples and docs as well. Still supports isInverted but console.warns.


rwaldron:
Nice! 

LGTM—land at will

PR #671, opened 2015-03-25
use environment to declare test mode flag; resolves #669

### LINKED ISSUE #669: IS_TEST_MODE could be nicer
`IS_TEST_MODE` is a global, set throughout several files in `test/`, but most importantly (I think) `test/bootstrap.js`.  

This is all well and good if you are running the tests via Grunt.  However, if you wish to run the Nodeunit tests directly via an interactive debugger--thinking JetBrains' debugger--there's no way to specify the order in which the test files should run.  

That's a limitation of my debugger, however, this is an edge case.  In a perfect world, the test files *should* run in any order.  Perhaps Nodeunit should support something like Mocha's `--require` which allows you to first require a "fixture" module, like `bootstrap.js`, before tests begin; then, all of the `purge()` functions could live in the fixture module).  Maybe we should switch to Mocha, but that's another can of worms.  :smile:

I haven't touched Java since 1.2--and I have my own test framework to worry about instead of Nodeunit--so the quickest fix for me is to modify J5 to support an environment variable instead.

**Summary of proposed changes:**

- Anywhere `IS_TEST_MODE` is retrieved, `process.env.IS_TEST_MODE` is retrieved instead
- Anywhere `IS_TEST_MODE` is set, `process.env.IS_TEST_MODE` is set instead
- `Gruntfile.js` explicitly sets `process.env.IS_TEST_MODE = true;`
- `test/bootstrap.js` and references are removed
- `IS_TEST_MODE` is removed from various JSHint configs

Then, you can use whatever debugger you want, as long as you can set an environment variable in some manner.  This is a more typical solution than setting a property on `global`--I think for good reason.

> *Note: A workaround exists, which is to debug `node grunt` itself, but that bypasses Jetbrains' Nodeunit reporter, which is a nicer interface.*

boneskull:
- Relevant files check `process.env.IS_TEST_MODE` instead of a global prop
- The value of `IS_TEST_MODE` is assumed to be a boolean
- Individual suites do not set `IS_TEST_MODE`
- `bootstrap.js` is no longer necessary and has been removed
- `Gruntfile.js` sets the `IS_TEST_MODE` flag; as long as you run the tests via `npm test` or `grunt`, your environment will be correct.  Otherwise, you must manually set the `IS_TEST_MODE` environment variable.
- simplified test targets within `Gruntfile.js` with globs
- remove `bootstrap.js` reference within `nodeunit:just`; while I was in there, corrected a potential Windows incompatibility, made slightly more flexible, and documented it for use with `grunt --help`

Happy to remove the changes I snuck into `Gruntfile.js` if you wish.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,6 +4,8 @@ var inspect = require("util").inspect;
 var fs = require("fs");
 var shell = require("shelljs");
 
+process.env.IS_TEST_MODE = true;
+
 module.exports = function(grunt) {
 
   var task = grunt.task;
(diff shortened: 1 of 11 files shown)


Gruntfile.js
```diff
@@ -33,44 +35,8 @@ module.exports = function(grunt) {
...
-        "test/servo.js",
-        "test/shiftregister.js",
-        "test/sonar.js",
-        "test/stepper.js",
-        "test/temperature.js",
-        "test/switch.js",
-        "test/wii.js"
-      ]
+      // remove test/distance.js (and this exception) when bumping to 0.9.0
+      tests: ["test/*.js", "!test/mock*", "!test/distance.js"]
```
rwaldron: I appreciate what you've done here, but I have about 30 untracked files in the test/ dir of my local repo that shouldn't be run—unless you know of a good way around that?

boneskull: we could tell it to avoid some specific naming convention?

rwaldron: If only I had used some naming convention. It's a lot less work for me (in the near term) if we just don't change this ;)

PR #672, opened 2015-03-25
reduce total size of dependencies by cherry-picking LoDash functions

boneskull:
- instances of `_.assign()` become `Object.assign()`
- remove dependency `lodash`
- add dependencies:
  - `lodash.clonedeep`
  - `lodash.debounce`
  - `lodash.mixin`
  - `lodash.every`
  - `lodash.pluck`
  - `lodash.includes`
  - `defaults` (smaller than the LoDash counterpart)
- added `lodash.includes` to `fn.js` (`__`) as `contains`

I'm not sure if this is worth the bother, as it saves us around one (1) MB, however it's setting a good precedent for future contributors--they'll know that j5 doesn't use LoDash proper.

### DIFF
--- a/lib/board.js
+++ b/lib/board.js
@@ -6,7 +6,6 @@ var Emitter = require("events").EventEmitter;
 var util = require("util");
 // var os = require("os");
 var chalk = require("chalk");
-var _ = require("lodash");
 var __ = require("../lib/fn.js");
 var Repl = require("../lib/repl.js");
 var Options = require("../lib/board.options.js");
@@ -215,7 +214,7 @@ function Board(opts) {
 
   // Initialize this Board instance with
   // param specified properties.
-  _.assign(this, opts);
+  Object.assign(this, opts);
 
   this.timer = null;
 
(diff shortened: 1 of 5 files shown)


boneskull:
hmm, wtf

boneskull:
Travis failure looks like some sort of race condition or timing problem

dtex:
@boneskull It is. That test fails intermittently on Travis. Hopefully we will be moving it to a separate group of tests that don't run automatically. For now I've restarted the Travis build.

dtex:
It's no longer just the Piezo test failing.

rwaldron:
@dtex I would need to dig around, but my first thought is that there maybe a "depth" problem with the change from `_.assign` to `Object.assign`

boneskull:
hmm, so recursive objects somewhere perhaps?

rwaldron:
Possibly, or just a property whose value is an object and that reference had some unexpected semantics attached to it :\

boneskull:
I'm going to abandon this.

Issue #647, opened 2015-03-15
Update ESC class


rwaldron:
- [x] Update to controller pattern
- [X] Implement PCA9685 controller

Issue #666, opened 2015-03-24
Servo offset bug


dtex:
There is a bug in servo that I wanted to document in case anybody else has struggled with it. It appears when a servo opts object has both a range and offset property and the problem is compounded when a servo is inverted.

The offset property applies a transform to the local coordinate space of the servo. We are currently enforcing our range limits after the offset has been applied. This means that the range is being applied on the local coordinate space. It should be applied on the global coordinate space.

Using this example of a coxa joint in global coordinate space:

````
var coxa = new Five.Servo({offset: -90, range: [155, 205]});
````

![jcs-vs-gcs](https://cloud.githubusercontent.com/assets/854911/6784042/5c12a00c-d14b-11e4-9546-2bf5ca174d7d.png)
black values are global, red values are local

When the user calls a value that is outside the local range, but within the global range.

````
coxa.to(180);
// Expected behavior is to set servo to 90°
// Actual behavior is servo goes to 65°
````

On the other hand if we configure our servo like this:

````
var coxa = new Five.Servo({offset: -90, range: [65, 115]});
````

when the user calls a value that is outside the local range, but within the global range.

````
coxa.to(180);
// Expected behavior is to set servo to 90°
// Actual behavior is servo goes to 115°
````

I was worried the fix might be a breaking change, but the only way this would have ever worked at all for a user is if they opened up their range properties to the point where both their local and global coordinate values were allowed:

````
var coxa = new Five.Servo({offset: -90, range: [65, 205]});
````

**But wait, it gets worse**

It is also a problem when a servo is inverted and this one is going to be tougher to fix. Assume the servo in our drawing is mounted face down so that our servo's local coordinate space is inverted and our servo is configured like this:

````
var coxa = new Five.Servo({offset: -90, invert:true, range: [65, 205]});
````

````
coxa.to(160);
// Expected behavior is to set servo to 90°
// Actual behavior is servo goes to 200°
````

The inversion is base on the global coordinate space range instead of the local coordinate space range.

This offset fix is easy, but I'm still noodling over the ```invert: true``` solution. A PR is coming soon.

rwaldron:
Nice find and even better documentation :)

dtex:
Fixed wtih 29783f8

PR #673, opened 2015-03-25
ESC: Update to controller model. Adds PCA9685. Closes gh-647

rwaldron:
cc @dtex


rwaldron:
TODO: Update https://github.com/rwaldron/johnny-five/wiki/ESC when this lands

rwaldron:
@dtex I'm going to wait on this one until your offset patch lands.

dtex:
This looks really good to me. A few notes:

* Do we need to add deadband on ESC? I recall you were having a conversation with someone recently whose ESC didn't do anything from 51 - 54 (or something like that).
* Shouldn't the API for Motor and ESC be identical?
* I know the ESC class predates the controller model, but if I had a time machine I would make ESC a controller on Motor.
* PCA9685 support? **Bring on the hexadecicopters!**

> @dtex I'm going to wait on this one until your offset patch lands.

You are a kind man, but no need. That actual changes required for #666 (the bug from hell) are minimal so rebasing will be easy. I just need to get the logic right.

rwaldron:
> Do we need to add deadband on ESC? I recall you were having a conversation with someone recently whose ESC didn't do anything from 51 - 54 (or something like that).

I'm waiting until I have a better understanding for bi-directional speed controllers. I think that makes a good follow up feature


> Shouldn't the API for Motor and ESC be identical?

I'll have a look, that might be a really good idea :)

> I know the ESC class predates the controller model, but if I had a time machine I would make ESC a controller on Motor.

Part of me agrees and part of me disagrees. I'm not sure which one to believe. 


> That actual changes required for #666 (the bug from hell) are minimal so rebasing will be easy. I just need to get the logic right.

Ok, sounds good.

Issue #530, opened 2014-12-13
ESC


carldanley:
I'm trying to use this ESC: https://traxxas.com/products/parts/otherparts/xl5waterproof. I have jumper wires that run from the ground and input on the ESC to the arduino. The arduino is being powered by my macbook (over USB) and the ESC is being powered properly with the RC car's battery.

I'm running this code:
```
const five = require('johnny-five');
const board = new five.Board();

board.on('ready', function() {

  // ENGINE
  const motorServo = new five.Servo({
    pin: 10
  });

  this.repl.inject({
    s: motorServo
  });
  
});
```

From the repl, I can run `s.to(90)` to stop the motor. I can increase the speed from 90 to 160 where 160 is the maximum speed (35mph) for the RC car.

The RC car is supposed to be able to reverse and I've found that no matter what values I use, I cannot switch from forward to reverse and back. I've gotten both directions to work independently but I do not know what I did to get them to work.

Any suggestions or ideas as to what I need to be doing for this to work properly?

landshark22:
Did anyone ever solve this problem? We are having a few issues communicating with our ESC including getting reverse to work properly.

rwaldron:
No, not yet. @carldanley was going to ship his bot to me so I could work on it... not sure if that's still the plan. I just bought an ESC that has reverse capabilities—that will be here on Tuesday, which is the soonest I can work on this. I appreciate your patience :)

rwaldron:
I just found this on another site...

> More or less all ESCs have a similar calibration procedure:
> - Power up the ESC while the having maximum forward throttle applied
> - You’ll hear a tone and some beeps and after a while (usually 2 seconds) you’ll hear a confirmation tone and the led will blink a few times with a different color: this indicates that the ESC has measured the wavelenght of max throttle.
> - At this point apply zero throttle (in a fwd/reverse ESC this means full throttle reverse), wait again few seconds for the tones and led to blink: full reverse measured.
> - Then move to central (only for fwd/reverse ESCs) and wait again for the tone and blinks.


Have you tried calibrating like this?

carldanley:
@rwaldron Thanks for the reminder! Email me your address so I can send my RC car your way.

rwaldron:
@carldanley let's hold off on that. I have a bidirectional ESC on it's way, so it might be able to solve this without needing your entire RC car :D

rwaldron:
Just an update: my speed controller arrived, so I can start working on this on my end.

[...]

PR #676, opened 2015-03-26
ESC: Bidirectional speed controller (Forward/Reverse) support. Fixes gh-530

rwaldron:
See: https://github.com/rwaldron/johnny-five/issues/530

Specifically: 

- https://github.com/rwaldron/johnny-five/issues/530#issuecomment-86597026
- https://github.com/rwaldron/johnny-five/issues/530#issuecomment-86605385

Notes about this implementation: 

- There is no way to implement bidirectional capability as we do with `Motor` class, ie. with a `dir` pin, since a single pin is used with a speed controller. 
- There may be a smart way to programmatically enforce the "calibrate" step, but I'm not seeing it. Another set of eyes/brains might help me :)
  - Calibration: 
    - Turn the speed controller OFF
    - Start the program (the program initializes a type=bidirectional ESC with an explicit neutral point)
    - Once the program is running, switch the speed controller ON
    - Wait for the tones to indicate calibration is complete. 
  - This only needs to be done once.

  
http://www.youtube.com/watch?v=g2GqoWXJC30


eg/esc-bidirectional.js
```diff
@@ -0,0 +1,30 @@
+var five = require("../");
+var board = new five.Board();
+
+board.on("ready", function() {
+  var start = Date.now();
+  var esc = new five.ESC({
+    type: "bidirectional",
```
rwaldron: Alternatively...


```js
// Assume...
// ESC.F = 0, Forward (default)
// ESC.FR = 1, Forward/Reverse
// ESC.FBR = 2, Forward/Brake/Reverse (doubt this can be supported, but included for illustration)


var esc = new five.ESC({
  mode: ESC.FR, 
  neutral: 50,
  pin: 11
});
```

rwaldron:
cc @dtex 

@carldanley this is a first draft of bidirectional speed controller support

dtex:
Are we missing an opportunity to leverage the device/controller model?

```ESC.forward()```, ```ESC.reverse()```, etc could be methods on ```Devices.FR```, ```Devices.FBR```

**Comment by carldanley** [comment]
Great idea, @dtex; :+1:.

**Comment by carldanley** [comment]
Reviewed the commit and looks outstanding! Will test later tonight (after work).

**Comment by rwaldron** [comment]
> Are we missing an opportunity to leverage the device/controller model?

@dtex yeah, that seems like a good idea. I'll update this when I have some time :)

**Comment by rwaldron** [comment]
@dtex Devices implemented.

File: lib/esc.js [hunk 2]
```diff
@@ -72,6 +73,7 @@ var Controllers = {
       }
     },
     write: {
+      writable: true,
```
dtex: Just curious, why is ```writable: true``` here? It doesn't hurt anything but ```write``` isn't on Devices so it doesn't ever get changed.

dtex:
I don't have an esc to test with, but the devices implementation looks good to me.

dtex:
I think @Resseguie 's work on #693 will have to be reconciled with this.

PR #688, opened 2015-04-03
Consolidate all the proximity/ranging devices

frxnz:
This isn't done yet, but I wanted to get some input on the device/controller stuff. The way I understand it a device is used to transform the raw data from the sensor, while a controller is used to actually retrieve that data from the sensor. [Specifically here](https://github.com/rwaldron/johnny-five/pull/688/files#diff-c5fe53df19f93fb5988d5336a99e42a9R3).

cc @rwaldron @dtex


rwaldron:
> device is used to transform the raw data from the sensor, while a controller is used to actually retrieve that data from the sensor.

`controller` provides all facets of read/write to a component. A `device` is used for device-specific functionality to be layered in, eg. overwriting methods of the component class. 

There isn't any clear need to use a `device` for these components. `Temperature` or `Accelerometer` are good examples to align with

eg/proximity.js
```diff
@@ -3,49 +3,13 @@ var five = require("../lib/johnny-five.js"),
...
-  //   new five.IR({
-  //    device: "GP2Y0D805Z0F",
-  //    freq: 50
-  //   });
-  // )
-  //
+  var proximity = new five.Proximity({
+      pin: "A0",
+      controller: "OA41SK"
+    });
```
rwaldron: Weird indentation here

eg/proximity.js
```diff
@@ -3,49 +3,13 @@ var five = require("../lib/johnny-five.js"),
...
-  // when no movement has occurred in X ms
-  //
-  prox.on("motionstart", function(err, timestamp) {
-    led.on();
-  });
-
-  prox.on("motionend", function(err, timestamp) {
-    led.off();
+  proximity.on("data", function() {
+      console.log("cm: ", this.centimeters, this.raw);
```
rwaldron: indented too far 

(run `grunt jsbeautifier`)

lib/johnny-five.js
```diff
@@ -78,6 +79,7 @@ module.exports.IR.Motion = function(opt) {
 };
 
 module.exports.IR.Proximity = function(opts) {
+  console.log("IR.Proximity is deprecated. Use Proximity instead");
   // Fix a naming mistake.
   if (module.exports.Distance.Controllers.includes(opts.controller)) {
     return new module.exports.Distance(opts);
```
rwaldron: :heavy_check_mark:

lib/proximity.js
```diff
@@ -0,0 +1,207 @@
...
+
+  //     prime.call(this);
+  //   },
+  //   toCm: {
+  //     value: function() {
+  //       var median = priv.get(this).median;
+  //       return +((((median / 2) * 343.2) / 10) / 1000).toFixed(1);
+  //     }
+  //   }
+  // },
```
rwaldron: This doesn't have to block this patch—we can add the new controller later

frxnz:
@rwaldron I added `SRF10` but I'm not confident it works. If you have one you could test that would be great. I can't get tests to run on it, so I'm guessing either there's an error or I'm terrible at tests (both are likely).

rwaldron:
I will test that for you today!

eg/proximity.js
```diff
@@ -3,49 +3,14 @@ var five = require("../lib/johnny-five.js"),
...
-    led.on();
+  var proximity = new five.Proximity({
+    pin: "A0",
+    controller: "OA41SK",
+    freq: 250
   });
 
-  prox.on("motionend", function(err, timestamp) {
-    led.off();
+  proximity.on("data", function(err, data) {
```
rwaldron: For events, we're trying to get rid of the `err, data` arguments. Update the class, tests, etc. to emit only a `data` object

lib/proximity.js
```diff
@@ -0,0 +1,249 @@
...
+        // Startup parameter
+        this.io.i2cWrite(address, [0x01, 16]);
+        this.io.i2cWrite(address, [0x02, 255]);
+
+        // Is this still a thing?
+        //this.io.setMaxListeners(100);
+
+        function read() {
+          this.io.i2cWrite(address, [0x02]);
+          this.io.i2cRead(address, 2, function(data) {
```
rwaldron: I tested this locally and the only change that needs to be made is here: 

`this.io.i2cRead` => `this.io.i2cReadOnce`

(This device can't be read continuously)

rwaldron:
> I can't get sinon tests to run on it,

I'll take a look at this next

eg/proximity.js
frxnz: Ah I thought that was the case, but I opened a couple files and they had `err`. I'll clean those up.

lib/proximity.js
```diff
@@ -0,0 +1,249 @@
+var Board = require("../lib/board.js"),
+  events = require("events"),
```
rwaldron: Add: 

```js
  within = require("./mixins/within"),
  __ = require("./fn"),
```

lib/proximity.js
```diff
@@ -0,0 +1,249 @@
...
+Proximity.Controllers = [
+  "2Y0A21", "GP2Y0A21YK",
+  "2D120X", "GP2D120XJ00F",
+  "2Y0A02", "GP2Y0A02YK0F",
+  "OA41SK", "GP2Y0A41SK0F",
+  "0A21", "GP2Y0A21YK",
+  "0A02", "GP2Y0A02YK0F",
+];
+
+util.inherits(Proximity, events.EventEmitter);
```
rwaldron: Add: 

```js
__.mixin(Proximity.prototype, within);
```

rwaldron:
> so I'm guessing either there's an error or I'm terrible at tests (both are likely).

Definitely bugs :(

rwaldron:
I branched from your branch and committed a patch to fix the tests. You can pull it into your branch directly with: 

```
git pull https://github.com/rwaldron/johnny-five.git frxnz-proximity-refactor-fix-tests
```

eg/proximity.js
rwaldron: Clean up only for your Proximity and Motion classes. I have a patch I'm maintaining for all the rest

frxnz:
@rwaldron I just updated the examples. If there are any changes let me know, otherwise I think this is ready to land. I'll open a separate PR for `Ping`.

rwaldron:
> otherwise I think this is ready to land.

This **does** look beautiful

>  I'll open a separate PR for Ping.

Not sure I understand this—why? 


Also, Ping is going to have two different controllers, so we need to coordinate with @ajfisher on naming.

frxnz:
> Not sure I understand this—why?

If you're not in a hurry to land I can do it here, but I may not be able to get to Ping until Monday.

rwaldron:
I'm not in a hurry :)

rwaldron:
@frxnz sorry I missed this update a week ago—next time don't hesitate to ping me :)

frxnz:
@rwaldron No worries, I've been super busy and haven't had much time to look at the tests myself. I got an HC-SR04 last week and fixed up the controller for it yesterday. Just the `change` and `within` tests holding us back now.

rwaldron:
Cool, that's helpful (good direction when I sit down to test)

frxnz:
@rwaldron Rebased, updated the docs, and did a little cleanup. I don't know if you've had a chance to look at them, but the `change` and `within` tests for HC-SR04 are still giving me trouble.

rwaldron:
I haven't, but I will tomorrow. Thanks for all your works so far :)

eg/proximity-hcsr04.js
```diff
@@ -0,0 +1,19 @@
+var five = require("johnny-five.js"),
```
rwaldron: `"johnny-five.js"` => `"../"` (this is automatically converted by the examples task)

eg/proximity-mb1003.js
```diff
@@ -0,0 +1,19 @@
+var five = require("johnny-five.js"),
```
rwaldron: `"johnny-five.js"` => `"../"` (this is automatically converted by the examples task)

frxnz:
Big thanks to @rwaldron and @Resseguie for helping me get the HC-SR04 tests working :tada:

PR #689, opened 2015-04-03
Fixes to servo

dtex:
Addresses #666

1) Range was being constrained after offset was applied which was
wrong. User should not have to think in two different coordinate spaces.
2) Invert was based on global coordinate range which led to incorrect
values being passed to the servo. Invert has to be based on the
physical range of the servo in local coordinate space. We need to
track two ranges, the global coordinate range for ease of use and the
physical device range of the servo. We are assuming 180° servos for
now. We will have to add in 45° and 270° servos as Devices.
3) Added tests

### DIFF
--- a/lib/servo.js
+++ b/lib/servo.js
@@ -261,11 +261,10 @@ Servo.prototype.to = function(degrees, time, rate) {
 
   var target = degrees;
 
-  degrees += this.offset;
-
   // Enforce limited range of motion
   degrees = Board.constrain(degrees, this.range[0], this.range[1]);
 
+  degrees += this.offset;
   this.value = degrees;
 
   var last, distance, percent;
@@ -275,8 +274,8 @@ Servo.prototype.to = function(degrees, time, rate) {
   if (this.invert) {
     degrees = Board.map(
       degrees,
-      this.range[0], this.range[1],
-      this.range[1], this.range[0]
+      0, 180,
+      180, 0
     );
   }
 
--- a/test/servo.js
+++ b/test/servo.js
@@ -161,6 +161,160 @@ exports["Servo"] = {
     test.done();
   },
 
+  range: function(test) {
+    test.expect(3);
+
+    this.servo = new Servo({
+      pin: 11,
+      board: board,
+      range: [20, 160]
+    });
+
+    this.servo.to(180);
+
+    test.ok(this.servoWrite.calledWith(11, 160));
+
+    this.servo.to(135);
+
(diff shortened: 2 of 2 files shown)


lib/servo.js
```diff
@@ -261,11 +261,10 @@ Servo.prototype.to = function(degrees, time, rate) {
 
   var target = degrees;
 
-  degrees += this.offset;
-
   // Enforce limited range of motion
   degrees = Board.constrain(degrees, this.range[0], this.range[1]);
 
+  degrees += this.offset;
```
rwaldron: ✓

lib/servo.js
```diff
@@ -275,8 +274,8 @@ Servo.prototype.to = function(degrees, time, rate) {
   if (this.invert) {
     degrees = Board.map(
       degrees,
-      this.range[0], this.range[1],
-      this.range[1], this.range[0]
+      0, 180,
+      180, 0
```
rwaldron: I'm confused about this change. If `degrees += this.offset;` is moved after constraining to the user range, then why wouldn't this inversion operation be constrained to the user range. Can you give me a quick example of how this falls down?

rwaldron: I think I know what my disconnect is: I've assumed that a user specified range was exclusively to define the physical range. I've used range: [a, b] to limit the bounds of servos that have 180° range

dtex: If we invert based on user range:
```
var servo = new Five.Servo({
  range: [20, 140],
  invert: true
});
servo.to(20); // Moves servo to 40
```
If the values in range are not equidistant from the center position of the servo it will not be mapped correctly.

dtex: Sorry, saw your second note after I replied. Yes, we really have two ranges. Global coordinate range and local/device coordinate range (typically 0-180).

rwaldron: > If the values in range are not equidistant from the center position of the servo it will not be mapped correctly.

I wonder if this is what caused me headaches with some of the bipeds I built—in most cases I was specifying an artificial range.

rwaldron:
LGTM, land at will

lib/servo.js
dtex: I bet it is. On my hexapod I had started specifying absurdly loose range values just to get the servo to move when using a global coordinate space for the legs. I thought the problem was with my IK formulas, but I traced it back to here. It wasn't a problem when I was using a joint coordinate space for specifying positions because the offsets were small and I wasn't using ranges yet. It's a huge relief to have figured this one out.

PR #690, opened 2015-04-04
Support auto-calibration for reflectance array

BrianGenisio:
Instead of an explicit calibration step, or the requirement to load a calibration file, setting `autoCalibrate: true` will cause the unit to always be calibrating.  This will make the learning curve on the sensor to be less steep because they don't need to worry about calibrating.  Instead, as long as the unit is reading data, it will be setting the min/max.  This will also help in environments like Robotnik, where the pre-ready code becomes awkward.

Users can still tweak the calibration by using the original mechanism.  This is just easier with less control.  

We are using a similar mechanism in the `Accelerometer`, and I was quite fond of it after it was complete.

### DIFF
--- a/lib/reflectancearray.js
+++ b/lib/reflectancearray.js
@@ -60,6 +60,10 @@ function onData(sensorState, value) {
   if (allRead) {
     this.emit("data", null, this.raw);
 
+    if (state.autoCalibrate) {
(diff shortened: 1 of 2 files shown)


rwaldron:
LGTM

Issue #692, opened 2015-04-05
OneWire temperature sensors (DS18B20) can't have multiple devices per pin


BrianGenisio:
The OneWire spec allows multiple devices per pin, but the `Temperature` class doesn't handle multiple temperature sensors properly.
- [x] Bust out the DS18B20 controller into a sharable driver, like the IMU
- [x] Re-work the controller to register with the driver
- [x] Make sure the un-addressed sensor still works
- [x] Throw an error if there is more than one un-addressed sensor
- [x] Unit tests
- [ ] More real-world testing

dtex:
Fixed by #694

PR #694, opened 2015-04-05
Multi ds18b20

BrianGenisio:
This is to fix the bug https://github.com/rwaldron/johnny-five/issues/692

In order to support more than one DS18B20 OneWire sensors, we need to coordinate the OneWire requests to read so that they aren't clobbering each other.  I'm using the "Driver" approach that we used with the IMU where there may be multiple components/controllers, but they all use the same instance of a "driver" to coordinate a single resource.  In the case of the IMU, it was the single-unit I2C board.  In the case of this OneWire device, it is the OneWire bus itself.

I'm going to do a bit more manual testing tonight to make sure everything is good, but I think this code is solid enough for a PR.  I'll make minor tweaks if necessary.

### DIFF
--- a/docs/temperature-ds18b20.md
+++ b/docs/temperature-ds18b20.md
@@ -19,6 +19,7 @@ five.Board().on("ready", function() {
 
   temperature.on("data", function(err, data) {
     console.log(data.celsius + "°C", data.fahrenheit + "°F");
+    console.log("0x" + this.address.toString(16));
   });
 });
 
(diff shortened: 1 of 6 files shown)


rwaldron:
Replace `contains` with `this.addresses.includes(...)`

rwaldron:
I'm mobile now, will review more later

rwaldron:
Need to run `grunt examples` for travis to pass ;)

Issue #695, opened 2015-04-06
Better organization for breadboards and generated docs


Resseguie:
In #693 I found a couple aspects of the examples and generated docs process to be troublesome.

1) The `grunt examples` task updates docs for examples in programs.json but doesn't delete previously generated docs that are no longer referenced. We had a bunch of old outdated .md files in there.
2) To fix above, I just cleared out the docs folder before regenerating, but that required being careful that you don't accidentally delete the breadboard folder that contains all the diagrams.
3) There are also lots of duplicate breadboards (think led, led-blink, etc) that are difficult to maintain because you have to manually copy an updated breadboard to all the associated examples.

I have the following recommendations:

1) Move breadboards to a sub folder under eg and have a grunt task that clears the docs folder and copies breadboards from eg/breadboard to docs/breadboard before generating examples
2) Instead of requiring duplicate files for breadboards, add a breadboards.json that maps an example to a breadboard. If nothing is included for an example in breadboards.json, default to current behavior looking for one with the same name, or none included if that's also not found. This would significantly reduce the number of breadboard files we're maintaining and also make it easier to update them if needed in the future (like I just did to make them all look more consistent).

Example breadboards.json snippet:

``` js
{
  "led.js": "led-13",
  "led-blink.js": "led-13",
  "led-pulse.js": "led-11-pwm",
  "led-fade.js": "led-11-pwm",
  "led-demo-sequence.js": "led-11-pwm",
  "led-slider.js": "led-11-pwm",
  "led-rgb.js": "led-rgb",
  "led-rainbow.js": "led-rgb"
}
```

rwaldron:
> Move breadboards to a sub folder under eg and have a grunt task that clears the docs folder and copies breadboards from eg/breadboard to docs/breadboard before generating examples

This will break links to those files :\ (I've used the raw links in slide decks). I think it's safe to just glob for the files you want to delete. 


> Instead of requiring duplicate files for breadboards, add a breadboards.json that maps an example to a breadboard. If nothing is included for an example in breadboards.json, default to current behavior looking for one with the same name, or none included if that's also not found. 

I like this alot. If we change it to: 

```
{
  "led.js": ["led-13-uno", "led-7-spark", "led-whatever-someotherboard"],
}
```

Then we will solve another major problem that's been haunting me: how to **easily** include multiple diagrams for an example. Eventually I want diagrams for every single supported platform :)

Resseguie:
> This will break links to those files

[...]

PR #699, opened 2015-04-07
Allow more than one breadboard image to be specified

Resseguie:
Can add entries to `breadboards.json` if you want multiple breadboards for an example or to use an existing one with a different name. (Defaults to example name as before)


Resseguie:
Fixes #695, minus moving the breadboard diagrams to a new sub-folder.

Resseguie:
@rwaldron I made a single entry in breadboards.json to test with. I used led.js and added the simple diagram with LED direction in pin 13 and the one with a resistor. 

I'm not entirely happy with the markdown formatting for multiple diagrams. Because of the whitespace at the bottom of the image where the fritzing logo is, the .fzz link for the first image visually appears to go with the second diagram. I'm guessing I could fix that with some creative, but messy and non-obvious, markdown. Alternatively, the link on the image could point to the fzz (though I'm not a fan of that either).

I could change the simple breadboards.json format to use an object instead of an array, then you could assign a title to each diagram and have a header that visually separates them if multiples are given. Though it adds complexity to breadboards.json, that might solve the visual problem and give a way to describe each breadboard, especially once you get to your vision of one per platform.

thoughts? Or is this good enough for now?

Gruntfile.js
```diff
@@ -207,8 +203,34 @@ module.exports = function(grunt) {
...
+          } else {
+            verbose.writeln("Missing PNG: " + png);
+          }
+
+          if (hasFzz) {
+            images += templates.fritzing({ fzz: fzz });
+          } else {
+            verbose.writeln("Missing FZZ: " + fzz);
+          }
+        });
```
rwaldron: :white_check_mark:

Gruntfile.js
```diff
@@ -218,8 +240,7 @@ module.exports = function(grunt) {
           example: eg,
           file: md,
           markdown: markdown.join("\n"),
-          breadboard: hasPng ? templates.img({ png: png }) : "",
-          fritzing: hasFzz ? templates.fritzing({ fzz: fzz }) : ""
+          breadboards: images.slice(0, -1)
```
rwaldron: What is the last character?

docs/led.md
```diff
@@ -31,8 +31,10 @@ board.on("ready", function() {
 ## Breadboard/Illustration
 
 
-![docs/breadboard/led.png](breadboard/led.png)
-[docs/breadboard/led.fzz](breadboard/led.fzz)
+![docs/breadboard/led-13.png](breadboard/led-13.png)  
+[Fritzing diagram: docs/breadboard/led-13.fzz](breadboard/led-13.fzz)
+![docs/breadboard/led-resistor.png](breadboard/led-resistor.png)  
+[Fritzing diagram: docs/breadboard/led-resistor.fzz](breadboard/led-resistor.fzz)
```
rwaldron: :white_check_mark:

Gruntfile.js
Resseguie: Just an extra newline. I was trying to make it generate as close to the previous versions as possible.

PR #706, opened 2015-04-08
Remove broken links for missing fritzing images

leobalter:


leobalter:
ok, this will be generated either way through grunt examples. I'm in a dead end here. Maybe the solution is to find a fritzing image for the given examples.

PR #709, opened 2015-04-08
move grove-led image to grove-led.js eg

Resseguie:


tpl/programs.json
```diff
@@ -606,7 +606,10 @@
     "examples": [
       {
         "file": "grove-led.js",
-        "title": "Grove - LED"
+        "title": "Grove - LED",
+        "images": [
+          {"file": "grove-led.jpg", "title": "Grove LED"}
```
rwaldron: This is ok

tpl/programs.json
```diff
@@ -621,7 +624,6 @@
         "title": "Grove - Sensor",
         "images": [
           {"file": "grove-shield.jpg", "title": "Grove Shield"},
-          {"file": "grove-led.jpg", "title": "Grove LED"},
```
rwaldron: This is not ok

PR #670, opened 2015-03-24
breakout LED-related code into led/ folder; closes #668

boneskull:
- Moved all the things
- Should be completely backwards-compatible (tests pass; removed files were replaced)
- Maintained private state between `Led`, `Led.RGB` and `Led.Array`.  (This probably merits some discussion, but I didn't want to break anything)

PR #710, opened 2015-04-09
Led: Clear interval after animation

scottgonzalez:
Animation intervals weren't cleared until another animation started or `.stop()` was called. As a result, once the animation completed, the interval would keep running and the `step` and `complete` callbacks would be called repeatedly.

### DIFF
--- a/lib/led/led.js
+++ b/lib/led/led.js
@@ -338,9 +338,10 @@ Led.prototype.animate = function(opts) {
     opts.step(delta);
 
     if (progress === 1) {
+      this.stop();
       opts.complete();
     }
-  }, opts.delay || 10);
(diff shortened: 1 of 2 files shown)


dtex:
Should there be a test in the extended group to catch possible regression? It could be a separate issue/PR.

scottgonzalez:
I'll update this tonight, now that #670 has landed.

rwaldron:
Thanks @scottgonzalez!

scottgonzalez:
That actually rebased cleanly :-)

rwaldron:
Needs a test that confirms stop() is called

scottgonzalez:
I added some basic tests. I'll flesh these out more in #724 after there's some discussion there.

PR #718, opened 2015-04-10
Led.RGB shorthand constructors

Resseguie:
now accepts:
Byte, Byte, Byte
[Byte, Byte, Byte]
{red: Byte, green: Byte, blue: Byte}

### DIFF
--- a/docs/led-rainbow.md
+++ b/docs/led-rainbow.md
@@ -19,7 +19,7 @@ var board = new five.Board();
 
 board.on("ready", function() {
   var rgb = new five.Led.RGB([6, 5, 3]);
-  var rainbow = ["FF000", "FF7F00", "00FF00", "FFFF00", "0000FF", "4B0082", "8F00FF"];
+  var rainbow = ["FF0000", "FF7F00", "00FF00", "FFFF00", "0000FF", "4B0082", "8F00FF"];
   var index = 0;
 
   setInterval(function() {
--- a/eg/led-rainbow.js
+++ b/eg/led-rainbow.js
@@ -3,7 +3,7 @@ var board = new five.Board();
 
 board.on("ready", function() {
   var rgb = new five.Led.RGB([6, 5, 3]);
-  var rainbow = ["FF000", "FF7F00", "00FF00", "FFFF00", "0000FF", "4B0082", "8F00FF"];
+  var rainbow = ["FF0000", "FF7F00", "00FF00", "FFFF00", "0000FF", "4B0082", "8F00FF"];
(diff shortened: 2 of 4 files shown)


Resseguie:
@scottgonzalez I modified the Led.RGB API here to allow three different forms:

``` js
Led.RGB(Byte, Byte, Byte);
Led.RGB([Byte, Byte, Byte]);
Led.RGB({
  pins: {
    red: Byte,
    green: Byte,
    blue: Byte
  }
});
```

You care to sync up the docs for that while you're doing the other LED doc updates?

Note that we're hoping to land #670 soon and we should verify that none of the class names change, etc.

scottgonzalez:
Sure. The format for listing constructor parameters is a bit awkward, though. Can you point to an existing page that shows a constructor with multiple parameters so I can see what it looks like?

Resseguie:
I'll be honest, I've never fully understood the structure on those listings. I've just always tried to kinda mimic best I could. Perhaps @rwaldron has a pointer to an example of how he'd prefer it be done.

Resseguie:
@rwaldron the tests I added for these are pretty simplistic. I just used doesNotThrow on the constructor of each form. That happens to work for the current implementation because it tries to call `isOn()` internally and if the pins aren't mapped correctly it blows up. But my simple approach to tests might not protect against regression if there is a fundamental change in the implementation sometime in the future.

Perhaps @divanvisagie has a good suggestion for an improved test that verifies that all the forms actually produce comparable LED objects? (See the new `params` test in `tests/led.js`)

scottgonzalez:
Checking for `led.red`, `led.green`, and `led.blue` having the correct pin values should be sufficient to verify that the basic functionality is working.

Resseguie:
Yeah, I guess I was making it too complicated. That is simple and a nice improvement. Good call.

``` js
    test.equal(led.red.pin, 9);
    test.equal(led.green.pin, 10);
    test.equal(led.blue.pin, 11);
```

scottgonzalez:
At this point, you probably don't need to `test.doesNotThrow()` assertions, right?

Resseguie:
Right. I left it in because the previous implementation did throw, so it would never have gotten to the equals tests. I guess either way it fails. No strong opinion leaving it in or taking it out.

rwaldron:
@Resseguie I just landed your `boneskull-led-refactor` branch... is this separate? Was I supposed to land this first?

Resseguie:
@rwaldron you're good. I had put the changes in both places so you could go either way.

PR #720, opened 2015-04-10
Pass callback to next iteration of .pulse()

Resseguie:
The callback was only called the first time the pulse completed.


Resseguie:
How do you test that a callback is called repeatedly?

Resseguie:
See #721 instead to avoid a messy merge.

PR #721, opened 2015-04-10
Pass callback to next iteration of .pulse()

Resseguie:


Resseguie:
How should we test that a callback is called repeatedly in cases like this?

rwaldron:
LGTM

Issue #715, opened 2015-04-09
Document `Led#animate()`


scottgonzalez:
This method isn't currently documented, but it seems like it's probably public API. If this *is* intended to be public API, I can write up the documentation. I'm also wondering if the API should be a little more user-friendly:

* Make `duration` optional, defaulting to `1000`, just like `.pulse()` and `.fade()`.
* Optional `complete` callback, also like `.pulse()` and `.fade()`.

There's also an existing TODO in the code to give the `step` callback a default value. Perhaps just make the default increase the brightness to `0xff`? I'm not sure that having a default for `step` really makes sense though.

scottgonzalez:
Just adding a note for myself to check the callback implementation for all the animation methods. If `complete` becomes optional, we can simplify some code in methods like `.fade()`.

rwaldron:
> This method isn't currently documented, but it seems like it's probably public API. 

Yes

> If this is intended to be public API, I can write up the documentation

That would be awesome :)



> Make duration optional, defaulting to 1000, just like .pulse() and .fade().

Yes

> Optional complete callback, also like .pulse() and .fade().

Yes

> There's also an existing TODO in the code to give the step callback a default value. Perhaps just make the default increase the brightness to 0xff? I'm not sure that having a default for step really makes sense though.

I haven't given it much thought, so I will defer to @Resseguie (and whoever else has valuable input)

Resseguie:
Do we know of any case where .animate() is used directly as part of the public API?

rwaldron:
> Do we know of any case where .animate() 

No, but it's exposed on `Led` instances and safe enough for users that want to roll their own goodies. No harm, right?

Resseguie:
No, it's just never been intuitive to me the relationship between delta/step/etc at a cursory glance. But because I'm slow doesn't mean it shouldn't be there. ;)

scottgonzalez:
Maybe some docs will clear that up :wink:

Resseguie:
Re: switching to the Animate implementation instead of the separate one in led.js: #730

dtex:
Since this issue was originally opened we've changed how animate works and even removed this method from led. Do we want to add it back? If not we can close this.

PR #724, opened 2015-04-10
Led animate args

scottgonzalez:
Ref #715

I split this into three commits to make it easier to back out changes if we don't want all of these.

* f3655b650ead829fbf6de86006c0707184a4f032 implements the defaults agreed upon in #715.
* c5f3e879ee9258ef8c58c6ede35ff56ae93791f0 removes the TODO about having a default `step`. If you're calling `.animate()`, you should define what animation you want.
* 4385875405da5976d70bfcb96a68b533321bc1bd removes the explicit default duration for methods that use `.animate()`, falling back to the new default inside `.animate()` (same value as before).

I'll squash these once there's agreement on what should stay. This will need to be rebased once #710 lands anyway.

### DIFF
--- a/lib/led/led.js
+++ b/lib/led/led.js
@@ -366,7 +366,7 @@ Led.prototype.pulse = function(time, callback) {
 
   if (typeof time === "function") {
     callback = time;
-    time = 1000;
+    time = null;
   }
 
   var step = function(delta) {
@@ -391,7 +391,7 @@ Led.prototype.pulse = function(time, callback) {
   }.bind(this);
 
   return this.animate({
-    duration: time || 1000,
+    duration: time,
     complete: complete,
     step: step
   });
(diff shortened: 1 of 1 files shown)


rwaldron:
https://github.com/rwaldron/johnny-five/commit/1bd8b9cd10bfe50ec91da0db89d44d08b2f2be3f

lib/led/led.js
```diff
@@ -427,7 +432,7 @@ Led.prototype.fade = function(val, time, callback) {
   }.bind(this);
 
   return this.animate({
-    duration: time || 1000,
+    duration: time,
     complete: complete,
     step: step
   });
```
rwaldron: Needs follow up commit with tests. cc @scottgonzalez

scottgonzalez: I've got a slew of tests I plan to write :-) I was just waiting on #710 to be merged first.

PR #726, opened 2015-04-11
Wii: Remove check for opts existing before use

scottgonzalez:
There are uses of `opts` which require it to exist prior to these lines, so the guard isn't necessary.

### DIFF
--- a/lib/wii.js
+++ b/lib/wii.js
@@ -73,8 +73,8 @@ function Wii(opts) {
   this.freq = opts.freq || 500;
 
   // Button instance properties
-  this.holdtime = opts && opts.holdtime || 500;
-  this.threshold = opts && opts.threshold || 10;
+  this.holdtime = opts.holdtime || 500;
+  this.threshold = opts.threshold || 10;
 
   // Initialize components
   device.initialize.call(this);


rwaldron:
I'm sure there's a lot of weird shit in this class...

scottgonzalez:
I'm not sure how much I can clean up in here right now, since I'm still very new to both microcontrollers and J5, but I can go through and see if there's more general stuff like this.

I have noticed some stuff since filing this PR that I'm not sure if you want to keep just for clarity. For example:

```js
Update.button.call(
  this.up, (decodeByte(data[5]) & 0x01) === 0 ? true : false
);
```

Would you prefer dropping the ternary since the equality check alone provides the same values?

Also, this line exists inside a `setInterval()`:

```js
this.io.sendI2CReadRequest(address, bytes, data.bind(this));
```

Do you want the clarity that `data` is being bound or can the binding move out of the timer callback so that it's not happening on every iteration?

rwaldron:
> Would you prefer dropping the ternary since the equality check alone provides the same values?

Yes

The next part is a bit more involved. That code needs to be updated to use the new I2C API. I'll take a longer look later and write up some notes about what needs to change there.

PR #733, opened 2015-04-15
support time/rate in min/max/center

### LINKED ISSUE #731: Accept time for servo.min/max/center so move:complete is fired
The `move:complete` event is only fired for timed moves, but `.min()`, `.max()` and `.center()` don't accept a time parameter.

Resseguie:
Fixes #731 
Added `rate` too


rwaldron:
Nice!

PR #734, opened 2015-04-15
Fix RGB.prototype.stop()

Resseguie:
Also removes duplicate code left from led refactor.
Adds tests for both stop and toggle to Led and Led.RGB


lib/led/led.js
```diff
@@ -493,303 +493,6 @@ Led.prototype.stop = function() {
...
-    this.toggle();
-  }.bind(this), rate || 100);
-
-  return this;
-};
-
-Led.RGB.prototype.blink = Led.RGB.prototype.strobe;
-Led.RGB.prototype.toggle = Led.prototype.toggle;
-Led.RGB.prototype.stop = Led.prototype.stop;
-
```
rwaldron: Whoa. How'd that happen?

PR #740, opened 2015-04-19
stepper max per board; err if no dir; tests

### LINKED ISSUE #736: Should throw error when trying to Stepper.step() when Stepper.direction is undefined
Right now you can call step() on a Stepper when its direction is undefined, without getting any errors. I spend a few hours trying to figure out why my stepper motor wasn't working, and it was because the direction was undefined. Stepper motors are nontrivial to wire so it would be nice to be alerted when code is broken.

I added a small warning in the Stepper wiki.

Thank you for your work on this awesome module!

### LINKED ISSUE #739: Stepper's MAXSTEPPERS should be per board
The Stepper component defines a MAXSTEPPERS value (6) and says "correlates with MAXSTEPPERS in firmware".

But that value should be per board and not global.

Resseguie:
Fixes #739 - limits max steppers (currently 6) per board instead of global
Fixes #736 - returns error object to callback function

### DIFF
--- a/lib/stepper.js
+++ b/lib/stepper.js
@@ -1,9 +1,9 @@
-var Board = require("../lib/board.js"),
-  events = require("events"),
-  util = require("util");
+var Board = require("../lib/board.js");
+var events = require("events");
+var util = require("util");
 
-var priv = new Map(),
-  steppers = [];
+var priv = new Map();
+var steppers = new Map();
 
 var MAXSTEPPERS = 6; // correlates with MAXSTEPPERS in firmware
 
@@ -60,7 +60,7 @@ function isSupported(io) {
  *  type: five.Stepper.TYPE.DRIVER
  *  stepsPerRev: number,
  *  pins: {
- *    step: number
+ *    step: number,
  *    dir: number
  *  }
  * });
@@ -109,7 +109,7 @@ function isSupported(io) {
  */
 
 function Stepper(opts) {
-  var initial, params = [];
+  var state, params = [];
 
   if (!(this instanceof Stepper)) {
     return new Stepper(opts);
@@ -134,10 +134,11 @@ function Stepper(opts) {
   if (!opts.stepsPerRev) {
     throw new Error(
       "Stepper requires a `stepsPerRev` number value"
-    );
+      );
   }
 
-  this.id = steppers.length;
+  steppers.set(this.board, steppers.get(this.board) || []);
+  this.id = steppers.get(this.board).length;
 
   if (this.id >= MAXSTEPPERS) {
     throw new Error(
@@ -188,7 +189,7 @@ function Stepper(opts) {
   if (opts.type === Stepper.TYPE.DRIVER) {
     if (!this.pins.dir || !this.pins.step) {
       throw new Error(
-        "Stepper.TYPE.DRIVER expects: pins.dir, pins.step"
+        "Stepper.TYPE.DRIVER expects: `pins.dir`, `pins.step`"
       );
     }
 
@@ -200,7 +201,7 @@ function Stepper(opts) {
   if (opts.type === Stepper.TYPE.TWO_WIRE) {
     if (!this.pins.motor1 || !this.pins.motor2) {
       throw new Error(
-        "Stepper.TYPE.TWO_WIRE expects: pins.motor1, pins.motor2"
+        "Stepper.TYPE.TWO_WIRE expects: `pins.motor1`, `pins.motor2`"
       );
     }
 
@@ -212,7 +213,7 @@ function Stepper(opts) {
   if (opts.type === Stepper.TYPE.FOUR_WIRE) {
     if (!this.pins.motor1 || !this.pins.motor2 || !this.pins.motor3 || !this.pins.motor4) {
       throw new Error(
-        "Stepper.TYPE.FOUR_WIRE expects: pins.motor1, pins.motor2, pins.motor3, pins.motor4"
+        "Stepper.TYPE.FOUR_WIRE expects: `pins.motor1`, `pins.motor2`, `pins.motor3`, `pins.motor4`"
       );
     }
 
@@ -231,15 +232,31 @@ function Stepper(opts) {
 
   this.io.stepperConfig.apply(this.io, params);
 
-  steppers.push(this);
+  steppers.get(this.board).push(this);
 
-  initial = Step.PROPERTIES.reduce(function(state, key, i) {
+  state = Step.PROPERTIES.reduce(function(state, key, i) {
     return (state[key] = typeof opts[key] !== "undefined" ? opts[key] : Step.DEFAULTS[i], state);
   }, {
-    isRunning: false
+    isRunning: false,
+    type: opts.type,
+    pins: this.pins
   });
 
-  priv.set(this, initial);
+  priv.set(this, state);
+
+  Object.defineProperties(this, {
+    type: {
+      get: function() {
+        return state.type;
+      }
+    },
+
+    pins: {
+      get: function() {
+        return state.pins;
+      }
+    }
+  });
 }
 
 Object.defineProperties(Stepper, {
@@ -368,9 +385,9 @@ Stepper.prototype.step = function(stepsOrOpts, callback) {
 
   isValidStep = true;
 
-  function failback() {
+  function failback(error) {
     isValidStep = false;
-    callback();
+    if (callback) { callback(error); }
   }
 
   params.push(steps);
@@ -400,13 +417,17 @@ Stepper.prototype.step = function(stepsOrOpts, callback) {
 
   if (steps === 0) {
     failback(
-      new Error("Must set a number of steps")
+      new Error(
+        "Must set a number of steps when calling `step()`"
+      )
     );
   }
 
   if (step.direction < 0) {
     failback(
-      new Error("Must set a direction")
+      new Error(
+        "Must set a direction before calling `step()`"
+      )
     );
   }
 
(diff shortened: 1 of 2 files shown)


Resseguie:
@rwaldron A couple discussion points on this PR...

1) I don't have hardware, so I just implemented against (new) tests. Need someone with hardware to verify
2) Are you okay with exposing type and pins as properties like I did so that I can access them in tests? (Is there a better way to do this?)
3) I wanted to test that the callback to `.step()` was executed when successful, but apparently the mock-firmata doesn't actually call the callback? Again, making assumptions here since I don't have hardware. Instead, I just tested that the callback WASN'T called with an error. See "TODO" in tests.

lib/stepper.js
```diff
@@ -411,6 +428,7 @@ Stepper.prototype.step = function(stepsOrOpts, callback) {
   }
 
   if (isValidStep) {
+    console.log("here");
```
scottgonzalez: stray `console.log()`

test/stepper.js
```diff
@@ -62,6 +62,349 @@ exports["Stepper Firmware Requirement"] = {
...
 
+exports["Stepper - constructor"] = {
+  setUp: function(done) {
+
+    this.board = new Board({
+      io: new MockFirmata({
+        pins: [
+          {
+            supportedModes: [8],
+          },
```
scottgonzalez: extra comma; previous line too

test/stepper.js
```diff
@@ -62,6 +62,349 @@ exports["Stepper Firmware Requirement"] = {
...
+      repl: false
+    });
+
+    done();
+  },
+  tearDown: function(done) {
+    done();
+  },
+
+  infertTypeAmbiguous: function(test) {
```
scottgonzalez: typo

test/stepper.js
```diff
@@ -62,6 +62,349 @@ exports["Stepper Firmware Requirement"] = {
...
+  setUp: function(done) {
+
+    this.maxSteppers = 6;
+
+    this.board1 = new Board({
+      io: new MockFirmata({
+        pins: [
+          {
+            supportedModes: [8],
+          },
```
scottgonzalez: extra commas again

test/stepper.js
```diff
@@ -62,6 +62,349 @@ exports["Stepper Firmware Requirement"] = {
...
+      debug: false,
+      repl: false
+    });
+
+    this.board2 = new Board({
+      io: new MockFirmata({
+        pins: [
+          {
+            supportedModes: [8],
+          },
```
scottgonzalez: extra commas again

test/stepper.js
```diff
@@ -62,6 +62,349 @@ exports["Stepper Firmware Requirement"] = {
...
+
+exports["Stepper - set direction required before step"] = {
+  setUp: function(done) {
+
+    this.board = new Board({
+      io: new MockFirmata({
+        pins: [
+          {
+            supportedModes: [8],
+          },
```
scottgonzalez: extra commas again

test/stepper.js
```diff
@@ -62,6 +62,349 @@ exports["Stepper Firmware Requirement"] = {
...
+  tearDown: function(done) {
+    done();
+  },
+
+  directionSet: function(test) {
+    test.expect(1);
+
+    // If we haven't set a direction, we should get error
+    this.stepper.step(1, function(error) {
+      test.ok(!!error);
```
scottgonzalez: Is there a type or code or message to check for?

test/stepper.js
```diff
@@ -62,6 +62,349 @@ exports["Stepper Firmware Requirement"] = {
...
+    });
+
+    // after setting direction, we shouldn't
+    this.stepper.cw();
+    this.stepper.step(1, function(error) {
+      if (!!error) {
+        // Testing that we shouldn't get here.
+        // If we do, expected number of tests will fail
+        test.ok(false, "should not have gotten error after setting direction");
```
scottgonzalez: Could just be `test.ok(!error, "..." )` instead of a conditional, but I haven't looked through other tests to see what pattern is used.

test/stepper.js
Resseguie: Good question. The error message returned is just "Must set a direction" so I could use that to check the specific error I'm interested in. But do I really want to assume a specific error wording? Maybe so, and another issue can address setting proper error codes?

test/stepper.js
Resseguie: Not sure about pattern elsewhere, but I like your suggestion. I arrived at this after testing several different ways and didn't notice the obvious (now that you point it out) simplification. I'll make this change, but we might want to handle it differently based on my previous comment regarding error codes.

What I'm specifically trying to test is that it throws a "must set direction" error if you don't call `.cw()` or `.ccw()` before `.step()` but doesn't throw that specific error after a direction is set.

I'd like this test to check for that exact error condition and let other tests handle other errors that might be thrown from `.step()`

thoughts?

Resseguie:
@scottgonzalez thanks for all the comma notes, too. Obviously I did a lot of cut-n-pasting on those tests. :) I wish `grunt jshint` warned you on those. Is there a config flag we're missing for that?

scottgonzalez:
> I wish `grunt jshint` warned you on those. Is there a config flag we're missing for that?

Nope. Trailing commas are valid in ES5, so unless the `es3` option is used, it won't be caught. There's no option for this because trailing commas are purely stylistic, and JSHint has decided to remove all style-related options. This can, however, be built as an extension, but so far I'm not aware of anyone doing that.

test/stepper.js
scottgonzalez: Right now, testing the error message is the best option if you don't want to change the code that's throwing the error. Since error message should never be localized, it's actually a fairly safe option. Any change to the error message would require a change to the tests, but those should be trivial.

A common pattern in node is to assign a `code` property which contains a string which can be used similar to a constant. For example, `"EISDIR"` and `"ENOENT"` are common error codes from the `fs` module. This is generally nicer to work with in JS rather than custom `Error` subclasses. This can make the code a bit verbose since you can't just write `callback(new Error("..."))`; I've dealt with that before by creating a utility method: https://github.com/scottgonzalez/debt/blob/13d43f3b116df7e769cefaf58f5d8d0e475c64c2/lib/util.js#L100

Resseguie: Got it. Let me take a crack at going ahead and putting that in while I'm at it.

Resseguie: @scottgonzalez when using that `error.code` technique, instead of using `test.throws` do you just use a try/catch and then `test.equals` to check the code value? Or am I missing a cleaner method?

Resseguie:
> I wish grunt jshint warned you on those. Is there a config flag we're missing for that?

Oh wait, we're using jscs and you can `disallowTrailingComma` there if we wanted. But they're all over, so I'll leave that to a separate issue. (And I don't know @rwaldron's preference for that anyway.)

Resseguie:
@scottgonzalez I think I got all your suggestions into that new commit.

test/stepper.js
scottgonzalez: Yes, what you're doing is exactly what I do in my tests.

lib/fn.js
```diff
@@ -185,5 +185,20 @@ Fn.int16 = Fn.ToInt16FromTwoBytes = function(high, low) {
...
+  if ( settings.code ) {
+    message = settings.code + ": " + message;
+  }
+  delete settings.message;
+
+  var error = new Error( message );
+  Fn.extend( error, settings );
+
+  return error;
+};
```
scottgonzalez: Should probably remove all the spacing to match the exiting style.

Resseguie:
@hail-seitan did a hardware sanity check for me on this and didn't run into problems.

@scottgonzalez feel free to sanity check that last commit.

@rwaldron if you're okay with question 2 in:
https://github.com/rwaldron/johnny-five/pull/740#issuecomment-94229954
then this should be good to go.

test/stepper.js
rwaldron: trailing commas are ok ;)

rwaldron:
>  I wish grunt jshint warned you on those.

It never will. Trailing commas are fine.

rwaldron:
Don't hate me for this, but I don't want the `__.createError` changes in this patch. That's a completely unrelated issue that needs more discussion. Make no mistake: I would love to standardize the all of the errors throughout the whole project, but that's a different issue and patchset.

Resseguie:
I don't have a strong opinion either way on that. Do you want stepper to just throw generic Error with string as before, or are you okay with using the error codes, but keeping it local to stepper.js?

Is there a clean way to back out those couple commits here? Or just one more commit that removes? Then squash maybe? I just never get that to work right.

Resseguie:
Or I can just do a new PR that's clean.

lib/stepper.js
```diff
@@ -1,9 +1,10 @@
...
-  util = require("util");
+var Board = require("../lib/board.js");
+var __ = require("../lib/fn.js");
+var events = require("events");
+var util = require("util");
 
-var priv = new Map(),
-  steppers = [];
+var priv = new Map();
+var steppers = new Map();
```
rwaldron: If this is changed to a Map, then occurrences of `steppers` needs to be changed accordingly: https://github.com/Resseguie/johnny-five/blob/stepper-test/lib/stepper.js#L144-L145

lib/stepper.js
```diff
@@ -231,15 +241,31 @@ function Stepper(opts) {
 
   this.io.stepperConfig.apply(this.io, params);
 
-  steppers.push(this);
+  steppers[this.board.id].push(this);
 
   initial = Step.PROPERTIES.reduce(function(state, key, i) {
```
rwaldron: Rename `initial` to `state`

Resseguie:
@rwaldron Ok, all feedback incorporated (Map usage fix, s/initial/state/, backed out error codes) and squashed.

PR #741, opened 2015-04-20
[WIP] Progress check on Piezo Changes

mcdonnelldean:
@lyzadanger @Resseguie @julianduque @BrianGenisio 

Hi all,

I have tidied up this PR some what clarifying what is proposed. Currently the new public API is
fully implemented but needs more tests. There is a lump of logic that we may be able to remove
from the internal API if we include Octavian to handle all logic except the handling of melodies.
## Public API

The new public API assumes that we are using Octavian (talked about below). By using Octavian the 
new api becomes drastically simplified. The same constructor is retained and allows the provision
of default values.
- Piezo(opts)
- .pulsepin(pulse, duration)
- .playFrequency(frequency, duration)
- .playNote(note, duration)
- .playMelody(melody, callback)
- .stop()
## RTTF

Ring Tone Transfer Format or RTTF is a language that was designed for typing melodies into 
cellphones via a keypad. It's simple format and pre-existing database of sounds make this the
perfect input format to enable more advanced melodies to be played via the Piezo class.
### Sample RTTF Input

```
'd=8 o=4 b=125: f, a, 4b, f, a, 4b, f, a, 4b, e5, 4d5, b, c5, b, g, 2e, a, b, g, 2e, p2'
```

The input string has a header section, broken down as follows:
- _d=_ - The default beat for notes who's beat is not specified.
- _o=_ - The default octave for notes who's octave is not specified.
- _b=_ - The overall tempo of the melody.

After the colon each note in the melody is specified with at least the note value and optionally
the octave and beat, where this is not provided the defaults are used.

RTTF Has some more advanced features that need to be built in, including repeating and half notes,
these are not currently supported but will be before the PR goes live.
### RTTF+

```
'b=8 o=4 t=125: f a 4b f a 4b f a 4b e5 4d5 b c5 b g 2e a b g 2e p2
```

RTTF+ is a simplification of RTTF, it removes the need to use comma's as well as fixes the header
naming scheme.
## Octavian

[See #809] [Octavian](https://github.com/stevekinney) is a lib written by  [Steve Kinney](https://github.com/stevekinney). It manages Notes and all of the infrastructure around 
them. Usage of Octavian would most likely rest on having the following features implemented.
- RTTF input support for notes
- Beats scale, beats on notes, beat normalization
- (optionally) Pulse calculation

Because of the logic around beats and parsing we would need the first two supported as a minimum
otherwise we will need to add more parsing logic to handle RTTF to Octavian format.

@SteveKinney It would be awesome to discuss if any of these changes would make sense for your
library. I can help with the changes but it's conditional on a roughly 3 week lead time as I
am swamped. I'm available for skype or I.M/mail though to help whenever.

### DIFF
--- a/lib/piezo.js
+++ b/lib/piezo.js
@@ -1,6 +1,5 @@
 var Board = require("../lib/board.js"),
-  Timer = require("nanotimer"),
-  _ = require("lodash");
+  Timer = require("nanotimer");
 
 function Piezo(opts) {
   Board.Component.call(this, opts = Board.Options(opts));
@@ -158,7 +157,7 @@ function normalizeBeat(beat) {
   }
 
   var normalizedBeat = null;
-  _.forEach(Object.keys(Piezo.Beats), function(beatElement) {
+  Object.keys(Piezo.Beats).forEach(function(beatElement) {
     if (beatElement === beat.toString()) {
       normalizedBeat = Piezo.Beats[beatElement];
     }
@@ -254,30 +253,31 @@ function normalizeMelody(melody, defaults) {
   melody.sheet = melody.sheet.replace(/\s{1,}=|=\s{1,}/g, "=");
 
   // Ensure we have defaults properly set up.
-  melody.notes = melody.notes = [];
+  melody.notes = melody.notes || [];
   melody.octave = melody.octave || defaults.octave;
   melody.beat = melody.beat || defaults.beats;
   melody.tempo = melody.tempo || defaults.tempo;
 
   // The sheet can contain defaults for beat (b=), octave (o=),
   // and tempo (t=). They always come before the first note.
-  var sheetElements = melody.sheet.match(/\S+/g) || [];
-  _.forEach(sheetElements, function(element) {
-    if ((element.indexOf("b=") > -1 && melody.sheet.indexOf("t=") > -1) ||
-        (element.indexOf("d=") > -1 && melody.sheet.indexOf("b=") > -1)) {
+  var sheet = melody.sheet;
+  var sheetElements = sheet.match(/\S+/g) || [];
+  sheetElements.forEach(function(element) {
+    if ((element.includes("b=") && sheet.includes("t=")) ||
+        (element.includes("d=") && sheet.includes("b="))) {
       var beatValue = element.substring(2);
       var beat = beatValue || melody.beat;
 
       return melody.beat = normalizeBeat(beat);
     }
 
-    if (element.indexOf("o=") > -1) {
+    if (element.includes("o=")) {
       var octaveValue = element.substring(2);
       return melody.octave = normalizeOctave(octaveValue) || melody.octave;
     }
 
-    if ((element.indexOf("t=") > -1 && melody.sheet.indexOf("b=") > -1) ||
-        (element.indexOf("b=") > -1 && melody.sheet.indexOf("d=") > -1)) {
+    if ((element.includes("t=") && sheet.includes("b=")) ||
+        (element.includes("b=") && sheet.includes("d="))) {
       var tempoValue = element.substring(2);
       return melody.tempo = normalizeTempo(tempoValue) || melody.tempo;
     }
@@ -307,10 +307,7 @@ function frequencyToPulse(frequency) {
 }
 
 function noteToPulse(note) {
-  var frequency = noteToFrequency(note);
-  var pulse = frequencyToPulse(frequency);
-
-  return pulse;
+  return frequencyToPulse(noteToFrequency(note));
 }
 
 function noteToFrequency(note) {
@@ -411,10 +408,15 @@ Piezo.prototype.playMelody = function(melody, callback) {
 };
 
 Piezo.prototype.stop = function() {
-  this.io.digitalWrite(this.pin, 0);
   clearTimer.call(this);
 
   return this;
+};
+
+Piezo.prototype.off = function() {
+  this.io.digitalWrite(this.pin, 0);
+
+  return this.stop();
 };
 
 module.exports = Piezo;
--- a/lib/piezo.js
+++ b/lib/piezo.js
@@ -1,150 +1,398 @@
 var Board = require("../lib/board.js"),
-  Timer = require("nanotimer");
+    Timer = require("nanotimer"),
+    _ = require("lodash");
 
 function Piezo(opts) {
+  Board.Component.call(this, opts = Board.Options(opts));
 
-  Board.Component.call(
-    this, opts = Board.Options(opts)
-  );
-
-  // Hardware instance properties
   this.mode = this.io.MODES.OUTPUT;
   this.pin = opts.pin || 3;
-
-  this.io.pinMode(this.pin, this.mode);
-
-  // Piezo instance properties
   this.isPlaying = false;
-}
 
-// These notes are rounded up at .5 otherwise down.
-Piezo.Notes = {
-  "c0": 16,
-  "c#0": 17,
-  "d0": 18,
-  "d#0": 19,
-  "e0": 21,
-  "f0": 22,
-  "f#0": 23,
-  "g0": 25,
-  "g#0": 26,
-  "a0": 28,
-  "a#0": 29,
-  "b0": 31,
-  "c1": 33,
-  "c#1": 35,
-  "d1": 37,
-  "d#1": 39,
-  "e1": 41,
-  "f1": 44,
-  "f#1": 47,
-  "g1": 49,
-  "g#1": 52,
-  "a1": 55,
-  "a#1": 58,
-  "b1": 62,
-  "c2": 65,
-  "c#2": 69,
-  "d2": 73,
-  "d#2": 78,
-  "e2": 82,
-  "f2": 87,
-  "f#2": 93,
-  "g2": 98,
-  "g#2": 104,
-  "a2": 110,
-  "a#2": 117,
-  "b2": 124,
-  "c3": 131,
-  "c#3": 139,
-  "d3": 147,
-  "d#3": 156,
-  "e3": 165,
-  "f3": 175,
-  "f#3": 185,
-  "g3": 196,
-  "g#3": 208,
-  "a3": 220,
-  "a#3": 233,
-  "b3": 247,
-  "c4": 262,
-  "c#4": 277,
-  "d4": 294,
-  "d#4": 311,
-  "e4": 330,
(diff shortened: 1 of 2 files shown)


lib/piezo.js
```diff
@@ -1,150 +1,398 @@
...
+  "f8":  5588,
   "f#8": 5920,
-  "g8": 6272,
+  "g8":  6272,
   "g#8": 6645,
-  "a8": 7040,
+  "a8":  7040,
   "a#8": 7459,
-  "b8": 7902
+  "b8":  7902
```
rwaldron: Just run the file through `grunt jsbeautifier`

rwaldron:
A lot of time and discussion between several contributors went into the existing API. I urge you to work with them on this.

lib/piezo.js
```diff
@@ -1,150 +1,398 @@
...
-function clearTimer() {
-  if (!this.timer) {
-    return;
+Piezo.Beats = {
+  "second":   ["2", 2],
+  "fourth":   ["4", 1],
+  "eight":    ["8", 1/2],
+  "sixteenth":["16", 1/4]
+};
+
```
rwaldron: Why is the first value a string of a number?

lib/piezo.js
```diff
@@ -1,150 +1,398 @@
...
+  // Ensure we have defaults properly set up.
+  melody.notes = melody.notes = [];
+  melody.octave = melody.octave || defaults.octave;
+  melody.beat = melody.beat || defaults.beats;
+  melody.tempo = melody.tempo || defaults.tempo;
+
+  // The sheet can contain defaults for beat (b=), octave (o=),
+  // and tempo (t=). They always come before the first note.
+  var sheetElements  = melody.sheet.match(/\S+/g) || [];
+  _.forEach(sheetElements, function(element) {
```
rwaldron: Is sheetElements an Array? If yes, then use sheetElements.forEach

lib/piezo.js
```diff
@@ -1,150 +1,398 @@
...
+    var octavedNote = note.notation + note.octave;
+    if (Piezo.Notes[octavedNote]) {
+      frequency =  Piezo.Notes[octavedNote];
+    }
+  }
+
+  return frequency;
+};
+
+Piezo.prototype.pulsePin = function(pulse, duration) {
```
rwaldron: Can this be just `pulse`?

lib/piezo.js
```diff
@@ -1,150 +1,398 @@
...
+  this.isPlaying = true;
+  playNextNote();
+};
+
+Piezo.prototype.clearTimer = function() {
+  this.io.digitalWrite(this.pin, 0);
+  clearTimer.call(this);
+
+  return this;
+};
```
rwaldron: Why is this exposed?

lib/piezo.js
```diff
@@ -1,150 +1,398 @@
...
+        normalizedNote.duration = note[0];
+        normalizedNote.notation = note[1];
+      }
+
+      return normalizedNote;
+    }
+  }
+
+  return null;
+};
```
rwaldron: This is basically a constructor. Why not just define a private Note class?

McDonnellDean:
@rwaldron absolutely! That's why ai have it here in the rough and why I kept the old api in tact. I'm open to any suggestions even keeping the existing bits and merging functionality. I'd love to get some feedback and this early before getting into finalising anything.

rwaldron:
I'll take a closer look in the AM, but everyone else should review as well.

McDonnellDean:
Oh no! I'm missing some sample code and a description of pulsePin() which is really just note() named more like its function. I'll fix this up first thing in the morning.

lib/piezo.js
McDonnellDean: I incorrectly assumed I couldn't have the number part as the first part. Will change

lib/piezo.js
McDonnellDean: Done

lib/piezo.js
McDonnellDean: yes, it makes more sense like this because it's opposite is just called stop

lib/piezo.js
McDonnellDean: Wrong name, should be stop

lib/piezo.js
McDonnellDean: Is there a need for it? This is the only place it would be used

lib/piezo.js
```diff
@@ -129,22 +137,274 @@ Piezo.Notes = {
...
+    if (beatElement === beat.toString()) {
+      normalizedBeat = Piezo.Beats[beatElement];
+    }
+  });
+
+  return normalizedBeat;
+};
+
+Piezo.normalizeOctave = function(octave) {
+  if (octave === "" || isNaN(octave) || octave < 0 || octave > 8) {
```
rwaldron: `isNaN` => `Number.isNaN`

lib/piezo.js
```diff
@@ -129,202 +137,284 @@ Piezo.Notes = {
...
+    return null;
   }
-  if (typeof tone === "string") {
-    tone = tone.toLowerCase();
 
-    var needsOctave = isNaN(tone.substr(tone.length -1));
-    if (needsOctave) {
-      tone = tone + Piezo.prototype.defaultOctave().toString();
+  var normalizedBeat = null;
+  _.forEach(Object.keys(Piezo.Beats), function(beatElement) {
```
rwaldron: `_.forEach(Object.keys(Piezo.Beats)` => `Object.keys(Piezo.Beats).forEach(`

lib/piezo.js
```diff
@@ -129,202 +137,284 @@ Piezo.Notes = {
...
+  // Ensure we have defaults properly set up.
+  melody.notes = melody.notes = [];
+  melody.octave = melody.octave || defaults.octave;
+  melody.beat = melody.beat || defaults.beats;
+  melody.tempo = melody.tempo || defaults.tempo;
+
+  // The sheet can contain defaults for beat (b=), octave (o=),
+  // and tempo (t=). They always come before the first note.
+  var sheetElements = melody.sheet.match(/\S+/g) || [];
+  _.forEach(sheetElements, function(element) {
```
rwaldron: ` _.forEach(sheetElements` => `sheetElements.forEach(`

lib/piezo.js
```diff
@@ -129,202 +137,284 @@ Piezo.Notes = {
...
+    if ((element.indexOf("b=") > -1 && melody.sheet.indexOf("t=") > -1) ||
```
rwaldron: Use `element.includes()`

lib/piezo.js
```diff
@@ -129,202 +137,284 @@ Piezo.Notes = {
...
-  return this;
-};
+    if (element.indexOf("o=") > -1) {
+      var octaveValue = element.substring(2);
+      return melody.octave = normalizeOctave(octaveValue) || melody.octave;
+    }
+
+    if ((element.indexOf("t=") > -1 && melody.sheet.indexOf("b=") > -1) ||
+        (element.indexOf("b=") > -1 && melody.sheet.indexOf("d=") > -1)) {
+      var tempoValue = element.substring(2);
```
rwaldron: Use `element.includes()`

lib/piezo.js
```diff
@@ -129,202 +137,284 @@ Piezo.Notes = {
...
-    lastNote = song[song.length - 1];
-    if (lastNote && lastNote[0] === note) {
-      lastNote[1] += beats;
-    } else {
-      song.push([note, beats]);
+function noteToPulse(note) {
+  var frequency = noteToFrequency(note);
+  var pulse = frequencyToPulse(frequency);
+
+  return pulse;
```
rwaldron: `return frequencyToPulse(noteToFrequency(note));`

lib/piezo.js
```diff
@@ -129,202 +137,284 @@ Piezo.Notes = {
...
-  next();
+  playNextNote();
 
   return this;
 };
 
-Piezo.prototype.noTone = function() {
+Piezo.prototype.stop = function() {
   this.io.digitalWrite(this.pin, 0);
   clearTimer.call(this);
```
rwaldron: Let's change this to match `Led`: 

- `stop` just stops an internal timer, but doesn't change the state 
- `off` changes the state to off

It's a slight change from before, but we should be ok with breaking small things in the name of consistency.

lib/piezo.js
```diff
@@ -129,202 +137,284 @@ Piezo.Notes = {
...
+
+  // 1. remove colons and semi-colons
+  // 2. remove additional whitespace
+  // 3. clean up defaults if present.
+  melody.sheet = melody.sheet.replace(/,|:/g, " ");
+  melody.sheet = melody.sheet.replace(/\s{2,}/g, " ");
+  melody.sheet = melody.sheet.replace(/\s{1,}=|=\s{1,}/g, "=");
+
+  // Ensure we have defaults properly set up.
+  melody.notes = melody.notes = [];
```
rwaldron: `melody.notes = melody.notes = [];`? Should this be: `melody.notes = melody.notes || [];`?

test/piezo.js
```diff
@@ -181,222 +192,55 @@ exports["Piezo"] = {
...
-        tempo: tempo
-      };
-    test.expect(1);
-
-    this.piezo.play(tune, function() {
-      test.ok(1); // We made it this far, no choking on bad values
-      test.done();
-    }.bind(this));
-
-  },
```
rwaldron: I'm concerned about all of these deleted tests.

mcdonnelldean: There are a lot more tests to come, i just have the bear minimum to get the octvian work in.

lib/piezo.js
mcdonnelldean: Noted

lib/piezo.js
mcdonnelldean: Done as of next commit

lib/piezo.js
mcdonnelldean: Fixed on next commit

mcdonnelldean:
@All,

I'll get this PR cleaned up and finalised for review in the next 10 days.

rwaldron:
@Resseguie ^^

mcdonnelldean:
Ugh, I'm super sorry! I completely forgot this. So, is it still valid (apart from cleanup). I need to re familiarise myself with the code again but is the way this was done still the 'correct' way to do this? I've seen some talk around packs or plugins for J5.

PR #743, opened 2015-04-21
Proper license for fma

scottgonzalez:
All rights reserved is not a license ;-)

This code is actually just Simplified BSD.


scottgonzalez:
Feel free to close in favor of #744 if you want to fully comply with the license ;-)

Resseguie:
Ah, sorry. I think we both put in a PR about the same time.

PR #747, opened 2015-04-23
JSHint Updates

scottgonzalez:
* Use `.jshintrc` in Grunt config instead of hard-coding the configuration.
* Add `unused: true` to the JSHint settings.
  * Add `eg/.jshintrc` which is a copy of the main `.jshintrc` without `unused: true`.


lib/animation.js
```diff
@@ -211,12 +211,9 @@ Animation.prototype.speed = function(speed) {
  * play
  *
  * Start a segment
- *
- * @param  {Object} opts Options: reverse, loop, progress
- *
  */
 
-Animation.prototype.play = function(opts) {
+Animation.prototype.play = function() {
```
rwaldron: cc @dtex

lib/board.js
```diff
@@ -11,13 +11,10 @@ var __ = require("../lib/fn.js");
 var Repl = require("../lib/repl.js");
 var Options = require("../lib/board.options.js");
 var Pins = require("../lib/board.pins.js");
-// var temporal = require("temporal"),
-var IO;
```
rwaldron: Comment instead of remove. It's part of some other temporarily "off" code.

lib/led/ledcontrol.js
```diff
@@ -562,8 +559,8 @@ Controllers = {
       }, this);
       // NOTE: Currently unused, these will form
       // the basis for the `setup` constructor option
-      var setup = Object.assign({}, LedControl.DEFAULTS, opts.setup || {});
-      var keys = Object.keys(setup);
+      // var setup = Object.assign({}, LedControl.DEFAULTS, opts.setup || {});
+      // var keys = Object.keys(setup);
```
rwaldron: :+1:

rwaldron:
Looks awesome. Nice to have this crap cleaned up. I want to give @dtex a chance to take a look at those animation args, but otherwise this is good to go.

scottgonzalez:
Great. I've got the one modification in `lib/board.js` done. I'll wait for feedback from @dtex before pushing another commit though.

lib/animation.js
dtex: I think it was my intent to allow the user tomodify the currently running animation (duration, speed stuff like that). That never happened, and I can't think of a specific use case for it so I'm cool with this change.

eg/.jshintrc
```diff
@@ -0,0 +1,25 @@
...
+  "newcap": false,
+  "noarg": true,
+  "sub": true,
+  "undef": true,
+  "boss": true,
+  "eqnull": true,
+  "node": true,
+  "strict": false,
+  "esnext": true,
+  "quotmark": "double",
```
dtex: I could have sworn I already added this. My mind playing tricks on me I guess.

scottgonzalez:
Updated.

PR #748, opened 2015-04-24
Led: Cleanup

scottgonzalez:
Just some little stuff I noticed while reviewing `Led` to get an understanding of how the controller model works.


Resseguie:
LGTM

PR #752, opened 2015-04-25
Pintest

mMerlin:
Add more unit test for Pin class.  This has some failing tests that I need to cleanup.

What should the result be if tst = new Pin() or new Pin({}) is used?  The code (pin.js:30) sets pinValue to 0 for those cases, but that never gets back into this.pin or this.addr.  Should it?  Should an error be thrown by Pin()?  Right now, the result leaves the pin and addr properties as undefined, which my unit test complains about.

### DIFF
--- a/lib/pin.js
+++ b/lib/pin.js
@@ -26,6 +26,11 @@ function Pin(opts) {
   if (!(this instanceof Pin)) {
     return new Pin(opts);
   }
+  if (opts === undefined || (typeof opts === "object" &&
+      opts.addr === undefined && opts.pin === undefined)
+      ) {
+    throw new Error("Pins must have a pin number");
+  }
 
   var pinValue = typeof opts === "object" ? (opts.addr || opts.pin || 0) : opts;
   var isAnalogInput = Pin.isAnalog(opts);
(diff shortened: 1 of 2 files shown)


rwaldron:
This is awesome :)

> new Pin() or new Pin({})

These will throw an exception. 

> The code (pin.js:30) sets pinValue to 0 for those cases, but that never gets back into this.pin or this.addr. Should it

No, it shouldn't. `pinValue` is used to capture whatever the calling code actually passed (or didn't pass). That way Johnny-Five can better understand how to infer what kind of pin the calling code intended to create. 


You'll need to rebase this, as I just pushed changes to `lib/pin.js` and `test/pin.js`. It should be as simple as: 
- `git remote add upstream https://github.com/rwaldron/johnny-five.git` (if you haven't already)
- `git pull --rebase upstream master` 
- Resolve conflicts if there are any, commit and then push your `pintest` branch with `-f`

mMerlin:
Did the rebase, then added the code to handle no pin number.

lib/pin.js updated to throw exception when no pin number found
test/pin.js updated to expect exception when not pin number pass to Pin constructor

PR #756, opened 2015-04-27
Use Animation() class for Led.fade() and Led.pulse()

dtex:
* Changed some property names in animation to be more generic
```degrees``` became ```value```
* Created @@normalize and @@render methods for Led
* Removed LED’s custom animation feature
* Added the ability to pass in an opts object to pulse or fade (same
opts as Animation.enqueue)
* Tweaked Servo[“@@normalize”] to use new generic value instead of
degrees
* Updated animation tests to use new generic name
* Move time based Led tests to extended/led.js
* Removed redundant pulse() and fade() tests from PCA-9685 group

### CODE REVIEW

**Comment by Resseguie** [comment]
I like the concept of opts, just thinking through ordering of params (along with backwards compat) that is consistent elsewhere. I just made a quick scan through on my phone. I can look more in detail later.

**Comment by dtex** [comment]
Forgot to mention, this also means that instances of Led() can be used in Animation() targets.

**Comment by rwaldron** [comment]
Wow. Will need a bunch of reality testing, but this looks awesome :)

**Comment by dtex** [comment]
@rwaldron I'm really wanting to get your input on the args situation. Three possible in pulse and four in fade. Lot's of ugly ```if (typeof arg...``` stuff going on. I don't see an obvious way around it while maintaining backwards compatibility.

**Comment by rwaldron** [comment]
@dtex for `pulse`, instead of: `rate, callback, opts`, only allow one signature or the other.

```js
// Signature 1
// These are variations of the same:
led.pulse(500)
led.pulse(500, function() {});

// Signature 2
led.pulse({ rate: 500, onloop: function() { ... }, ... });
```

Same thing for all methods that are getting upgrades.

Resseguie:
That would be somewhat similar to `RGB.color()` except with a callback added in:
https://github.com/rwaldron/johnny-five/blob/master/lib/led/rgb.js#L102

dtex:
I've simplified the signatures per @rwaldron 's suggestion (thank you for the dose of sanity) and added a couple of examples.

Resseguie:
@dtex Don't forget to add those examples to programs.json so they'll get generated automatically. (`grunt examples`)

dtex:
Thanks for the reminder. It's my first time with the new example stuff.

docs/led-fade-animation.md
```diff
@@ -0,0 +1,72 @@
+<!--remove-start-->
+
+# LED - Fade with animation
+
+
+LED fade example that fades an LED up and down to varying values. Requires LED on pin that supports PWM (usually denoted by ~).
```
rwaldron: This is going to be lost when these are regenerated :\

dtex: I don't follow. This comes from programs.json

rwaldron: Haha. I'm **also** still getting used to the new programs.json ;)

Resseguie: I like to keep you on your toes.

rwaldron: The new way is better.

rwaldron:
I'm pretty sure these examples actually **look** nicer now. I just ran all the examples, then a few modified to animate many-at-a-time and it all looks super smooth :)

PR #758, opened 2015-04-28
Cleanup for Board.Pins

scottgonzalez:


lib/board.pins.js
```diff
@@ -192,40 +190,6 @@ Pins.translations = {
       D4: 4,
       D2: 2
     }
-  },
-  LEONARDO: {
```
scottgonzalez: This was overwritten immediately after to just be a pointer to `Pins.translations.UNO`.

PR #763, opened 2015-05-01
Update I2C API

scottgonzalez:
The only component I was able to test with a real device was `wii`, but the API change is very straight-forward.

### DIFF
--- a/lib/ir.js
+++ b/lib/ir.js
@@ -146,20 +146,20 @@ function IR(opts) {
   });
 
   // Set up I2C data connection
-  this.io.sendI2CConfig();
+  this.io.i2cConfig();
 
   // Enumerate and write each set of setup instructions
   setup.forEach(function(byteArray) {
-    this.io.sendI2CWriteRequest(address, byteArray);
+    this.io.i2cWrite(address, byteArray);
   }, this);
 
   // Read Request Loop
   setInterval(function() {
     // Set pointer to X most signficant byte/register
-    this.io.sendI2CWriteRequest(address, preread);
+    this.io.i2cWrite(address, preread);
 
     // Read from register
-    this.io.sendI2CReadRequest(address, bytes, data.bind(this));
+    this.io.i2cReadOnce(address, bytes, data.bind(this));
 
   }.bind(this), delay);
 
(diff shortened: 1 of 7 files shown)


Resseguie:
From @frxnz: "Just tried to test IR with a GP2Y0D805Z0F. Technically it behaves the same, but it looks like there's a bug in master" (See #766)

rwaldron:
I'll try a few devices, but I'm pretty sure this is totally fine.

lib/ir.js
```diff
@@ -146,20 +146,20 @@ function IR(opts) {
...
   setInterval(function() {
     // Set pointer to X most signficant byte/register
-    this.io.sendI2CWriteRequest(address, preread);
+    this.io.i2cWrite(address, preread);
 
     // Read from register
-    this.io.sendI2CReadRequest(address, bytes, data.bind(this));
+    this.io.i2cReadOnce(address, bytes, data.bind(this));
 
   }.bind(this), delay);
```
rwaldron: Confirmed, this is ok. 

Also, this file has reached end of life.

lib/led/ledcontrol.js
```diff
@@ -540,7 +540,7 @@ Controllers = {
       if (arguments.length !== 3) {
         throw new Error("`send` expects three arguments: device, opcode, data");
       }
-      this.io.sendI2CWriteRequest(this.addresses[addr], [opcode | data]);
+      this.io.i2cWrite(this.addresses[addr], [opcode | data]);
       return this;
     }
   },
```
rwaldron: Confirmed, this is ok.

lib/sonar.js
```diff
@@ -129,7 +129,7 @@ Devices = {
 
       function prime() {
         // 0x52 result in us (microseconds)
-        this.io.sendI2CWriteRequest(address, [0x00, 0x52]);
+        this.io.i2cWrite(address, [0x00, 0x52]);
 
         setTimeout(read.bind(this), delay);
       }
```
rwaldron: Confirmed, this is ok.

Also, this file has reached end of life.

scottgonzalez:
Sounds like this is good to land then?

Resseguie:
LGTM

rwaldron:
I haven't reality checked the Wii stuff, but you have, so that covers everything!

PR #765, opened 2015-05-01
Remove Gripper until device/controller model ready

### LINKED ISSUE #698: Refactor gripper component using controller/device model
Deemed too specific.

Resseguie:
Fixes #698 

To prepare for 0.9.0 I recommend just removing the incomplete Gripper component. We can add it back later when the device/controller model is ready for it. Until then, grippers can still be controlled just using Servo.


rwaldron:
The eg files don't need to be deleted, just remove the entry from programs.json

Resseguie:
I  was afraid someone might see them and wonder why they didn't work... No strong opinion either way though, so I can just leave them in. I'll amend later. My hope is to get a new version of Gripper back in with device/controller support in the not too distant future.

rwaldron:
Why doesn't navigator.js work?

Resseguie:
It uses Gripper, which this PR removes. ;)

It could be updated to just use Servo instead.

rwaldron:
I don't want to remove those. They were used in the program I wrote to control the robot that used to propose to my wife. There's no reason to remove gripper, just to bring it back later.

Resseguie:
:heart_eyes: Then it stays for now!

PR #775, opened 2015-05-05
stepper tests getCall/args consistency

Resseguie:


Resseguie:
Note, it's somewhat arbitrary to use either `spy.args[x][y]` vs `spy.getCall(x).args[y]`

But this PR makes this set of tests consistent. Might still want to do a look across all tests...

PR #776, opened 2015-05-05
fix RGB-PCA9685 lead mismatch in fzz

Resseguie:
Visually it looked right, but upon inspection, the fzz leads for the RGB were mixed up.


rwaldron:
Once travis is green, go ahead and land. Thanks!

Resseguie:
I had forgotten to update the example doc. I just did an amend commit for that...

rwaldron:
Seems good. When it's green, it's good to go.


Also, I like the change to the ground line.

PR #777, opened 2015-05-05
Led: Test cleanup

scottgonzalez:
Here's the first phase of cleanup. Just the Led component so far. I'll get to the other components (Array, RGB, etc.) after.


rwaldron:
Nice!

PR #781, opened 2015-05-07
Examples: Add common anode RGB with PCA9685 controller

scottgonzalez:


frxnz:
The common pin should be connected to V+ instead of GND. 

I'm not in a place I can test this right now, but I believe the correct wiring is to connect the PWM pins directly (no resistor), and then put a 330 ohm resistor between V+ and the common pin.

scottgonzalez:
> The common pin should be connected to V+ instead of GND.

Oops. I meant to connect it to V+, that's why I used the same color wire as the lead going into V+. I've updated the diagram.

> I believe the correct wiring is to connect the PWM pins directly (no resistor), and then put a 330 ohm resistor between V+ and the common pin.

I assumed the resistor usage was the same as a direct connection to the Arduino. @rwaldron confirmed this in gitter when he said "Be sure to put resistors between the r, g, b lines and the PWM outs."

rwaldron:
@frxnz that will work too, but will give you the opposite color :D

Check it out:

PR #785, opened 2015-05-08
Joystick: Controller refactor, Esplora support.

rwaldron:


scottgonzalez:
I'm just starting to dig into this. As a first step, I ran `eg/joystick.js` with a [Parallax joystick](https://www.parallax.com/product/27800) and I ran `eg/joystick-esplora.js` with an Esplora. I get values that I would expect from the Parallax joystick (-1 is left and down), but I get inverted values with the Esplora (-1 is right and up).

lib/joystick.js
```diff
@@ -1,6 +1,116 @@
...
+
+          // Ensure that both pins have been read before
+          // reporting a set of values.
+          if (count === 2) {
+            count = 0;
+            dataHandler(dataPoints);
+          }
+
+          // Remove this handler to all the multiplexer
+          // to setup the next pin for the next read.
```
scottgonzalez: It looks like you might have started writing one sentence than changed what you were writing.

lib/joystick.js
```diff
@@ -1,6 +1,116 @@
...
+          this.io.removeListener("analog-read-4", handler);
```
scottgonzalez: I haven't looked through the code to see how common this is. Would it make sense to have an `analogReadOnce()` similar to `i2cReadOnce()`?

lib/joystick.js
```diff
@@ -11,149 +121,109 @@ var Board = require("../lib/board.js"),
...
-    y: 0
-  };
+  if (typeof this.initialize === "function") {
+    this.initialize(opts, function(data) {
+      var isChange = false;
 
+      Object.keys(data).forEach(function(axis) {
+        var value = data[axis];
+        var sensor = state[axis];
+        var zeroV = sensor.zeroV < 0 ? sensor.zeroV * -1 : sensor.zeroV;
```
scottgonzalez: Why not just `Math.abs(sensor.zeroV)`?

lib/joystick.js
```diff
@@ -11,149 +121,109 @@ var Board = require("../lib/board.js"),
...
 
-      // register X axis, L/R (A0)
-      if (pin === 1) {
-        this.axis.x = val;
-        this.raw.x = data;
-        // register Y axis, U/D (A1)
-      } else {
-        this.axis.y = val;
-        this.raw.y = data;
+      if (isChange) {
```
scottgonzalez: Should joystick support a threshold similar to the sensor component?

test/joystick.js
```diff
@@ -0,0 +1,301 @@
...
+  repl: false
+});
+
+
+exports["Joystick -- Analog"] = {
+
+  setUp: function(done) {
+
+    this.clock = sinon.useFakeTimers();
+    this.analogRead = sinon.spy(MockFirmata.prototype, "analogRead");
```
scottgonzalez: Out of curiosity, why do you create a spy on the prototype instead of the instance?

lib/joystick.js
```diff
@@ -1,6 +1,116 @@
...
+        var dataPoints = {
+          x: 0,
+          y: 0
+        };
+
+        opts.pins.forEach(function(pin, index) {
+          this.io.pinMode(pin, this.io.MODES.ANALOG);
+          this.io.analogRead(pin, function(value) {
+            dataPoints[axes[index]] = value;
+            count++;
```
scottgonzalez: Is this really a safe method for tracking when both pins have been read?

test/joystick.js
```diff
@@ -0,0 +1,301 @@
...
+    test.ok(spy.calledOnce);
+    test.deepEqual(spy.args[0], [{
+      x: 512,
+      y: 560
+    }]);
+
+    test.done();
+  },
+
+  change: function(test) {
```
scottgonzalez: There should be an additional check to ensure the event is not emitted when the value has not changed.

test/joystick.js
```diff
@@ -0,0 +1,301 @@
...
+
+    this.clock = sinon.useFakeTimers();
+    this.analogRead = sinon.spy(MockFirmata.prototype, "analogRead");
+    this.stick = new Joystick({
+      controller: "ESPLORA",
+      freq: 100,
+      board: board
+    });
+
+    this.instance = [{
```
scottgonzalez: Since the shape is consistent across controllers, you might consider moving this up to keep the code DRY. This was recently done in the Led tests: https://github.com/rwaldron/johnny-five/blob/fd73b85b96f54bcddfe14f61313ddb2c82ab1fd9/test/led.js#L31-L39

scottgonzalez:
On a slightly related note, the Joystick wiki page says "This list will continue to be updated as more Joystick devices are confirmed." but there is no list.

scottgonzalez:
If you'd like, I can add a Fritzing diagram for the Parallax joystick. It requires two 5v leads, so I'm not sure if there's a good way to show that without using a breadboard. Here's what I've got right now:

Parallax joystick wiring

Resseguie:
@scottgonzalez Personal preference: I'd just line that joystick up on the breadboard and wire it there instead of directly to the spots on the joystick itself. That's more likely how it would physically be done anyway I would imagine.

scottgonzalez:
> I'd just line that joystick up on the breadboard and wire it there instead of directly to the spots on the joystick itself.

I can do that for most of the wires, but the joystick is too wide to fit on the breadboard and still have a row on each side for the wires. This is only true in the diagram; in reality, I have the joystick centered on a breadboard and have just enough room to connect all the wires on either side. The joystick actually comes with the headers soldered on, so the diagram I created can't really be done anyway. I'll spend some time figuring out how to use the Fritzing parts editor so I can make it the right size (and maybe get some more legible text).

lib/joystick.js
rwaldron: Haha. Looks like it

rwaldron:
> I get values that I would expect from the Parallax joystick (-1 is left and down), 

Unfortunately, this has uncovered a bigger issue.

- You're right
- Some Joysticks, whether they are DIY thumb controls, or console game controllers, report these values differently.

I think the solution is to default to: 


```
    -1
-1   *   1
     1      
```

And allow the user to specify inversion.

```js
var joy = new Joystick({ pins: ["A0", "A1"], invertY: true });

// Produces a joystick that will report like this: 
     1
-1   *   1
     -1      
```

```js
var joy = new Joystick({ pins: ["A0", "A1"], invertX: true });

// Produces a joystick that will report like this: 
     -1
1   *   -1
     1      
```

```js
var joy = new Joystick({ pins: ["A0", "A1"], invert: true });

// Produces a joystick that will report like this: 
     1
1   *   -1
     -1      
```

lib/joystick.js
rwaldron: Artifact from porting C code.

lib/joystick.js
rwaldron: Will do

test/joystick.js
rwaldron: When using the instance, I've been experiencing strange behavior that I don't have the time resources to deeply investigate. While trying to figure that out, I discovered that using the prototype eliminated the issue and I've been converting them as I go along since. There is plenty that I dislike about solving it this way and even the explanation, but I don't have the time in the day to spend figuring it out.

lib/joystick.js
rwaldron: I'll find another approach

lib/joystick.js
rwaldron: This is so uncommon that this is the only case, and it's only necessary because the Esplora forces all its built in analog components to report their values through 1 pin—which is insane. `i2cReadOnce` is fairly common.

test/joystick.js
rwaldron: Actually, I just encountered the reason why: 

```js
exports["Joystick -- Analog"] = {
  setUp: function(done) {

    this.stick = new Joystick({
      pins: ["A0", "A1"],
      freq: 100,
      board: board
    });
    this.analogRead = sinon.spy(this.stick, "analogRead");
    done();
  },
  tearDown: function(done) {
    this.analogRead.restore();
    done();
  },
  data: function(test) {
    var x = this.analogRead.args[0][1];
    var y = this.analogRead.args[1][1];
    var spy = sinon.spy();

    test.expect(2);

    this.stick.on("data", spy);

    x(512);
    y(560);

    this.clock.tick(200);

    test.ok(spy.calledOnce);
    test.deepEqual(spy.args[0], [{
      x: 512,
      y: 560
    }]);

    test.done();
  },
  invertX: function(test) {
    test.expect(2);

    // Need to construct a Joystick in here
    // to test the invertY option...
    var = new Joystick({
      pins: ["A0", "A1"],
      freq: 100,
      invertY: true,
      board: board
    });

    // Now the this.analogRead spy is useless.
    
    test.done();
  }
};
```

rwaldron:
w/r to the diagram: I would prefer to just create a new component/part that works correctly and looks like the actual parallax joystick.

test/joystick.js
scottgonzalez: Ah, yeah, that makes sense.

scottgonzalez:
> I think the solution is to default to:
> ```
>     -1
> -1   *   1
>      1
> ```

I would think pushing up would result in higher values, not lower. But really as long as it's consistently normalized, any order should be fine.

scottgonzalez:
I'll do another review of the code when I get home.

I'll (eventually) work on a diagram for the Parallax joystick. I had just done a search for "parallax joystick fritzing" and found that part, but it's pretty shoddy.

rwaldron:
> I would think pushing up would result in higher values, not lower. But really as long as it's consistently normalized, any order should be fine.

Exactly. I was researching other joystick configs and it's all over the map, so it occurred to me that Johnny-Five should just let the author get it however they want it.

lib/joystick.js
```diff
@@ -11,149 +125,124 @@ var Board = require("../lib/board.js"),
...
-     * @name magnitude
-     * @property
-     * @type Number
-     */
-    magnitude: {
-      get: function() {
-        return Math.hypot(this.axis.x, this.axis.y);
+    hasAxis: {
+      value: function(axis) {
+        return state[axis] ? typeof state[axis].value !== "undefined" : false;
```
scottgonzalez: Won't this always return `true` for `"x"` and `"y"` and `false` for any other value?

It looks like this was designed to handle a three axis joystick as well, but there's no way to actually read from the third axis. Similarly, I can power my Parallax joystick such that only one axis will work, but I wouldn't be able to use this component without providing two pins to read from since the data events are blocked on reading values from both axes.

test/joystick.js
```diff
@@ -0,0 +1,576 @@
...
+
+    test.done();
+  },
+
+  change: function(test) {
+    test.expect(11);
+
+    var spy = sinon.spy();
+    var x = this.stick.io.analogRead.args[0][1];
+    var y = this.stick.io.analogRead.args[1][1];
```
scottgonzalez: This should probably be consistent with the previous test (`this.analogRead` vs `this.stick.io.analogRead`).

test/joystick.js
```diff
@@ -0,0 +1,576 @@
...
+
+    test.equal(spy.callCount, 3);
+
+    test.deepEqual(spy.args[0], [{
+      x: 1,
+      y: 1
+    }]);
+
+    test.equal(Math.round(spy.args[1][0].x), 0);
+    test.equal(Math.round(spy.args[1][0].y), 0);
```
scottgonzalez: Seems strange to have two `deepEqual()` assertions surrounding these two when you're testing the same thing.

lib/joystick.js
rwaldron: Interesting use case, I think I'll refactor the controllers to only enforce axes collection for those that the user defines.

test/joystick.js
rwaldron: Yep, an oversight—I thought I had updated them all. Fix in next push

test/joystick.js
rwaldron: I think I need to just revisit the data handling, because it seems wrong to require Math.round

rwaldron:
@scottgonzalez updates pushed. 

- Punting on "optional" axes for this patch
- Fixed data handling to get a proper `0` at center

scottgonzalez:
Looks good. I ran the examples against my Parallax joystick and my Esplora and they both worked great.

rwaldron:
Awesome. I really like that new inversion feature.

PR #789, opened 2015-05-10
Consolidate motion components

frxnz:
Consolidates `PIR` and `GP2Y0D805Z0F`

@rwaldron I can't get my GP2Y0D805Z0F sensor to work, so if you could reality-check that would be awesome. I did successfully test with a PIR sensor.


frxnz:
A couple things:

* I added breadboards for GP2Y0D805Z0F, but didn't have a sketch for the device, so it's just wires.
* Ugh, those merge commits. Any tips on how to get rid of those? Do I just need to squash? My git-fu is weak.

rwaldron:
The trick is not to git merge when you're on a feature branch, instead do `git pull --rebase upstream master`. I can probably cherry pick your commits

rwaldron:
Great work, I'll test with that component as soon as I have a chance :)

PR #790, opened 2015-05-14
Accelerometer: Add support for Esplora

scottgonzalez:
Includes some other minor cleanup.

### DIFF
--- a/lib/accelerometer.js
+++ b/lib/accelerometer.js
@@ -196,6 +196,23 @@ var Controllers = {
         }
       }
     }
+  },
+  ESPLORA: {
+    DEFAULTS: {
+      value: {
+        zeroV: [320, 330, 310],
+        sensitivity: 170
+      }
+    },
+    initialize: {
+      value: function(opts, dataHandler) {
+        opts.pins = [5, 11, 6];
+        analogInitialize.call(this, opts, dataHandler);
+      }
+    },
+    toGravity: {
+      value: analogToGravity
+    }
   }
 };
 
--- a/lib/accelerometer.js
+++ b/lib/accelerometer.js
@@ -86,11 +86,11 @@ var Controllers = {
         sensitivity: 66.5
       }
     },
-    initialize: { 
+    initialize: {
       value: analogInitialize
     },
-    toGravity: { 
-      value: analogToGravity 
+    toGravity: {
+      value: analogToGravity
     }
   },
   ADXL345: {
@@ -180,7 +180,7 @@ var Controllers = {
           this.board.pinMode(state.sleepPin, 1);
           this.board.digitalWrite(state.sleepPin, 1);
         }
-        
+
         analogInitialize.call(this, opts, dataHandler);
       }
     },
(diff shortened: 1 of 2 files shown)


rwaldron:
@BrianGenisio can you review this?

lib/accelerometer.js
```diff
@@ -167,7 +167,7 @@ var Controllers = {
   MMA7361: {
     DEFAULTS: {
       value: {
-        zeroV: [336, 372, 287],
+        zeroV: [330, 350, 290],
```
rwaldron: This could be avoided by defining the ESPLORA controller like this: 

```js
Controllers["ESPLORA"] = {
  DEFAULTS: {
    value: {
      zeroV: [x, y, z], 
      sensitivity: 170
    }
  }, 
  initialize: {
    value: function(opts, dataHandler) {
      opts.pins = [5, 11, 6];
      analogInitialize.call(this, opts, dataHandler);
    }
  },
  toGravity: Controllers.MMA7361.toGravity
};
```

scottgonzalez: Sure. I did this because the calibration is never going to be perfect for all users. If we're going to pick values based on the first person's calibration (I'm assuming that's where these numbers came from), we might as well just round the values to nice looking numbers that represent an average across tested devices. An MMA7361 in an Esplora shouldn't really have different values than an MMA7361 in other settings. FWIW, the values that ship in `Esplora.cpp` are `[320,330,310]`.

rwaldron: > Sure. I did this because the calibration is never going to be perfect for all users.

True, which is why Accelerometer allows for user specified zeroV:

```js
var accelerometer = new five.Accelerometer({
  controller: "MMA7361",
  pins: ["A0", "A1", "A2"],
  zeroV: [zeroVx, zeroVy, zeroVz],
});
```


> If we're going to pick values based on the first person's calibration (I'm assuming that's where these numbers came from), 

TBH, I don't know how the default zeroV numbers were derived, but they were correct for both @BrianGenisio (who authored the controller) and myself. 


> An MMA7361 in an Esplora shouldn't really have different values than an MMA7361 in other settings. 

That's not strictly true. Any number of differences in the circuit that the chip is attached to could cause slight variations.

> FWIW, the values that ship in Esplora.cpp are [320,330,310].


Ah, in that case: 

```js
Controllers["ESPLORA"] = {
  DEFAULTS: {
    value: {
      zeroV: [320, 330, 310], 
      sensitivity: 170
    }
  }, 
  initialize: {
    value: function(opts, dataHandler) {
      opts.pins = [5, 11, 6];
      analogInitialize.call(this, opts, dataHandler);
    }
  },
  toGravity: Controllers.MMA7361.toGravity
};
```

scottgonzalez: Ok, I'll update to use the values from `Esplora.cpp`.

rwaldron: Perfect! 

Once the update is green, this is good to go :D

scottgonzalez:
updated

PR #799, opened 2015-05-21
Expander: MCP23017, MCP23008, PCF8574, PCF8574A, PCF8575, PCA9685

rwaldron:
- Can create standalone Expander instances
- Use as virtual boards


scottgonzalez:
The MCP23017 diagram has the rightmost blue LED connected to an empty row on the breadboard.

rwaldron:
Nice catch!

rwaldron:
Fixed!

Issue #742, opened 2015-04-20
Refactor Led.RGB to device/controller model. 


rwaldron:
- Add support for BlinkM controller.

scottgonzalez:
If you have some explanations or examples of this model, I could try to tackle this. Would probably give me some better understanding of how everything works.

Resseguie:
I need to learn more about the device/controller model too, so I'm happy to think it through with you @scottgonzalez 

Servo and LCD are two examples I was going to start with.

@dtex is a good resource for understanding the organization.

dtex:
Motor is also a good example. Mentioning @BrianGenisio because he's been in the position of having to make sense of it all.

dtex:
This helps: https://github.com/rwaldron/johnny-five/issues/665#issuecomment-85332468

And there was another, more detailed explanation but I'm having trouble finding it right now.

dtex:
Here's the other: https://github.com/rwaldron/johnny-five/issues/486#issuecomment-63515604

Resseguie:
Seems like we always have to hunt that stuff down. We need to put an "understanding devices/controllers" section on the wiki somewhere and make it easy to find.

scottgonzalez:
I'll take notes as I work on this tonight, then we can flesh it out into full API docs or a tutorial or something.

frxnz:
> We need to put an "understanding devices/controllers" section on the wiki somewhere

[We have this](https://github.com/rwaldron/johnny-five/wiki/Control-System-Hierarchy-Overview), but looking at the comments @dtex linked we could maybe expand on it a little.

Resseguie:
Thanks @frxnz. An example or two there to put them in context would probably also be helpful.
See #746

scottgonzalez:
Since `Led.RGB` is documented as having `red`, `green`, and `blue` properties which contain `Led` instances, how do you want to handle devices like BlinkM where you can't use a composite of multiple `Led` instances to construct the `Led.RGB` instance? Should we just drop those properties from the public API and update the documentation for the shape of the component?

rwaldron:
TBH...

> Should we just drop those properties from the public API and update the documentation for the shape of the component?

Sounds like what should've been done the first time. So yeah, do this.

PR #803, opened 2015-05-24
Led.RGB: Refactor to device/controller model

scottgonzalez:
Ref #742 

The initial work is done. `Led.RGB` now uses the device/controller model and has full support for direct use and PCA9685. I'll be adding support for BlinkM and Esplora controllers as well. Since the majority of the work is already done, I figured I'd file the PR now to start getting feedback.

### DIFF
--- a/lib/led/rgb.js
+++ b/lib/led/rgb.js
@@ -1,12 +1,117 @@
-var Led = require("./led"),
-  priv = new Map();
+var Board = require("../board.js");
+var nanosleep = require("../sleep.js").nano;
+
+var priv = new Map();
+
+var Controllers = {
+  DEFAULT: {
+    initialize: {
+      value: function(opts) {
+        RGB.colors.forEach(function(color, index) {
+          var pin = opts.pins[index];
+
+          if (!this.board.pins.isPwm(pin)) {
+            Board.Pins.Error({
+              pin: pin,
+              type: "PWM",
+              via: "Led.RGB"
+            });
+          }
+
+          this.io.pinMode(pin, this.io.MODES.PWM);
+          this.pins[index] = pin;
+        }, this);
+      }
+    },
+    write: {
+      value: function(colors) {
+        var state = priv.get(this);
+
+        RGB.colors.forEach(function(color, index) {
+          var pin = this.pins[index];
+          var value = colors[color];
+
+          if (state.isAnode) {
+            value = 255 - Board.constrain(value, 0, 255);
+          }
+
+          this.io.analogWrite(pin, value);
+        }, this);
+
+        Object.assign(state, colors);
+      }
+    }
+  },
+  PCA9685: {
+    COMMANDS: {
+      value: {
+        PCA9685_MODE1: 0x0,
+        PCA9685_PRESCALE: 0xFE,
+        LED0_ON_L: 0x6
+      }
+    },
+    initialize: {
+      value: function(opts) {
+        this.address = opts.address || 0x40;
+
+        if (!this.board.Drivers[this.address]) {
+          this.io.i2cConfig();
+          this.board.Drivers[this.address] = {
+            initialized: false
+          };
+
+          // Reset
+          this.io.i2cWrite(this.address, [this.COMMANDS.PCA9685_MODE1, 0x0]);
+          // Sleep
+          this.io.i2cWrite(this.address, [this.COMMANDS.PCA9685_MODE1, 0x10]);
+          // Set prescalar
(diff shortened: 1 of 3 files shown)


test/led.js
```diff
@@ -38,6 +38,26 @@ var instanceProperties = [{
...
+  name: "strobe"
+}, {
+  name: "blink"
+}, {
+  name: "stop"
+}];
+
+var rgbInstanceProperties = [{
+  name: "interval"
+}];
```
scottgonzalez: This property is tested on `Led`, so I tested it here as well, but I would expect this to be hidden inside the private state in both cases.

lib/led/rgb.js
```diff
@@ -1,12 +1,117 @@
...
+          this.io.i2cWrite(this.address, [this.COMMANDS.PCA9685_MODE1, 0xa1]);
+
+          this.board.Drivers[this.address].initialized = true;
+        }
+
+        RGB.colors.forEach(function(color, index) {
+          var pin = opts.pins[index];
+          this.pins[index] = pin;
+        }, this);
+      }
```
rwaldron: Shouldn't this just defer to `Led` with the same controller?

test/led.js
rwaldron: > This property is tested on Led, so I tested it here as well, but I would expect this to be hidden inside the private state in both cases.

Yeah, that's a legacy bug and _yes_ it should be in private state.

lib/led/rgb.js
scottgonzalez: It could. I don't think it buys us much, especially if controller initialization gets abstracted away in the future.

rwaldron: Strike that, I see now what you've done and this is fine.

lib/led/rgb.js
```diff
@@ -15,47 +120,43 @@ var RGB = function(opts) {
...
-  }.bind(this));
+  if (controller == null) {
+    controller = Controllers["DEFAULT"];
+  }
+
+  Object.defineProperties(this, controller);
 
   this.interval = null;
 
+  // The default color is #ffffff, but the light will be off
```
rwaldron: √

scottgonzalez:
I just pushed a bunch of updates, including BlinkM support. I'll add Esplora support when I get home. Since Esplora support is really just delegating out to the default controller, this is ready for a full review now.

scottgonzalez:
Esplora support has been added.

rwaldron:
I'm at JSConf and don't have an Esplora to test on. Either we can wait until Sunday, or you can whip up a quick video showing some basic color cycling?

lib/led/rgb.js
```diff
@@ -1,12 +1,158 @@
...
+          this.board.io.i2cWrite(this.address, [this.COMMANDS.STOP_SCRIPT]);
+
+          this.board.Drivers[this.address].initialized = true;
+        }
+      }
+    },
+    write: {
+      writable: true,
+      value: function(colors) {
+        this.board.io.i2cWrite(this.address,
```
rwaldron: One thing I want to clean up here (and eventually everywhere else, as we go along) is to migrate all occurrences of `this.board.io` to `this.io`

lib/led/rgb.js
```diff
@@ -81,11 +222,23 @@ var RGB = function(opts) {
     values: {
       get: function() {
         return RGB.colors.reduce(function(current, color) {
-          return (current[color] = this[color].value, current);
-        }.bind(this), {});
+          return (current[color] = state[color], current);
+        }, {});
+      }
+    },
+    writeColor: {
```
rwaldron: I grok the motivation for splitting out this method's functionality from `write`, but I'd like to suggest renaming `writeColor` to `update` because I think it's nicer to explain as:


- `update`: updates the internal state and then calls `write`
- `write`: writes the raw value to the component

This is my last bit of change-centric feedback, after that it's just reality checks with hardware.

scottgonzalez:
> I'm at JSConf and don't have an Esplora to test on. Either we can wait until Sunday, or you can whip up a quick video showing some basic color cycling?

I can take a video tonight.

scottgonzalez:
I made the `this.board.io` -> `this.io` and `.writeColor()` -> `.update()` changes in a1f3150849fc0f56067fd434b48ac7b907c27718.

scottgonzalez:
Here's the video of the Esplora: http://f.cl.ly/items/3O3h3d2M0P3c3t0q3I1c/IMG_0952.MOV.m4v

PR #807, opened 2015-05-27
Led: Remove type option

scottgonzalez:
I asked @rwaldron about this the other day and he said there's no reason for this to stay around.


scottgonzalez:
I'll update the wiki once this is merged.

rwaldron:
Awesome, thanks for cleaning that up!

scottgonzalez:
The wiki is updated.

Issue #810, opened 2015-05-30
Board ready event does not fire in multi-board configuration


dtex:
Reported by @sharp6
> I'm trying to work with two different boards (Arduino Uno and Pro Mini). I followed the example board-multi (https://github.com/rwaldron/johnny-five/blob/master/eg/board-multi.js), and J5 tells me that both boards are connected, but the "ready" event never gets fired, and as such, I can't get my board objects to work with.
> I tried manually assigning the ports, but got the same results. 

Confirmed by @dtex using two Unos. Neither board's ready even is emitted... Actually, I don't know that. I do know that the listener added on https://github.com/rwaldron/johnny-five/blob/master/lib/board.js#L997 is added but never triggered.

PR #812, opened 2015-05-31
Boards: restore "lost" ready event for multi-board cases. Fixed gh-810

rwaldron:
This may also correct an issue that I think @brianshaler was trying to explain to me at JSConf, but that needs to be confirmed.

cc @dtex @Resseguie @Sharp6

### DIFF
--- a/lib/board.js
+++ b/lib/board.js
@@ -424,7 +424,7 @@ function broadcast(err, type, io) {
     //
     // If the user program has not disabled the
     // REPL, initialize it.
-    if (!Repl.isBlocked && this.repl) {
+    if (this.repl) {
       this.repl.initialize(this.emit.bind(this, "ready"));
     }
 
@@ -450,7 +450,7 @@ function broadcast(err, type, io) {
       this.emit(type, err);
     }
   } else {
-  // The REPL is disabled, propagate all events
+    // The REPL is disabled, propagate all events
     this.emit(type, err);
   }
 }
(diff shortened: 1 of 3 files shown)


brianshaler:
I'm not in a position to test today, but scanning the commit, I have a hunch the crash I was experiencing will still occur. The failure path is calling `new Board()` with no options and getting to [Serial.detect.call(this, function(path) {](https://github.com/rwaldron/johnny-five/blob/810/lib/board.js#L316) where [serialport.list(function(err, result) {](https://github.com/rwaldron/johnny-five/blob/810/lib/board.js#L75) blows up if it doesn't receive an array as the second parameter, which [node-serialport will do _sometimes_](https://github.com/voodootikigod/node-serialport/blob/master/serialport.js#L565) (e.g. when err exists but err.errno != 34)

rwaldron:
> 'm not in a position to test today, but scanning the commit, I have a hunch the crash I was experiencing will still occur. The failure path is calling new Board() with no options and getting to Serial.detect.call(this, function(path) { where serialport.list(function(err, result) { blows up if it doesn't receive an array as the second parameter, which node-serialport will do sometimes (e.g. when err exists but err.errno != 34)

I didn't mean this error—maybe it wasn't you that was telling me about some no-repl-in-browser issue?

rwaldron:
https://github.com/rwaldron/johnny-five/commit/17315f1a61cb233fcbb5e9a257a6ddcae6964d8e

PR #814, opened 2015-06-01
Match pins in LCD example to comments

stevekinney:
We played around with this for a while and couldn't get it working with the original version. After some research, this proposed change works. Coincidentally, it matches up with the commented documentation directly above it.

### DIFF
--- a/docs/lcd.md
+++ b/docs/lcd.md
@@ -22,8 +22,8 @@ board.on("ready", function() {
   lcd = new five.LCD({
     // LCD pin name  RS  EN  DB4 DB5 DB6 DB7
     // Arduino pin # 7    8   9   10  11  12
-    pins: [7, 8, 9, 10, 11, 12],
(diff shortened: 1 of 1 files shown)


rwaldron:
I'm mobile at the moment, so this isn't just me being lazy: does the fritzing diagram match the changes?

rwaldron:
I just realized something else: the change needs to be made in the `eg/lcd.js` file, then run `grunt examples`, otherwise your change gets paved over.

stevekinney:
Ah. I will check everything and let you know.

rwaldron:
@stevekinney cool, and be sure to @ ping me when this is ready for a review :)

rwaldron:
I updated the patch and set you as author. Landed: https://github.com/rwaldron/johnny-five/commit/9ad3e0fe7469a3322cbcd6b972317d20fffaec19

Issue #811, opened 2015-05-30
Feature idea: RGB LED color correction


mize:
Caveat: I'm extremely new to hardware.

Tinkering with an RGB LED, I noticed I got mixed colors that were different from what I expected based on hex values passed via the color method. So I read some background info on color variations:
- [Understand RGB LED mixing ratios to realize optimal color in signs and displays](http://www.ledsmagazine.com/articles/print/volume-10/issue-6/features/understand-rgb-led-mixing-ratios-to-realize-optimal-color-in-signs-and-displays-magazine.html)
- [FastLED Color Correction](https://github.com/FastLED/FastLED/wiki/FastLED-Color-Correction)

Looks like the FastLED lib has some constant [white point values](https://github.com/FastLED/FastLED/blob/master/color.h) that they allow the user to designate via setCorrection or setTemperature methods, and the resulting value [transforms any further colors passed in](https://github.com/FastLED/FastLED/blob/master/controller.h#L123-L151).

Could be cool to add this feature to J5, in Led.RGB constructor options and as an instance method for flexibility during runtime. I think it would be handy to use a standard set of values, as well as letting J5 user designate a custom white point as a hex string.

I'm willing to give it a go. Would other folks find this useful? I'll also look around to see whether this is a solved problem via some other JS lib.

rwaldron:
Any updates?

dtex:
Hi @mize 

I'm going to close this since we haven't heard back, but it still sounds like a good addition. To make sure it doesn't get forgotten I've added it to the Johnny-Five [Requested Features](../wiki/Requested-Features) page.

PR #818, opened 2015-06-03
Led.RGB: Add intensity method

### LINKED ISSUE #737: Add Led.RGB.intensity()
Similar to current `Led.brightness()` but has range 0-100 (percent).

Should consider deprecating `.brightness()` completely and add `.intensity(0-100)` to Led as well.

Reasoning: "brightness" can be misinterpreted. "intensity" is standard term in lighting industry.

scottgonzalez:
Fixes #737

I want to make sure this is the expected functionality before I write the tests for this.

The idea is that we find the color with the highest value and scale based on that. The internal values may contain fractional values as a result, but this ensures that we're not losing information as we scale to very low intensities. As a result, `.color()` when used as a getter has been modified to round the values. This currently works fine, but I'm not sure if we should be rounding the values before sending them to `io` just to be safe.

### DIFF
--- a/lib/led/rgb.js
+++ b/lib/led/rgb.js
@@ -269,7 +269,7 @@ RGB.prototype.color = function(red, green, blue) {
     // not a reference to the state object itself.
     colors = this.isOn ? state : state.prev;
     return RGB.colors.reduce(function(current, color) {
-      return (current[color] = colors[color], current);
+      return (current[color] = Math.round(colors[color]), current);
     }, {});
   }
 
@@ -401,6 +401,39 @@ RGB.prototype.stop = function() {
 
   clearInterval(state.interval);
   delete state.interval;
+
+  return this;
+};
+
+RGB.prototype.intensity = function(intensity) {
+  var state = priv.get(this);
+  var colors = this.isOn ? state : state.prev;
+
+  // Calculate current intensity on a scale of 0-1
+  var currentIntensity = RGB.colors.reduce(function(current, color) {
+    return Math.max(colors[color], current);
+  }, 0) / 255;
+
+  if (arguments.length === 0) {
+    return Math.round(currentIntensity * 100);
+  }
+
+  var scale = 1 / currentIntensity * Math.min(100, Math.max(0, intensity)) / 100;
+  var scaledColors = RGB.colors.reduce(function(current, color) {
+    return (current[color] = colors[color] * scale, current);
+  }, {});
+
+  // If the intensity is low enough to turn off all three colors,
+  // go through .off() to store the current color
+  var turningOff = RGB.colors.reduce(function(current, color) {
+    return current && (Math.round(scaledColors[color]) === 0);
+  }, true);
+
+  if (turningOff) {
+    this.off();
+  } else {
+    this.update(scaledColors);
+  }
 
   return this;
 };
--- a/test/led.js
+++ b/test/led.js
@@ -842,6 +842,37 @@ exports["Led.RGB"] = {
     test.expect(1);
     test.equal(this.ledRgb.blink, this.ledRgb.strobe);
     test.done();
+  },
+
+  intensity: function(test) {
+    test.expect(13);
+
+    this.ledRgb.color("#33aa00");
+    test.equal(this.ledRgb.intensity(), 67);
+    this.write.reset();
+
+    // full intensity
+    test.equal(this.ledRgb.intensity(100), this.ledRgb);
+    test.ok(this.write.calledOnce);
+    test.ok(this.write.calledWith({ red: 76.5, green: 255, blue: 0 }));
+    test.ok(this.ledRgb.intensity(), 100);
+    this.write.reset();
+
+    // fully off
+    test.equal(this.ledRgb.intensity(0), this.ledRgb);
+    test.ok(this.write.calledOnce);
+    test.ok(this.write.calledWith({ red: 0, green: 0, blue: 0 }));
+    test.ok(this.ledRgb.intensity(), 0);
+    this.write.reset();
+
+    // restore from off
+    test.equal(this.ledRgb.intensity(50), this.ledRgb);
+    test.ok(this.write.calledOnce);
+    test.ok(this.write.calledWith({ red: 38.25, green: 127.5, blue: 0 }));
+    test.ok(this.ledRgb.intensity(), 50);
+    this.write.reset();
+
+    test.done();
   }
 };
 


Resseguie:
Should `.intensity()` set state that is then used to translate any calls to `.color()`/`.on()` etc? (Would the translation actually be in `.update()` so we don't have to duplicate it in anything that sets properties like this?)

Right now if you manually set a color, it overrides what you've got there. I'd rather control color and intensity values completely independently. i.e. set an intensity to 50% and then be able to change the color at will.

Resseguie:
And if it sets a state property, should a call to `.intensity()` without params return the current intensity value? (similar to `.color()`)

Resseguie:
Need to test that `.values()` returns the translated values that are actually sent to the pins.

mize:
> @Resseguie Right now if you manually set a color, it overrides what you've got there. I'd rather control color and intensity values completely independently. i.e. set an intensity to 50% and then be able to change the color at will.

and 
> @Resseguie And if it sets a state property, should a call to .intensity() without params return the current intensity value? (similar to .color())

+1 to both adding state property and following pattern of `.color()` if that will work for your intended uses of `.intensity()`. I'm thinking about similar setup for white point compensation in #811

scottgonzalez:
> Should .intensity() set state that is then used to translate any calls to .color()/.on() etc?

I personally find that confusing. A color of `#aaaaaa` is inherently not full intensity, so controlling the intensity completely separately from the color seems odd. Though I will admit that I know nothing about color theory, so maybe this is just a naive understanding.

mize:
@scottgonzalez This is a fun distinction! To better understand the desired behavior: Is intensity a property of the current color or of the LED?

**Example:**
1. current color is `#EEEEEE`
2. call `.intensity(.5)`
3. call `.color("EEEEEE")`.

Would you expect the resulting value to be...

a) `#EEEEEE`? `.intensity()` transforms current color only. Assumes user knows setting a brighter color will update intensity level of LED. In this case, I might consider intensity as an optional parameter of `.color()`, rather than a separate method of the LED.

b) `#777777`? `.intensity()` would be a setter for the current max intensity of the LED, applied as a persistent transform (similar to white point compensation), in this case 0.5:0.5:0.5 across RGB channels until a new intensity is set. Assumes user knows colors will be capped by the current intensity value of the LED.

scottgonzalez:
I view it as a modification of the current color. So I'd expect that if I ran `led.color("#333333").intensity()` that I'd get back `20`. And if I then ran `led.intensity(100).color()`, I'd get back `{ red: 255, green: 255, blue: 255 }`.

mize:
Thanks for the extra info! Extending this for how it might work with the white point correction idea...

```js
led.color("#333333").whitePoint("#FFE08C").intensity(100);
// #FFE08C, comment lines show hex value of actuals sent to pins

// white point value persists, but intensity doesn't
led.color("#FF9900");
// #FF8600

led.intensity(50);
// #7F4300

led.whitePoint("#FFFFFF");
// #7F4C00

led.intensity(100);
// #FF9900
```

That look correct and reasonable? :smiley:

scottgonzalez:
Yes, that's what I would expect. Though I think the values would be rounded, not floored, so after `led.intensity(50)`, you'd get `#804300` instead of `#7F4300`.

rwaldron:
To help clarify, what we're looking for is effectively "fade on and off", as if vertically here: 




(Grey at the top being "off")

I found that image [here](http://lodev.org/cgtutor/color.html#Change_the_Brightness_), which also might be useful.

scottgonzalez:
> To help clarify, what we're looking for is effectively "fade on and off"

That's what is implemented here (achieved by animating the intensity). The main question that @Resseguie was asking in https://github.com/rwaldron/johnny-five/pull/818#issuecomment-108480026 and @mize was asking in https://github.com/rwaldron/johnny-five/pull/818#issuecomment-108581244 is what do you expect `.color()` to return in the middle of the fade? My suggestion is that it returns the actual color currently being displayed. @mize seems to agree with his latest comment.

rwaldron:
> That's what is implemented here (achieved by animating the intensity). 

Cool, I haven't actually pulled for review and I'm a little brain-mushy right now, so I think I will wait until later or tomorrow AM

> what do you expect .color() to return in the middle of the fade?

The color it currently is. As long as it can get back to the original color, then that's ok

lib/led/rgb.js
```diff
@@ -269,7 +269,7 @@ RGB.prototype.color = function(red, green, blue) {
     // not a reference to the state object itself.
     colors = this.isOn ? state : state.prev;
     return RGB.colors.reduce(function(current, color) {
-      return (current[color] = colors[color], current);
+      return (current[color] = Math.round(colors[color]), current);
```
rwaldron: Does it matter if this is floor or ceiling?

lib/led/rgb.js
```diff
@@ -405,4 +405,17 @@ RGB.prototype.stop = function() {
...
+  var state = priv.get(this);
+  var currentIntensity = RGB.colors.reduce(function(current, color) {
+    return Math.max(state[color], current);
+  }, 0) / 255;
+  var scale = 1 / currentIntensity * Math.min(100, Math.max(0, intensity)) / 100;
+  var scaledColors = RGB.colors.reduce(function(current, color) {
+    return (current[color] = state[color] * scale, current);
+  }, {});
+
+  this.update(scaledColors);
```
rwaldron: `return this`

lib/led/rgb.js
scottgonzalez: If it's floor, then it's likely that many colors won't be able to reach 0xff at 100% intensity. If it's ceiling, it should be fine, since anything at 0% intensity will reach 0x00. I figured rounding was better than ceiling, since a value like 32.1 should end up as 32 instead of 33.

lib/led/rgb.js
scottgonzalez: Yup. There's also the getter. I just wanted feedback on the approach (which turned out to generate lots of discussion), because I did the full implementation.

scottgonzalez: *before* not because

mize:
> @scottgonzalez what do you expect .color() to return in the middle of the fade? My suggestion is that it returns the actual color currently being displayed. @mize seems to agree with his latest comment.

I just rubber-ducked the question of whether to keep intensity as a state property, and duh, the "right" answer is usually use-case-specific, so I've come around to Scott's flexible approach. It probably makes sense for the bulk of use cases, and avoids some potentially unexpected side effects of applying intensity transforms in `.color(value)` calls. Use cases I was considering are below for posterity.

Also, this is a hell of a good time. You all rock for diving into this discussion, and we are going to have the prettiest light shows around.

**Use cases**

- Could we change hue, maintaining relative intensity, as we fade? Example: shift from 50% pure red to 50% pure blue before continuing to fade out...
    - No intensity value kept in state? No problem, the consuming code just needs to get `.intensity()` and compensate by setting `led.color("000080")`.
    - Intensity in state? `led.color("0000FF")` would be transformed to the proper current intensity, but users would need to be aware they're really only using the hex value as a hue proportion at that point.
- Could I simultaneously update hue and intensity, say from 50% pure blue to 100% pure red?
    - No intensity in state? Yep, just call `led.color("FF0000")` to update.
    - Intensity in state? Eh, we'd need to chain here. `led.intensity(100).color("FF0000")` Perceptually, might be pretty close, but not as smooth as one update.

- What happens if you pause the fade, use `.off()`, then `.on()` and resume fade? If I'm reading the `.update()` method correctly, it looks like the previous value gets captured, regardless of which method calls it. Awesome. :+1:

scottgonzalez:
I've finished the implementation. This is ready for a full review.

Resseguie:
So I missed some of the earlier discussion, apologies, been away from keyboard a bit this weekend.

I read through the arguments for the current implementation and I'm still not convinced. But I'm just one vote. :) I'm coming at this from the perspective of having run a lighting console regularly for events.

The way it's implemented now probably makes sense for programmers. ;) But in the lighting world, "intensity" (the term I suggested Rick use here), is typically thought of as a fader. I set my stage light colors (say "#ff0000") and then I can fade them in and out by changing the intensity slider. If the slider is at 25%, it's still "red" just dimmer. If that slider is at 50% and I change the color to blue ("#0000ff") then the stage lights look blue but appear to only be at half their full brightness. I don't think of "intensity" as literally setting the intensity of the LEDs RGB pins. I think of it in terms of "how bright is the LED, respecting the currently set color."

You can still do that in the implementation here, it just seems convoluted to me.

In my head, I had thought that we'd have three methods:
`color()`
`whitepoint()`
`intensity()`

Each of those would store a value in the state, say '#ff0000', '#0000ff' (exaggerating), and '50' respectively. And when any of those values is changed, the `update()` function is called to calculate the actual values to send to the LED, which in operator overload shorthand pseudocode would be `color * whitepoint * intensity`.

At that point, `.color() === '{red: 255, green: 0, blue: 0}'`, `.whitepoint() === '{red: 0, green: 0, blue: 255}'`, and `.intensity() === 50` still as originally set by the user. But `.values()` would return an object that represents the calculated values as above, i.e. the "actual" current values of the RGB pins.

My $.02.

Resseguie:
@mize you raise a good point about `.fade()` in relation to `.intensity()` in particular.

Would it just fade from 0 to the specified intensity? (Either implementation, my suggestion or Scott's.)

In my way of thinking, fade would do it's thing, assuming it's calculating a fade from 0-100% no matter what. But then in `.update()` that resulting value would be scaled by the current intensity. (That's how the Master intensity slider would work on a lighting console, for example.)

Resseguie:
@mize 

> Intensity in state? Eh, we'd need to chain here. led.intensity(100).color("FF0000") Perceptually, might be pretty close, but not as smooth as one update.

But in my way of thinking, this version is much more obvious what it's doing. You'd have 50% blue using:

`led.intensity(50).color("0000FF");`
then change it to full red with
`led.intensity(100).color("FF0000");`

But in the most common case, you'd likely set it to a single color:
`led.color("FF0000");`

then change intensity programmatically as needed to make it "dimmer" or "brighter" without having to do hex color math.

rwaldron:
@Resseguie thanks for the feedback here, I was struggling to figure out a way to explain what we were looking for w/r to intensity.

Resseguie:
And in my example above with the 50% blue:

`led.intensity(50).color("0000ff");`

You should then be able to toggle it on and off, maintaining "50% blue".

And if you did a fade or pulse, it would fade between 0% blue and 50% blue (in reality, fading between 0-100% blue but scaled by the 50% intensity.)

Resseguie:
@mize 
> Also, this is a hell of a good time. You all rock for diving into this discussion, and we are going to have the prettiest light shows around.

Yeah, I love having a group of competent capable people to argue and discuss ideas with. Means we usually end up with a solution that's twice as good when we're done!

scottgonzalez:
> But in the lighting world, "intensity" (the term I suggested Rick use here), is typically thought of as a fader. I set my stage light colors (say "#ff0000") and then I can fade them in and out by changing the intensity slider. If the slider is at 25%, it's still "red" just dimmer. If that slider is at 50% and I change the color to blue ("#0000ff") then the stage lights look blue but appear to only be at half their full brightness.

Are you actually using a light which can change color digitally? Or are you switching which light you're using and they have gels in front of them? Basically, in the environment you're referring to, is it possible to have colors such as #33aa00 and #333333 without using the fader?

Resseguie:
Yeah, I can select an RGB LED channel and assign it a color. Then independently, I can use the slider to change the "brightness" of that color.

mize:
I think we have two conflicting color models at work. Please forgive and correct any imprecision here:

The current implementation of `led.color(input)` uses an RGB color model, which can affect both intensity and hue simultaneously, so you're setting absolute values for each channel (intensity) and the proportion among the three gives you the resulting hue.

**@scottgonzalez's version**
`led.intensity(input)` scales the output until a subsequent call to `led.intensity(input)` or `led.color(input)` is made with a different intensity. His version preserves *absolute* RGB values of the requested input when `led.color(input)` is called.

**@Resseguie's (and my earlier) version**
`led.intensity(input)` is a property of the LED, and is applied as a transform to any subsequent `led.color(input)` calls. In this case, it reduces the requested color to a *relative* ratio, to be transformed by the current `intensity`, resulting in a hybrid RGB/HSL color model, which "breaks" the expectation that `led.color(input)` functions like other colors we set with RGB in CSS. 

If you take the second route, it might be appropriate to allow the user to decide whether intensity scales further color/hue updates. Two ideas on that front:
- Add an optional boolean argument to `led.intensity(input, persist)`, to control whether it affects further updates.
- Add an `led.hue(input)` method and allow the `led.color(input)` method to perform as someone familiar with RGB colors would expect. So you'd have...
  - `led.color(input)` - RGB, outputting absolute values
  - `led.intensity(input)` - scales current output and keeps intensity as a state property
  - `led.hue(input)` - HSL, updates hue while respecting the current intensity from state

Is that helpful at all?

scottgonzalez:
I think many users will be confused by the fact that calling `led.color(value)` may not result in that value being set because of a previous call to `led.intensity()`. I'm inclined to say that we should not have a lot of logic around color systems (like allowing hue to be set) and instead should just provide the basic RGB functionality and defer to full-feature color libraries for more advanced use cases.

Resseguie:
> I think many users will be confused by the fact that calling led.color(value) may not result in that value being set because of a previous call to led.intensity()

Maybe we need to ask users and see which is most natural for majority? I fully admit I could be the minority.

But what you describe there is exactly what happens on a lighting console. I can set the RGB to a specific value, but if the intensity slider is at 50%, I'll only see 50% of the values that I entered output by the LED.

rwaldron:
> But what you describe there is exactly what happens on a lighting console. I can set the RGB to a specific value, but if the intensity slider is at 50%, I'll only see 50% of the values that I entered output by the LED.

Confirm—that's what I would expect "intensity" to do.

mize:
So if intensity sets state property that gets applied to any inputs, should `led.intensity()` and the upcoming `led.whitePoint()` act purely as getters/setters, and `this.update()` would be the one spot where the transforms are applied?

```js
```
Shorthand for `this.update(colors)` would be to write and preserve `colors * intensity * whitePoint`. 

If you called `this.update()` with no arguments (assume no color update but you called after updating whitePoint or intensity), it would write based on the current state values.

scottgonzalez:
@mize That's pretty much exactly what I'm in the middle of implementing :-)

scottgonzalez:
If the LED is off and you set the intensity, should it stay off or turn on?

mize:
@scottgonzalez woot! gold :star:

scottgonzalez:
See #821 for the alternative implementation that's been discussed.

Resseguie:
> If the LED is off and you set the intensity, should it stay off or turn on?

Good question. I would say turn on, where `.on()` and `.off()` would then really be shorthand for `.intensity(100)` and `.intensity(0)` respectively. You all agree?

scottgonzalez:
Ok, that's what I have implemented in #821.

PR #819, opened 2015-06-03
Motor doc: fix wrong comment

vrubiolo:
This commit also adds a oneliner stop() example.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,7 +85,7 @@ Docs should have tested and working sample code, [fritzing diagrams](http://frit
 
 The [wiki](https://github.com/rwaldron/johnny-five/wiki) contains documentation about the api, and contains code examples and fritzing diagrams.
 
-The [eg folder](https://github.com/rwaldron/johnny-five/tree/master/eg) contains working code examples that users can run out of the box. Each of these example files has an associated .md file in the [docs folder](https://github.com/rwaldron/johnny-five/tree/master/docs) that is generated from the eg files.
+The [eg folder](https://github.com/rwaldron/johnny-five/tree/master/eg) contains working code examples that users can run out of the box. Each of these example files has an associated .md file in the [docs folder](https://github.com/rwaldron/johnny-five/tree/master/docs) that is generated from the eg files. When contributing documentation updates for those code examples, modify the one in the [eg folder](https://github.com/rwaldron/johnny-five/tree/master/eg) and run `grunt examples` to regenerate its [docs folder](https://github.com/rwaldron/johnny-five/tree/master/docs) counterpart, then submit both of them as part of your contribution/PR. If you add a new documentation file, remember you will need to add it to `tpl/program.json` as well.
 
 
 <a name="sample-projects"></a>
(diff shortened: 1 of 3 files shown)


dtex:
@vrubiolo Thanks for this. Two notes:

1) The /docs files are generated from the /eg files so your change should be made in /eg and then run ```grunt examples``` to generate the updated docs/motor.md file

2) We already have a motor.stop() call in there after a two second delay. Calling stop() immediately after start() will make it appear as though nothing has happened.

vrubiolo:
Hi Donovan,

Thanks for the feedback and the advice. I'll rework the PR as you advised. It was not obvious to me which files to modify so I might submit a patch for that as well.

CONTRIBUTING.md
```diff
@@ -85,7 +85,7 @@ Docs should have tested and working sample code, [fritzing diagrams](http://frit
 
 The [wiki](https://github.com/rwaldron/johnny-five/wiki) contains documentation about the api, and contains code examples and fritzing diagrams.
 
-The [eg folder](https://github.com/rwaldron/johnny-five/tree/master/eg) contains working code examples that users can run out of the box. Each of these example files has an associated .md file in the [docs folder](https://github.com/rwaldron/johnny-five/tree/master/docs) that is generated from the eg files.
+The [eg folder](https://github.com/rwaldron/johnny-five/tree/master/eg) contains working code examples that users can run out of the box. Each of these example files has an associated .md file in the [docs folder](https://github.com/rwaldron/johnny-five/tree/master/docs) that is generated from the eg files. When contributing documentation updates for those code examples, modify the one in the [eg folder](https://github.com/rwaldron/johnny-five/tree/master/eg) and run 'grunt examples' to regenerate its [docs folder](https://github.com/rwaldron/johnny-five/tree/master/docs) counterpart, then submit both of them as part of your contribution/PR.
```
rwaldron: I guess this was way out of date, it needs to mention adding an entry to `tpl/programs.json` as well.

vrubiolo:
@rwaldron: thanks ! How about this one? I also updated the grunt commandline to be in line with the other grunt references in the file

CONTRIBUTING.md
```diff
@@ -85,7 +85,7 @@ Docs should have tested and working sample code, [fritzing diagrams](http://frit
...
 
-The [eg folder](https://github.com/rwaldron/johnny-five/tree/master/eg) contains working code examples that users can run out of the box. Each of these example files has an associated .md file in the [docs folder](https://github.com/rwaldron/johnny-five/tree/master/docs) that is generated from the eg files.
+The [eg folder](https://github.com/rwaldron/johnny-five/tree/master/eg) contains working code examples that users can run out of the box. Each of these example files has an associated .md file in the [docs folder](https://github.com/rwaldron/johnny-five/tree/master/docs) that is generated from the eg files. When contributing documentation updates for those code examples, modify the one in the [eg folder](https://github.com/rwaldron/johnny-five/tree/master/eg) and run `grunt examples` to regenerate its [docs folder](https://github.com/rwaldron/johnny-five/tree/master/docs) counterpart, then submit both of them as part of your contribution/PR. If you add a new documentation file, remember you will need to add it to `tpl/program.json` as well.
```
rwaldron: :+1: Thanks!

PR #821, opened 2015-06-10
Led.RGB: Add intensity method

### LINKED ISSUE #737: Add Led.RGB.intensity()

scottgonzalez:
Fixes #737 

Alternative to #818, storing the intensity as separate state from color.

### DIFF
--- a/lib/led/rgb.js
+++ b/lib/led/rgb.js
@@ -1,5 +1,6 @@
 var Board = require("../board.js");
 var nanosleep = require("../sleep.js").nano;
+var __ = require("../fn.js");
 
 var priv = new Map();
 
@@ -346,8 +347,7 @@ RGB.prototype.color = function(red, green, blue) {
       throw new Error("Led.RGB.color: invalid color ([" + [update.red, update.green, update.blue].join(",") + "])");
     }
 
-    // constrain to [0,255]
-    value = Math.min(255, Math.max(0, value));
+    value = __.constrain(value, 0, 255);
     update[color] = value;
   }, this);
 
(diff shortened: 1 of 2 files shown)


Resseguie:
Captures my version of "intensity" well.

Only discussion point for me is the behavior when you call `led.off()` then 'led.intensity(50)`, is it still "off" and waiting to be turned "on" and shown at 50%, or are `.off()` and `.on()` really shorthand for `.intensity(0)` and `.intensity(100)` respectively? (And thus, the led would immediately be lit at 50% when `.intensity(50)` is called, regardless of on/off state.)

scottgonzalez:
@rwaldron Did you get a chance to look at this?

dtex:
@Resseguie Motor has a similar thing in that a motor will remember it's speed when stopped so that if started later (with no params) it will resume at the same speed as before. I think you need to store the intensity independently of on/off state.

Resseguie:
@dtex good point, thanks!

lib/led/rgb.js
```diff
@@ -405,4 +419,18 @@ RGB.prototype.stop = function() {
...
 };
 
+RGB.prototype.intensity = function(intensity) {
+  var state = priv.get(this);
+
+  if (arguments.length === 0) {
+    return state.intensity;
+  }
+
+  state.intensity = Math.min(100, Math.max(0, intensity));
```
dtex: This pattern is also defined in fn.js. https://github.com/rwaldron/johnny-five/blob/master/lib/fn.js#L48-L50 Maybe we should call that for consistency and readability?

scottgonzalez: done

Resseguie:
@mize note it's ready for whitepoint now if you're still interested in adding that.

Resseguie:
#847 is an example demonstrating use of `intensity`

mize:
@Resseguie thanks, will do!

PR #823, opened 2015-06-13
Fixing broken links and images

pierceray:
I ran a site crawl on http://johnny-five.io and there were several broken links in the example files.  I've updated them as best I know.  They probably need verified that I picked the right images.

* Fixing broken link to the MPU-6050
* fixed broken images for grove temperature and rotary
* updated examples


rwaldron:
Thanks @pierceray!

PR #825, opened 2015-06-14
Updated the tinkerkit examples with working website.

pierceray:
The TinkerKit examples have links to the parts needed for the examples.  The http://www.tinkerkit.com website went dark around mid-2014 it looks like.

I updated to the http://tinkerkit.tihhs.nl/ domain and all of the examples now link appropriately.

### DIFF
--- a/docs/tinkerkit-accelerometer.md
+++ b/docs/tinkerkit-accelerometer.md
@@ -29,7 +29,7 @@ board.on("ready", function() {
   //
   // Devices:
   //
-  // - Dual Axis http://www.tinkerkit.com/accelerometer/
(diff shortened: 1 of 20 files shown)


rwaldron:
@pierceray this is totally awesome and I really appreciate your help on these docs issues.

Now for annoying request ;)

@resseguie redesigned the programs.json data scheme this spring and it afforded us the opportunity to include specific links for each example. Would you be willing to move the links in these examples to their appropriate entry in programs.json, then remove the content from the individual JS files?

pierceray:
I can do it, but it might take me a day or two to get to it.  Would you want to merge this to remove the broken link, and do that separately?

PR #836, opened 2015-06-21
Update motor-hbridge wiring

dtex:
Needed to wire up the enable pin on the unused side of the IC.


dtex:
@Resseguie If this matches your experience with the L293D and looks good, I think you can go ahead and land it.

Resseguie:
I'll double check it and then land. I had the second enable pin wired, but not the opposite grounds. That might be why I was getting strange results in my original tests, too. I'll check it right away...

Resseguie:
Ok, works fine. I'll land. thanks!

FYI, the reason I _thought_ I was getting strange behavior on the reverse is that I didn't notice that the example set reverse using a value of 50 instead of 255. It was so slow I thought it was just stuttering. 

Think I'll put in a PR to just change that to 255 so it's obvious.

Resseguie:
@rwaldron is there a process to notify someone that the J5 website needs to be updated when examples like this are fixed? Or do you just want to let it get picked up next time the site happens to get updated?

PR #837, opened 2015-06-21
make fw/rev both 255

Resseguie:
Make forward and reverse in the example both use a value of 255. 50 was so slow that it made it seem like the motor was just stuttering and not working properly in reverse.


dtex:
LGTM. Landing

PR #839, opened 2015-06-25
updated TinkerKit examples to use programs.json

pierceray:
updated TinkerKit examples to use programs.json
Alphabetized the tinker kit list
Fixed some mislinked locations.


pierceray:
This is a follow up for #825

PR #843, opened 2015-06-30
Expand test cases for the sensor class, and fix median value handling

mMerlin:
Restructure analog sensor shape test to check that all and only expected properties and methods exist
Verify that expected default property values are set
Check that data events are emitted at the expected time tick
Check that median processing give the expected results and emitted events


rwaldron:
This is awesome—thanks!

PR #850, opened 2015-07-03
change rgb-anode eg to use same pins as rgb cathode

Resseguie:
just for consistency between examples. (plus diagram has less crossy wires.


Resseguie:
This also puts the diagram in line with the code in the wiki documentation.

Issue #854, opened 2015-07-07
changing the frequency of events for an existing device instance


mMerlin:
Mentioned in a comment for PR #844, and confirmed by more recent additions to test/sensor.js, changing the freq property does not change the interval between data (or other) events.  A quick check of other component code seems to show that same problem everywhere.  It also showed several different default freq values.
* 20 : Pin
* 25 : Barometer, Motion, Pir, Proximity, Sensor, Temperature
* 50 : IR.Reflect, IR.Proximity (when instance of IR)
* 100 : Magnetometer, Sonar, Temperature (specific driver), Wii.Nunchuck, Wii.Classic
* 500 : IMU, IR, Wii

The main query here is, what is the intended / appropriate functionality when changing the freq property of an existing device instance?

Looking over the code to collect information about the existing functionality, I noticed a few other things:

expander.js uses state.frequency and opts.frequency versus freq property used by most other components.  Should some alignment (or aliasing) be done to get that to match?

compass.js has a comment showing creation with a freq value of 50, and another comment that says the default is 66 ms.  I do not see any actual references to the freq property in the code.  It appears that the comments are cut and pasted from the code for whatever component was used as the base for creating the Compass component.  joystick.js is similar.  A comment, but no apparent property.
Magnetometer is an instance of Compass, and passes a specific value for the freq property, but I do not see where that gets used.  The code that emits (compass) data and change events does not seem to use it.  Those are sent directly by the callback function passed through initialize to io.i2cRead.  No throttling is being done.  Does compass need a structure similar to sensor, to accumulate readings, then send only a filtered value?

rwaldron:
> changing the freq property does not change the interval between data (or other) events. A quick check of other component code seems to show that same problem everywhere.

It wasn't meant to—the property should be readonly and value storage moved to internal state. To update the interval times would require messy interaction with the internal intervals. 


> 20 : Pin

This is an attempt to matche the default sampling interval of StandardFirmata, but should just be updated to match the "Barometer, Proximity, Sensor, Temperature"

> Barometer, Motion, Pir, Proximity, Sensor, Temperature

- Barometer, Proximity, Sensor, Temperature... arbitrarily this might as well be the default. 
- Motion, Pir are deprecated and will no longer be updated for any reason.

> 100 : Magnetometer, Sonar, Temperature (specific driver), Wii.Nunchuck, Wii.Classic

-  Magnetometer should be updated to match "Barometer, Proximity, Sensor, Temperature"
[...]

PR #855, opened 2015-07-07
Demonstrate feasibility of change freq after instance creation

mMerlin:
This is a raw implementation.  I recommend NOT merging this until it has been cleaned up a bit.  This is just an easy way to show that changing the freq property, and associated event intervals is not that difficult.  Extending from conversation in #854.

The goal of this implementation was to minimize the changes shown by git diff while getting the added functionality to work.

### DIFF
--- a/lib/sensor.js
+++ b/lib/sensor.js
@@ -19,7 +19,24 @@ var priv = new Map(),
       // Flex Sensor (alias)
       "bend"
     ]
-  };
+  },
+  IS_TEST_MODE = !!process.env.IS_TEST_MODE;
+
+// To reduce noise in sensor readings, sort collected samples
+// from high to low and select the value in the center.
+function arrayMedian(input) {
+  var half, len, sorted;
+  // faster than default comparitor (even for small n)
+  sorted = input.sort(function(a, b) {
+    return a - b;
+  });
+  len = sorted.length;
+  half = Math.floor(len / 2);
+
+  // If the length is odd, return the midpoint m
+  // If the length is even, return average of m & m + 1
+  return len % 2 ? sorted[half] : (sorted[half - 1] + sorted[half]) / 2;
+}// ./arrayMedian(input)
 
 
 /**
@@ -37,14 +54,15 @@ function Sensor(opts) {
     return new Sensor(opts);
   }
 
-  // var value, last, min, max, samples;
-  var value, last, min, max, samples, intervalId, eventProcessing;
+  var value, last, min, max, samples, state, median, intervalId, eventProcessing;
 
   value = null;
   min = 1023;
   max = 0;
   last = -min;
   samples = [];
+  intervalId = null;
+  median = NaN;
 
   Board.Component.call(
     this, opts = Board.Options(opts)
@@ -62,7 +80,6 @@ function Sensor(opts) {
   this.io.pinMode(this.pin, this.mode);
 
   // Sensor instance properties
-  // this.freq = opts.freq || 25;
   this.range = opts.range || [0, 1023];
   this.limit = opts.limit || null;
   this.threshold = opts.threshold === undefined ? 1 : opts.threshold;
@@ -81,12 +98,11 @@ function Sensor(opts) {
   }.bind(this));
 
   // Throttle
-  // setInterval(function() {
-  // TODO: this function should be outside of the Sensor constructor function
-  // (with appropritate passed arguments), to avoid creating a separate copy
-  // (of the function) for each Sensor instance.
+  // TODO: The event (interval) processing function should be outside of the Sensor
+  // constructor function (with appropriate passed (and bound?) arguments), to
+  // avoid creating a separate copy (of the function) for each Sensor instance.
   eventProcessing = function () {
-    var err, median, low, high, boundary;
+    var err, boundary;
 
     err = null;
 
@@ -105,37 +121,21 @@ function Sensor(opts) {
       return;
     }
 
-
-    // To reduce noise in sensor readings, sort collected samples
-    // from high to low and select the value in the center.
-    median = (function(input) {
-      var half, len, sorted;
-      // faster than default comparitor (even for small n)
-      sorted = input.sort(function(a, b) {
-        return a - b;
-      });
-      len = sorted.length;
-      half = Math.floor(len / 2);
-
-      // If the length is odd, return the midpoint m
-      // If the length is even, return average of m & m + 1
-      return len % 2 ? sorted[half] : (sorted[half - 1] + sorted[half]) / 2;
-    }(samples));
+    // Keep the previous calculated value if there were no new readings
+    if (samples.length > 0) {
+      // Filter the accumulated sample values to reduce analog reading noise
+      median = arrayMedian(samples);
+    }
     // @DEPRECATE
     this.emit("read", err, median);
     // The "read" event has been deprecated in
     // favor of a "data" event.
     this.emit("data", err, median);
-    // If the median value for this interval is outside last +/- a threshold
-    // fire a change event
 
-    low = last - this.threshold;
-    high = last + this.threshold;
-
-    // Includes all aliases
-    // Prevent events from firing if the latest value is the same as (within
-    // the threshold of) the last value to trigger a change event
-    if (median < low || median > high) {
+    // If the filtered (median) value for this interval is at least ± the
+    // configured threshold from last, fire change events
+    if (median <= (last - this.threshold) || median >= (last + this.threshold)) {
+      // Include all aliases
       aliases.change.forEach(function(change) {
         this.emit(change, err, median);
       }, this);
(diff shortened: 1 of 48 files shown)


lib/sensor.js
```diff
@@ -223,6 +231,20 @@ function Sensor(opts) {
...
       }
     },
+    freq: {
+      get: function() {
+        return priv.get(this).freq;
+      },
+      set: function(newFreq) {
+        priv.set(this, {
+          freq: newFreq
+        });
```
rwaldron: This will pave over the existing state object

lib/sensor.js
```diff
@@ -159,14 +164,17 @@ function Sensor(opts) {
...
   // Create a "state" entry for privately
   // storing the state of the sensor
   priv.set(this, {
     booleanBarrier: opts.type === "digital" ? 0 : 512,
     scale: null,
-    value: 0
+    // value: 0
+    value: 0,
+    freq: opts.freq || 25
   });
```
rwaldron: Suggest changing: 

```js
priv.set(this, {
  booleanBarrier: opts.type === "digital" ? 0 : 512,
  scale: null,
  value: 0,
  freq: opts.freq || 25
});
```

To this: 

```js
var state = {
  booleanBarrier: opts.type === "digital" ? 0 : 512,
  scale: null,
  value: 0,
  freq: opts.freq || 25
};

priv.set(this, state);
```

lib/sensor.js
```diff
@@ -223,6 +231,20 @@ function Sensor(opts) {
         return this.constrained;
       }
     },
+    freq: {
+      get: function() {
+        return priv.get(this).freq;
```
rwaldron: With the change above, you'll now have access to to `state` from within here, so this can be: `return state.freq;`

test/sensor.js
```diff
@@ -8,7 +8,43 @@ var MockFirmata = require("./util/mock-firmata"),
     io: new MockFirmata(),
     debug: false,
     repl: false
-  });
+  }),
+  // TODO: query: does node.js / es6 have an object clone / deepcopy method?
```
rwaldron: Nope.

scottgonzalez: We could just expose another method from lodash in `fn.js` though.

scottgonzalez: Or if this is only needed in tests, just use underscore directly?

test/sensor.js
```diff
@@ -103,132 +143,612 @@ exports["Sensor - Analog"] = {
...
+    test.done();
+  },// ./limit: function(test)
+
+  freq: function(test) {
+    var spy = sinon.spy(),
+      newShape, newFreq, tickDelta, tickAccum;
+    test.expect(10);
+
+    this.sensor.on("data", spy);
+    test.deepEqual(this.defShape, getShape(this.sensor), "sensor instance properties should match default shape values");
```
rwaldron: Does this need to be in every test? Shape assertions should be in their own test.

test/sensor.js
```diff
@@ -103,132 +143,612 @@ exports["Sensor - Analog"] = {
...
+    tickDelta = 1;
+    this.clock.tick(tickDelta);
+    tickAccum += tickDelta;
+    test.strictEqual(spy.callCount, 1, "tick " + tickAccum + ": data event handler should be called first time at tick " + this.defShape.freq);
+
+    newShape = deepCopy(this.defShape);
+    newFreq = 35;
+    newShape.freq = newFreq;
+    this.sensor.freq = newFreq;
+    test.deepEqual(newShape, getShape(this.sensor), "sensor instance properties should match shape with new freq");
```
rwaldron: I'm not sure I understand what this is testing?

test/sensor.js
mMerlin: That was a note to myself, to be looked up.  Had not gotten around / back to it yet.  Answered now, so I do not need to do the research. :)

I had the code to do it myself, but thought new version might have it built in.  Much prefer to use 'native' code when it is available.  I was writing that section when I had no online access.

rwaldron: As @scottgonzalez said, I guess just use: https://lodash.com/docs#cloneDeep

rwaldron:
On the whole, this is a sound approach. I'm glad it didn't turn out too gnarly. My one bit of critical feedback is that the test changes are a bit a dense and the result is overwhelming. I'm having a hard time following what is being tested and how—it's not clear at a glance. I'll spend more time with it.

lib/sensor.js
mMerlin: I have not worked with javascript Map previously.  Transposing from other languages, I was thinking that 'this' was a context, not a key.  Instead, priv is a file scope variable (class static in other languages), and the instance ('this') is the key.  I actually implemented a similar structure in older (es5) javascript using a serial number incremented by the constructor as a key to create class and instance private storage.  Map and using a reference to the instance object makes it cleaner.  Time to write a test function, to make sure I get the capabilities straight.

test/sensor.js
mMerlin: In theory no.  The shape test should be sufficient.  This was just a sanity check to verify starting conditions before changing then verifying the change for the instance property.  Belt and suspenders.  I have not worked with nodeunit enough to 'just know' that things will be right at the start of the test case.  Getting there.

lib/sensor.js
rwaldron: > Since the existing unittests did not notice this, I also need to figure out what should be there to catch that.

After setting `sensor.freq` the values returned by any other getter would've been undefined—that might be useful for ensuring integrity of private data

test/sensor.js
rwaldron: Ok, I'll leave you to it then :)

test/sensor.js
mMerlin: It verifies that the data event does not get triggered before the default this.freq ms, that it does get triggered at the default time, that setting a new freq value changes (only the freq property value of) the instance state.  Following checks verify that future data events are triggered at the new interval, offset from when the freq value was changed.

rwaldron: Ok, I think I follow. I may just need to stop and eat dinner and maybe even come back to it tomorrow :)

mMerlin:
From the above comment, I need to add purpose / description comments to the tests.  As well as other things mentioned.  I DID start by saying this was pretty raw, though I was thinking more about the code than the tests.

rwaldron:
> add purpose / description comments to the tests

Sounds good :)

lib/sensor.js
mMerlin: Turns out that with the existing code, checking for that case is not easy.  The new freq get function is the only one that explicitly returns the private state value.  Everything else just uses the value as part of some calculation.  The *best* check I could come up with was to inject a value through analogRead, use booleanAt to set a barrier value that returns true for the value, then check that boolean was still true after setting the new freq property value.  Wiping out the booleanBarrier state value cased boolean to go from true to false.  Messy, and an incomplete check.

Is it *appropriate* to add a readonly state property to return the complete contents of the private data?  That could then be easily added to getShape() function I added for checking property changes.  Alternatively, and potentially more appropriate, defined property get functions could be created for each of the private state properties.  Something like: barrier, scale, val.  For my (testing) purposes, a read only state property would be easiest to use to provide full coverage.  The deepEqual check would automatically catch any unexpected changes.  New additions to the state properties would only need to be added to the defShape object to be fully included in all of the instance property validation done by the deepEqual tests .  That is actually what the shape checks were intended to catch.  Any unanticipated side effects.  But as is, it does not see changes in the instance private state properties.

mMerlin: How about:
```javascript
  var IS_TEST_MODE = !!process.env.IS_TEST_MODE;
  // ...
  if (IS_TEST_MODE) {
    Object.defineProperties(this, {
      state: {
        get: function() {
          return priv.get(this);
        }
      }
    });
  }
```
so the private information is only exposed for testing?

rwaldron: > so the private information is only exposed for testing?

This works for me

test/sensor.js
mMerlin: Currently only using the deep copy in the sensor test code.  When I get to adding the property to provide a readonly copy of the (to be) updated samples array, I'll be wanting to use it there as well.  Seems the simplest way to create a copy of an array of objects.

mMerlin:
Unposted changes are looking pretty good.  Just going back to add better documentation for the test cases, then I can commit and update the PR.

mMerlin:
PR branch updated earlier in the week.  Anyone had a look at the changes?

rwaldron:
@mMerlin bah, github doesn't ping when a PR is updated—so I had no idea :\

I will look at this weekend!

rwaldron:
I did a bunch of cleanups throughout the tests, making sure board instances were disposed of in all `tearDown`s. This will need a rebase with master, but it should be easy, just do: `git pull --rebase upstream master` (assuming you have a remote named `upstream` that points to this repo)

mMerlin:
I did the rebase.  It was not as simple as implied.  Had to resolved several merge conflicts, after looking at the current sources you committed. The merge made a mess in a few places, inserting my new function in the middle on one of your new functions.  Everything seems to be resolved now, but the PR does not look good.  It is showing 48 file changes, not just the 2 I changed.  The others seem to be the things you already did in the test refactoring.  I don't know what I/git did to cause that.  Looks like I may need to save the 2 changed files, delete the PR, refresh my branch, re apply the changes, and recreate the PR.  Confirm?  Other options?  Can I (and how) delete the pull request, or do you need to do it?  Got outside the places I have worked previously.  &lt;grin> again &lt;/grin>

rwaldron:
This looks like a merge, not a rebase. I will try to resolve it for you tomorrow.

mMerlin:
It *started* as a rebase, which said it was doing a rollback and replay.  Automatic merge failed on replay, and instructions said to fix that and continue.  I did, then tried to push.  That told me I needed to do a pull first, which generated more manual merges, before I could commit and push.

rwaldron:
>  I did, then tried to push. That told me I needed to do a pull first, 

This is the culprit, next time just force push your branch.

rwaldron:
Also, I see the function issue you mentioned

rwaldron:
Landed a79f2471c5c775009d1c9752b1542cb0a36e1e72

Issue #856, opened 2015-07-08
Test issues


dtex:
There is more than one so I'm just going to add them to this issue as I find them.

**Issue 1:**
If the readyWithRepl test doesn't run: https://github.com/rwaldron/johnny-five/blob/master/test/board.js#L136-L154 then the repl.js test fails.

rwaldron:
repl.js is missing a call to `Board.purge()` and `Repl.isBlocked` should be reset to whatever it was before the test.

dtex:
I'll fix that. Also, the "explicit" test: https://github.com/rwaldron/johnny-five/blob/master/test/board.js#L47-L64 times out exactly like ioHasError did https://github.com/rwaldron/johnny-five/blob/master/test/board.js#L93-L114

dtex:
Should Board.purge() also be done in the teardown of test/board.js? I know calling it in repl.js wil fix this, but I would think tests should clean up after themselves.

rwaldron:
Yes absolutely

dtex:
Hmm, Board.purge was already in the teardown. I assumed it cleared Repl.isBlocked when called, but I see now that would not be the case. In fact, I don't see how to unblock the repl with passing ```repl: true``` in a new board opts.

rwaldron:
Add `Repl.isBlocked = false` in the setup?

dtex:
I'm confused. It doesn't look like Repl.isBlocked actually does anything. 

Now I see that if both ```readyWithNoRepl``` and ```readyWithRepl``` run, the test does in fact clean up after itself because the second one passes repl:true in the opts so I should probably just leave that alone.

I added the missing Board.purge() in my upcoming PR.

Still have that timeout on the explicit test. Should I comment that one out too?

rwaldron:
While tracking down the remaining issue with `explicit`, I found a mountain of leaky tests in ESC.

dtex:
Closed with #862

rwaldron:
Looks like I eliminated the need for it and then forgot that I did that 0_o

PR #857, opened 2015-07-08
Test Issues

dtex:
Explicit test is timing out so I’ve commented it out for now. Addresses #856

The test just makes sure the sp object is being used on the board instance and moves on but leaves sp open and trying to connect.

If we could abort the attempt to connect that would be better.

### DIFF
--- a/test/board.js
+++ b/test/board.js
@@ -1,8 +1,8 @@
 require("es6-shim");
 
 
-var SerialPort = require("./util/mock-serial").SerialPort,
-  MockFirmata = require("./util/mock-firmata"),
+var MockFirmata = require("./util/mock-firmata"),
+  // SerialPort = require("./util/mock-serial").SerialPort,
   five = require("../lib/johnny-five.js"),
   sinon = require("sinon"),
   __ = require("../lib/fn.js"),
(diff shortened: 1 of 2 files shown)


test/board.js
```diff
@@ -1,8 +1,8 @@
 require("es6-shim");
 
 
-var SerialPort = require("./util/mock-serial").SerialPort,
-  MockFirmata = require("./util/mock-firmata"),
+var MockFirmata = require("./util/mock-firmata"),
+  // SerialPort = require("./util/mock-serial").SerialPort,
```
rwaldron: Can this be removed?

test/board.js
```diff
@@ -44,24 +44,24 @@ exports["Board"] = {
...
+  //     repl: false
+  //   });
 
-    test.equal(board.io.sp, sp);
+  //   test.equal(board.io.sp, sp);
 
-    test.done();
-  },
+  //   test.done();
+  // },
```
rwaldron: Super strange that this is causing issues.

rwaldron:
Landed https://github.com/rwaldron/johnny-five/commit/6b572899bda46b8229c4b30b68e0816c0cc19c51

PR #859, opened 2015-07-10
Wii: Cleanup

scottgonzalez:
Very late follow up to #726.

Unfortunately, I don't have a classic controller and I broke a pin on my nunchuk, so I couldn't test this with real hardware. I just ordered both on eBay, so if you want this verified with real hardware and nobody else can test in the next few weeks, I'll eventually be able to test this.

### DIFF
--- a/lib/wii.js
+++ b/lib/wii.js
@@ -2,8 +2,6 @@ var Board = require("../lib/board.js"),
   events = require("events"),
   util = require("util");
 
-var DEBUG = true;
-
 var Devices, Change, Update;
 
 // Event type alias map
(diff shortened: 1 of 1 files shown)


Resseguie:
What hardware did you get? I may pick some up too.

scottgonzalez:
http://www.ebay.com/itm/111596470005 and http://www.ebay.com/itm/371247475053

rwaldron:
Looking good so far, I'll _try_ to try it out this weekend.

Emphasis on _try_ because I just bought a bebop drone ;)

PR #862, opened 2015-07-11
When using mock-serial we should always assume success

dtex:
So this setTimeout was being kicked off even though we're using mock-serial and that's what was eventually causing our unspecified error.

### DIFF
--- a/lib/board.js
+++ b/lib/board.js
@@ -406,17 +406,19 @@ function broadcast(err, type, io) {
     // If "ready" hasn't fired and cleared the timer within
     // 10 seconds of the connect event, then it's likely
     // there is an issue with the device or firmware.
-    this.timer = setTimeout(function() {
+    if (!IS_TEST_MODE) {
+        this.timer = setTimeout(function() {
 
-      this.error(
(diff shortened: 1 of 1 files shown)


dtex:
This was the wrong way. We need to just emit the right event(s) to cancel the timeout. I'll try and get a new PR tomorrow.

lib/board.js
```diff
@@ -406,17 +406,19 @@ function broadcast(err, type, io) {
...
+        this.error(
+          "Device or Firmware Error",
+          "A timeout occurred while connecting to the Board. \n\n" +
+          "Please check that you've properly flashed the board with the correct firmware.\n" +
+          "See: https://github.com/rwaldron/johnny-five/wiki/Getting-Started#trouble-shooting"
+        );
+
+        this.emit("error", new Error("A timeout occurred while connecting to the Board."));
+      }.bind(this), 1e4);
+    }
```
rwaldron: LGTM

dtex:
Changed my mind. This is fine and I'm just being ridiculous. This conditional is used in many places so NBD.

Issue #863, opened 2015-07-11
Use new Travis-CI container based infrastructure


dtex:
See this blog post:

http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

I think this should be straightforward for us. Any reason not to?

rwaldron:
Maybe? tessel hit issues with installing USB drivers

dtex:
I'll try it out

rwaldron:
Sounds good

dtex:
Close by #864

PR #864, opened 2015-07-11
Upgrade to Travis container based infrastructure

dtex:
Seems to work fine. Also noticed it kicked off much more quickly than usual. Addresses #863


rwaldron:
LGTM

PR #866, opened 2015-07-15
Add Micro Magician examples and Shield Config

dtex:
Title says it all. Realized this probably doesn't count as a new board since it's Arduino Pro compatible. Just added examples in case someone comes here looking for J5 support on the device.

### DIFF
--- a/eg/micromagician-motor.js
+++ b/eg/micromagician-motor.js
@@ -1,0 +1,38 @@
+var five = require("../lib/johnny-five.js");
+
+var board = new five.Board({ debug: true});
+
+board.on("ready", function() {
+
+  // Create a new `motor` hardware instance.
+  var motorA = new five.Motor(five.Motor.SHIELD_CONFIGS.MICRO_MAGICIAN_V2.A);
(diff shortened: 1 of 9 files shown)


eg/micromagician-motor.js
```diff
@@ -0,0 +1,38 @@
+var five = require("../lib/johnny-five.js");
+
+var board = new five.Board({ debug: true});
+
+board.on("ready", function() {
+
+  // Create a new `motor` hardware instance.
+  var motorA = new five.Motor(five.Motor.SHIELD_CONFIGS.MICRO_MAGICIAN_V2.A);
+  var motorB = new five.Motor(five.Motor.SHIELD_CONFIGS.MICRO_MAGICIAN_V2.A);
```
rwaldron: `MICRO_MAGICIAN_V2.A` => `MICRO_MAGICIAN_V2.B`

rwaldron:
Once that typo is sorted out, go ahead and land it!

dtex:
Groovy, it's fixed. Travis failed on first try (in temperature.js) but passed with a restart. Something to keep an eye on I guess.

Issue #762, opened 2015-04-30
Controller initialization logic


scottgonzalez:
I noticed that many components support the same controllers and each component handles the controller initialization directly. For example, the PCA9685 controller is currently supported by:

* esc ([initialization block](https://github.com/rwaldron/johnny-five/blob/310cdd11a4331ff1552a69a49d12826ea5e0b0f0/lib/esc.js#L26-L45))
* servo ([initialization block](https://github.com/rwaldron/johnny-five/blob/310cdd11a4331ff1552a69a49d12826ea5e0b0f0/lib/servo.js#L53-L72))
* led ([initialization block](https://github.com/rwaldron/johnny-five/blob/310cdd11a4331ff1552a69a49d12826ea5e0b0f0/lib/led/led.js#L25-L44))
* motor ([initialization block](https://github.com/rwaldron/johnny-five/blob/310cdd11a4331ff1552a69a49d12826ea5e0b0f0/lib/motor.js#L128-L145))

The initialization blocks are mostly the same, but there are some inconsistencies:

* `esc` and `servo` use `i2cWriteReg()`; `led` uses `i2cWrite()`; and `motor` uses `sendI2CWriteRequest()`.
* `esc`, `servo`, and `led` set `0x70` for the prescalar; `motor` sets `0x3`.
* `led` and `motor` do additional setup during the controller intialization ([`led` logic](https://github.com/rwaldron/johnny-five/blob/310cdd11a4331ff1552a69a49d12826ea5e0b0f0/lib/led/led.js#L46-L56),[`motor` logic](https://github.com/rwaldron/johnny-five/blob/310cdd11a4331ff1552a69a49d12826ea5e0b0f0/lib/motor.js#L147-L150)).

I haven't looked at other controllers yet, but I assume similar behavior exists.

----

I'd recommend creating controller modules that contain the common logic and then utilizing those inside the components modules.

I don't know much about how these controllers work yet, so I'm not sure if the components can be mixed. Is the intent that only one type of component may be used on a specific controller? If so, is this actually a hardware limitation? If not, having any component-specific logic inside the driver conditional seems wrong, as the initialization is now dependent on the order in which the components are created.

scottgonzalez:
[...]

PR #876, opened 2015-07-24
[WIP] Prototype for DRYing out controller logic

scottgonzalez:
This is just to get some discussion going for #762. I've done some initial work for PCA9685 in `Led`/`Led.RGB`. Once we agree on the approach to take, I can do some work more and then update the tests.

There are a few questions written as TODOs in the controller file.

### DIFF
--- a/lib/controllers/pca9685.js
+++ b/lib/controllers/pca9685.js
@@ -1,0 +1,54 @@
+var nanosleep = require("../sleep").nano;
+
+function PCA9685(opts) {
+  if (!(this instanceof PCA9685)) {
+    return new PCA9685(opts);
+  }
+
+  this.board = opts.board;
+  // TODO: Just always require opts.io?
+  this.io = opts.io || this.board.io;
+  this.address = opts.address || 0x40;
+  // TODO: use this
+  this.pwmRange = opts.pwmRange || [0, 4095];
+
+  // TODO: This is the existing logic, but it seems odd.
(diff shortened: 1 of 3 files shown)


dtex:
Me gusto mucho.

Do we have a list of other controllers? 
How does this relate to expanders?
Does the PCA9685 ever accept anything besides 12-bit values? (this removes the ability to set PWM range)

scottgonzalez:
> Do we have a list of other controllers?

I can put a list together by searching for "var Controllers". I think many of them are actually component-specific so this abstraction wouldn't need to be applied to them. I'll go through and see which ones are used by more than one component.

> How does this relate to expanders?

At least based on the current design of Johnny-Five, users initialize the expanders themselves. This would only apply to the existing `controller` option within the individual components. Perhaps PCA9685 is just an odd case?

> Does the PCA9685 ever accept anything besides 12-bit values? (this removes the ability to set PWM range)

The PWM range will be supported once I get to a controller that doesn't use the full range. I know at least servo uses a different range.

rwaldron:
My plan was to use instances of `Expander` wherever possible; I have a bunch of work completed for migrating the PCA9685 (just that so far), but it's all in a git stash. I don't want another new PCA9685 thing.

rwaldron:
Out of the stash, committed, rebased, etc... https://github.com/rwaldron/johnny-five/commits/controllers-can-use-expanders

PR #880, opened 2015-07-30
Fix nanotimer dependency

Qard:
The nanotimer dependency was locked to a specific version.
This prevented newer patch releases supporting io.js from being used.


Qard:
Ugh...well, it fixed _my_ code anyway. Not sure why the tests don't like it. I'll have another look later. >.>

PR #881, opened 2015-07-30
Added LM335 temperature controller and sundries

ajfisher:
Updated branch to merge due to the crazy that was my last one with old previously committed changes.

Includes LM335 temperature controller, tests, documentation and circuit diagram. Note Fritzing component used is LM35 still but is has the identical footprint.


dtex:
Linking to #870

dtex:
I found an LM335 Fritzing part: https://github.com/lambert/fritzing-parts/blob/master/LM335.fzpz

I think if I PR'd an updated diagram to your repo it would create an unnecessary commit.

PR #889, opened 2015-08-24
Temperature: Add C, K, F values to the data event and temperature objects

gnarf:
I moved some stuff around in the test suite too.

Are there newer patterns for modules that this section should be brought up to date with?  It was actually a little difficult for me to work in this test suite, happy to do some cleanup here if there is work to do.

### DIFF
--- a/lib/temperature.js
+++ b/lib/temperature.js
@@ -392,7 +392,7 @@ function Temperature(opts) {
     this.toCelsius = opts.toCelsius || function(x) { return x; };
   }
 
-  Object.defineProperties(this, {
+  var propDescriptors = {
     celsius: {
       get: function() {
         return this.toCelsius(raw);
@@ -408,7 +408,13 @@ function Temperature(opts) {
         return this.celsius + 273.15;
       }
     }
-  });
+  };  
+  // Convenience aliases
+  propDescriptors.C = propDescriptors.celsius;
+  propDescriptors.F = propDescriptors.fahrenheit;
+  propDescriptors.K = propDescriptors.kelvin;
+
+  Object.defineProperties(this, propDescriptors);
 
   if (typeof this.initialize === "function") {
     this.initialize(opts, function(data) {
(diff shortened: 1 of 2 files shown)


rwaldron:
New pattern is to share a single set: 

- https://github.com/rwaldron/johnny-five/blob/master/test/button.js#L40-L55
- https://github.com/rwaldron/johnny-five/blob/master/test/button.js#L76-L88

Fix as you encounter

BrianGenisio:
Is there a reference to a conversation for why this PR exists?  What is the purpose for adding the aliases?  Just looking for context.

gnarf:
It's really tough to for me to remember how to spell fahrenheit in
particular. (I used right click spell check to fix that one)

I think temp.F temp.C temp.K looks pretty awesome as an API and will add a
lot of use and intuitiveness to some of the basic ardx style examples.

gnarf:
I also think it increases the immediate approachability for younger and/or
dyslexic experimenters

BrianGenisio:
Those reasons make sense.  I just wonder about the added confusion that comes with redundant API values.  Does it make more sense to deprecate the spelled-out values?  @rwaldron are there other examples that you can think of in J5 where values are aliased (copied) in the API?

BrianGenisio:
Other than the philosophical question of duplicating interfaces, the code looks reasonable.

Do you want to update the examples at all?  Should this be the preferred way of interfacing with the component?

You'll want to [update the Wiki](https://github.com/rwaldron/johnny-five/wiki/Temperature) after it lands.

frxnz:
@BrianGenisio `Proximity` has similarly aliased values, `inches` => `in and `centimeters` => `cm`. See https://github.com/rwaldron/johnny-five/blob/master/lib/proximity.js#L303-L334

rwaldron:
@gnarf and @frxnz's cases are good enough for me to support keeping both. As for examples, I'm not picky, but the docs definitely need updating. 

@BrianGenisio you good with that?

BrianGenisio:
:ship:

gnarf:
Added to wiki: https://github.com/rwaldron/johnny-five/wiki/Temperature/_compare/971c165bf24ec658b8a6fc9088a38fb672ff8c20...4f381164269ff071371ef0ab6188997a8565815d

PR #891, opened 2015-08-26
Temperature: add analog reference to TMP36 controller

gnarf:
This code adds an `aref` property to the `Temperature` object.  Defaults to whatever is defined on `this.io.aref` OR `5` if not defined at all.

This allows you to set the analog reference voltage for analog style temp controllers, first implementation I targeted the `TMP36` and will add to the rest of the analog controllers before we land this PR if this first one went well.

### DIFF
--- a/lib/temperature.js
+++ b/lib/temperature.js
@@ -188,7 +188,11 @@ var Controllers = {
     },
     toCelsius: {
       value: function(raw) {
-        return (raw * 0.4882814) - 50;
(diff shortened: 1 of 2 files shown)


rwaldron:
Looks good, but let's get them all: 

- https://github.com/gnarf/johnny-five/blob/aref/lib/temperature.js#L166
- https://github.com/gnarf/johnny-five/blob/aref/lib/temperature.js#L177

Thanks!

PR #894, opened 2015-08-26
Complete test conversion for analog controllers and aref

gnarf:
I'm not sure - but it seems that the math involved for these two controllers doesn't take aref into account at all.

### DIFF
--- a/lib/temperature.js
+++ b/lib/temperature.js
@@ -301,8 +301,6 @@ var Controllers = {
         var adcres = 1023;
         // Beta parameter
         var beta = 3975;
-        // 0°C = CELSIUS_TO_KELVIN K
-        var kelvin = CELSIUS_TO_KELVIN;
         // 10 kOhm (sensor resistance)
         var rb = 10000;
         // Ginf = 1/Rinf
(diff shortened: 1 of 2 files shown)


test/temperature.js
```diff
@@ -311,6 +311,62 @@ exports["Temperature -- ANALOG"] = {
...
+        controller: "GROVE",
+        pin: "A0",
+        freq: 100,
+        board: this.board
+      });
+
+      done();
+    },
+    shape: testShape,
+    aref: makeTestAnalogConversion({
```
gnarf: can someone verify that these two controllers should see no change when aref is different?

lib/temperature.js
```diff
@@ -311,7 +309,7 @@ var Controllers = {
         var tempr = 298.15;
 
         var rthermistor = (adcres - raw) * rb / raw;
-        var tempc = 1 / (Math.log(rthermistor / rb) / beta + 1 / tempr) - kelvin;
+        var tempc = 1 / (Math.log(rthermistor / rb) / beta + 1 / tempr) - CELSIUS_TO_KELVIN;
```
rwaldron: Cleeeeeean

PR #906, opened 2015-09-21
Adding support for Sparkfun TMP102 Temperature Sensor

BrianGenisio:
While testing the Particle I2C support, I tried to use this device but realized we didn't have support for it.  So I added support for it.


lib/temperature.js
```diff
@@ -206,6 +206,34 @@ var Controllers = {
       }
     }
   },
+
+  TMP102: {
+    initialize: {
+      value: function(opts, dataHandler) {
+        this.io.i2cConfig(opts);
+
+        this.io.i2cRead(0x48, 0x00, 2, function(data) {
```
rwaldron: Is the address limited to 0x48?

dtex: It is addressable but the docs are pretty abysmal. It's unclear what addresses are available. The TMP102 itself has 4 addresses but the breakout board from Sparkfun only seems to have two. :confused:

rwaldron: Awful. I will order one of these asap.

PR #907, opened 2015-09-22
Support Led.Digits with HT16K33 controller

### LINKED ISSUE #905: Support for 7-segment LED HT16K33 Backpack
Hi there! I have an [Adafruit 4-digit 7-segment display w/ I2C backpack](http://www.adafruit.com/products/879), and have been struggling to get it working. Originally I thought I could use `five.Led.Digits`, but the `controller: "HT16K33"` option seems to force the type to `five.Led.Matrix`. I dug into ledcontrol.js a bit and it looks like HT16K33 is assumed to be a matrix-only controller.

I finally did get things working:

```javascript
var five = require('johnny-five');
var Raspi = require('raspi-io');
var board = new five.Board({io: new Raspi()});

var digits = [0x3F, 0x06, 0x5B, 0x4F, 0x66, 0x6D, 0x7D, 0x07, 0x7F, 0x6F];

    // Create HT16K33 controller
    var matrix = new five.Led.Matrix({controller: 'HT16K33'});

    // Display 12:34
    matrix.displaybuffers[0][0] = digits[1];
    matrix.displaybuffers[0][1] = digits[2];
    matrix.displaybuffers[0][2] = 0xFFFF;
    matrix.displaybuffers[0][3] = digits[3];
    matrix.displaybuffers[0][4] = digits[4];
    matrix.writeDisplay(0);
});
```

The digits are from an [Adafruit python tutorial](https://github.com/adafruit/Adafruit-Raspberry-Pi-Python-Code/blob/master/Adafruit_LEDBackpack/Adafruit_7Segment.py).

I'm assuming there must be a better way to do this. Would you take a PR implementing `digit()` and `print()` for the HT16K33 controller, or would that even be the right approach? I skimmed through #665, which sounded like it would address my question as well.

Thanks!

donmccurdy:
Fixes #905.

Moves print() and digits() to parent controller, calling into delegated sendDigit() functions for controller-specific logic.

There were a few differences with the HT16K33 controller:

1. Column indexing is left-to-right, so I've added a `digitsOrder` property to the controller. I'm not sure which direction is actually more common.
2. The order of individual segments is reversed, so the constants in [DIGIT_CHARS](https://github.com/rwaldron/johnny-five/blob/master/lib/led/led-chars.js#L296) don't work. Rather than adding another set of constants for the HT16K33 ([example](https://github.com/adafruit/Adafruit-Raspberry-Pi-Python-Code/blob/master/Adafruit_LEDBackpack/Adafruit_7Segment.py#L18)) I've done some bit-flipping to reuse the original constants. Tested that on all values 0-9A-F. The lowercase letters look odd: not sure what expected behavior is.

Other changes:
- Adds tests for HT16K33 with segmented displays.
- Use `controller.digits` option when available, rather than assuming there are 8.
- Adds ":" support for ##:## displays. To use this, I might do something like:

```javascript
var digits = new five.Led.Digits({controller: 'HT16K33', digits: 5});
digits.print("12:00");
```

Tested against [Adafruit 4-digit 7-segment display with HT16K33 I2C backpack](http://www.adafruit.com/products/879).

I've not added other documentation or examples so far, but would be happy to add those if the changes themselves seem reasonable.


rwaldron:
@donmccurdy I'm at TC39 this week, I will be back home this weekend, but will be at Maker Faire, so I can review on Monday. Sorry for that delay and I appreciate your patience in the meantime :)

donmccurdy:
Thanks Rick – sounds good and have a safe trip, I'll be out over the weekend as well. I'm assuming this will need an example or two in any case.

rwaldron:
:clap:

Issue #912, opened 2015-09-29
Support the Photon Weather Shield


BrianGenisio:
This is the device:
https://www.sparkfun.com/products/13630

First pass, I plan to support 
- [x] Humidity Temperature
- [x] Humidity
- [x] Pressure
- [x] Altimeter
- [x] Pressure Temperature
- [ ] Multi Support for the shield itself?

Anything else that should be supported in the first pass?

rwaldron:
Are the second pass things built in?

BrianGenisio:
@rwaldron No.  As far as I can tell, the "second pass" things are not specifically called out.  They can be anything really, and just map to pins.  So they would get their own components based on what you decide to plug into those spaces.  Striking them.

BrianGenisio:
At first I had thought that this shield had a unifying I2C interface, but the more I read it, I realize that is not the case.  Really, there are two chips just connected to the I2C bus:
- HTU21D (Humidity and Temperature)
- MPL3115A2 (Pressure, Altimiter, and Temperature)

So, in reality, the devices should be for these chips specifically:
```javascript
var humidity = new five.Humidity({
  controller: "HTU21D"
});

var temperature = new five.Temperature({
  controller: "HTU21D"
});
```  

We can alias them as `PHOTON_WEATHER_SHIELD` to beef up the user experience... but really, this is all it is.

For `new five.Temperature({ controller: PHOTON_WEATHER_SHIELD'})`, we'll have to decide which temperature reading to take.  It is probably an arbitrary decision, but I'd probably go for HTU21D, because that is the one that is called out for having Temperature in the product on the website.

rwaldron:
There are lots of altimeter and humidity components that are commercially available, so let's go ahead and create two new component classes:

- Altimeter
  - BMP180 (implemented, just need to expose data)
  - MS5607 can be implemented
  - add more...
- Hygrometer (I don't mind if we also export an alias called Humidity \*)
  - SHT15
  - RHT03 via I2C backpack
  - add more...

Other Multi breakouts:

- BME280 


\* The existing `Barometer` class can also be aliased as `Pressure`, if we think that will be intuitive.

4math2379:
Hi , 

Waou ! Thanks for this information.
I will take a look .

Vesuviian

dougseven:
[...]

PR #916, opened 2015-10-04
HTU21D support

BrianGenisio:
Getting this up there so @dougseven can try it out.  Supports the HTU21D in this issue: https://github.com/rwaldron/johnny-five/issues/912

Still needs unit tests and circuit diagram

The examples that are interesting here are:

```
node eg/hygrometer-htu21d.js 
node eg/temperature-htu21d.js 
node eg/multi-htu21d.js 
```

**Not ready for merge until the following are included:**

- [x] Unit Tests
- [x] Fritzing Diagram


lib/hygrometer.js
```diff
@@ -0,0 +1,96 @@
...
+        var Multi = require("../lib/imu");
+        var driver = Multi.Drivers.get(this.board, "HTU21D", opts);
+        driver.on("data", function(data) {
+          dataHandler(data.humidity);
+        });
+      }
+    },
+    toRelativeHumidity: {
+      value: function(raw) {
+        return (125.0*(raw/65536)) - 6;
```
rwaldron: Is there a datasheet reference for these numbers?

lib/hygrometer.js
```diff
@@ -0,0 +1,96 @@
...
+  freq = opts.freq || 25;
+
+  if (opts.controller && typeof opts.controller === "string") {
+    controller = Controllers[opts.controller.toUpperCase()];
+  } else {
+    controller = opts.controller;
+  }
+
+  if (controller == null) {
+    controller = Controllers["ANALOG"];
```
rwaldron: Needs to throw when there is no default controller

lib/hygrometer.js
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    var data = {};
+    data.RH = data.relativeHumidity = this.relativeHumidity;
+
+    this.emit("data", null, data);
+
+    if (this.relativeHumidity !== last) {
+      last = this.relativeHumidity;
+      this.emit("change", null, data);
```
rwaldron: `this.emit("change", null, data);` => `this.emit("change", data);`

Trying to avoid making anymore places that have this mistake. (which reminds me: I need to update my patch for eliminating all of these)

lib/hygrometer.js
```diff
@@ -0,0 +1,96 @@
...
+
+  setInterval(function() {
+    if (raw === undefined) {
+      return;
+    }
+
+    var data = {};
+    data.RH = data.relativeHumidity = this.relativeHumidity;
+
+    this.emit("data", null, data);
```
rwaldron: `this.emit("change", null, data);` => `this.emit("change", data);`

lib/imu.js
```diff
@@ -14,6 +15,64 @@ var priv = new Map();
...
+        };
+      
+        var readCycle = function(isTemp) {
+          var register = isTemp ? this.REGISTER.TEMPERATURE : this.REGISTER.HUMIDITY;
+
+          io.i2cReadOnce(address, register, READLENGTH, function(data) {
+            var msb = data[0];
+            var lsb = data[1];
+
+            var value = (msb << 8) | lsb;
```
rwaldron: Replace lines 46-49 with: 

```
var value = int16(data[0], data[1]);
```

lib/temperature.js
```diff
@@ -287,6 +287,22 @@ var Controllers = {
...
+        var Multi = require("../lib/imu");
+        var driver = Multi.Drivers.get(this.board, "HTU21D", opts);
+        driver.on("data", function(data) {
+          dataHandler(data.temperature);
+        });
+      }
+    },
+    toCelsius: {
+      value: function(raw) {
+        return (175.25*raw/65536)-46.85;
```
rwaldron: Also needs datasheet refs

rwaldron:
Re: Fritzing, the breakout component is here: https://github.com/sparkfun/Fritzing_Parts/blob/master/products/12064_sfe_HTU21D_breakout.fzpz

BrianGenisio:
Ok, as far as I know, this PR is complete.  Hoping to finish the other sensor tonight or tomorrow.

#cc @rwaldron @dougseven.

rwaldron:
I will dig in tomorrow!

eg/hygrometer-htu21d.js
```diff
@@ -0,0 +1,16 @@
+var five = require("../lib/johnny-five.js");
+var board = new five.Board();
+
+board.on("ready", function() {
+  var hygrometer = new five.Hygrometer({
+    controller: "HTU21D"
+  });
+
+  hygrometer.on("data", function() {
+    console.log(this.relativeHumidity + " %");
```
rwaldron: Do these sensors produce relative humidity and absolute humidity?

I found this useful explanation: 

- **Absolute Humidity**, often just referred to as 'the humidity', is a measure of the actual amount of water vapour in a particular sample of air: measured as a partial pressure (vapour pressure/hPa or millibars); a mixing ratio (gm water vapour/kg of dry air), dew point etc.
- **Relative Humidity** - expressed commonly as a percentage value, is the ratio of the actual amount of water vapour present in a sample (the Absolute Humidity) to that amount that would be needed to saturate that particular sample.

rwaldron: Instead of being lazy, I went and did some quick reading. I learned that this doesn't produce an absolute humidity. https://www.adafruit.com/datasheets/1899_HTU21D.pdf

rwaldron:
> I will dig in tomorrow!

I couldn't wait.

rwaldron:
This is slick, nice work @BrianGenisio 

We need a new API docs page, which I can probably work on tomorrow

rwaldron:
Landed and released in [v0.8.97](https://github.com/rwaldron/johnny-five/commit/1dc4cf0f1d2ff64f48636ff6485b10932592cad0)

BrianGenisio:
Sweet. @dougseven give it a whirl when you get the chance. Update J5 to
0.8.97 and you should be good to go. Pressure/altitude on the heels.

dougseven:
@BrianGenisio Thanks! We are hacking on the lab solution this week and will be ready to run a few hundred people through it next week. 

It doesn't look like the following is supported:

```javascript
    var hygrometer = new five.Multi({
      controller: "HTU21D",
      threshold: 1 // Fire a change event if the value changes by 1 or more
    });
    
    hygrometer.on("change", function() {
      console.log("Thermometer");
      console.log("  celsius           : ", this.temperature.celsius);
      console.log("  fahrenheit        : ", this.temperature.fahrenheit);
      console.log("  kelvin            : ", this.temperature.kelvin);

      console.log("Hygrometer");
      console.log("  relative humidity : ", this.hygrometer.relativeHumidity);
      console.log("======================================");
    });
```
It doesn't error out, but it invokes even if the value is within the threshold amount. Am I mistaken?

BrianGenisio:
@dougseven I did some spelunking on this.

That's correct.  `Multi` does not support `threshold`.  What would it mean?  All values above a threshold?  Any value above a threshold?  And since the values are at a different scale, it gets complicated.  A filter inside your "change" event might be more customizable.

FWIW, none of the `Barometer`, `Hygrometer`, `Altitude`, or `Temperature` classes support `threshold` either.  The documentation for the "change" event on `Temperature` suggests that there is a threshold, but that is a documentation bug.  IIRC, it wasn't implemented on those classes either for the scaling issue... which `Temperature` value to you filter for `threshold`?  It's better done at the user level.

It appears `threshold` is only implemented in three input classes: `Sensor`, `Joystick`, `Wii`.  All three of these are essentially unitless sensors, so I'm guessing that the general philosophy is that we can do it easily for unitless data, but for data with varying unit scales, threshold filtering is left to the user.  Perhaps @rwaldron can tell me if I got this right ;)

rwaldron:
@BrianGenisio you nailed it. In fact, `threshold` should be removed from everything but `Sensor`, then an audit can be done to figure out if it makes sense anywhere else. `Wii` is thorn. I have no idea why I ever included `threshold` in `Joystick` 0_o

PR #917, opened 2015-10-05
Removed unused variable from the example for Temperature sensor TMP36

jukra:
Removed unused variable sensor


BrianGenisio:
Yeah, I noticed that the other day.  LGTM

rwaldron:
@jukra @dtex @BrianGenisio thanks!!

PR #918, opened 2015-10-06
Support MPL3115A2 Pressure/Altitude/Temperature sensor

BrianGenisio:
This implements the three readings from the [MPL3115A2](https://www.adafruit.com/products/1893) Barometer/Altimiter/Temperature sensor.

Code is ready for review and test, but it still needs the following to be considered complete:

- [x] Fritzing Diagrams
- [x] Unit Tests

#cc @dougseven


dougseven:
@BrianGenisio @rwaldron - I tested this with the Photon and the SparkFun Weather Shield - works great! Thank you.

BrianGenisio:
Alright, I think this PR is in good shape.  Will also need API docs updated after it lands.

Also, @rwaldron was seeing some really odd altitude values when testing near sea level, but I'm not sure if he is still seeing them?  If so, let me know what the MSB, CSB and LSB are for those values.  I'll write a unit test and work it out.

#cc @dougseven

PR #919, opened 2015-10-06
Collections: refactor to abstract Collection class

bsh314:


rwaldron:
Thanks! This patch was already landed ;)

PR #923, opened 2015-10-13
Add support for Tessel Servo module to Servo and LED

dtex:
Notes: Default address is different than most PCA9685, 0x73 instead of 0x40. Also, pins are mis-labeled 1-16 instead of 0-15.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,7 @@ To interactively navigate the examples, visit the [Johnny-Five examples](http://
 ### LED
 - [LED](https://github.com/rwaldron/johnny-five/blob/master/docs/led.md)
 - [LED - PCA9685](https://github.com/rwaldron/johnny-five/blob/master/docs/led-PCA9685.md)
+- [LED - Tessel Servo Module](https://github.com/rwaldron/johnny-five/blob/master/docs/led-tessel-servo-module.md)
 - [LED - Blink](https://github.com/rwaldron/johnny-five/blob/master/docs/led-blink.md)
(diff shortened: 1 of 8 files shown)


rwaldron:
If you'd like to help me finish this: https://github.com/rwaldron/johnny-five/tree/controllers-can-use-expanders, then the change would only need to be made in `lib/expander.js`. Otherwise, all of these need to be updated: 

- lib/esc.js
- lib/expander.js
- lib/led/led.js (done)
- lib/led/rgb.js
- lib/motor.js
- lib/servo.js (done)

dtex:
Oooh, that's better. Let me see if I can grok what's left to be done.

rwaldron:
Basically repeat the process I started on ESC and Servo, but for the rest. 

- In motor.js, the body of `setPWM`...
- In led/led.js, the body of `write`...
- In led/rgb.js, the body of `write`...

...Should be something like: 

```js
var state = priv.get(this);
state.expander.analogWrite(pin, ...);
```

And of course, the `initialization` bodies become something like this: https://github.com/rwaldron/johnny-five/commit/7fb4409afef5f12c311d668c4f7de421a23dbba2#diff-2ea177c952d168875093e55f1f6c6991R16, but with appropriate frequency and pwmRange. We still have the issue that only one frequency and range can be used per PCA9685 chip, but that's ok and I don't want to try and change that with this patch set.

dtex:
Okay to just add commits onto your branch and squash later?

rwaldron:
@dtex that works for me, or we can keep the whole history. Either is fine

dtex:
These changes are in PR #947

PR #928, opened 2015-10-16
Johnny-five pull requested

ziegfrid:


rwaldron:
Thanks, but it looks like these have already landed ;)

PR #930, opened 2015-10-18
Add support for metaserialport

aredridel:


rwaldron:
Thanks for the PR but there is already a work in progress that the johnny-five team has started testing in firmata.js https://github.com/jgautier/firmata Older versions of node and node-serialport need continued support for the foreseeable future.

rwaldron:
Resolved in v0.9.1

aredridel:
Awesome!

rwaldron:
Not sure if you saw the commit message, but I need to be sure that it's clear: I made a mistake when originally approaching this issue because I had misunderstood NAN 2's backward compatibility breakage.  My understanding now is that modules which depended on NAN 2 simply had to update to the new API for Node.js 3.x+ (that was the BC breakage), but once updated, the consumers of _those_ modules were fine all the way back 0.8.x. This latter part was my misunderstanding—I thought that NAN 2 _only_ worked with Node.js 3+ and that older versions of Node would need an older NAN version—this was **incorrect**. So there you have it and Johnny-Five is good to go

aredridel:
Yeah! That was my understanding as well, but I am happy to know that's not the case!

PR #940, opened 2015-10-24
added err parameter to function on example code

lucasjmoore:
Eg/nunchuk.js example wouldn't run with out throwing "Cannot read property 'axis' of null". This is fix by adding err to the function parameters before passing in the event.

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,7 +280,7 @@ To interactively navigate the examples, visit the [Johnny-Five examples](http://
 - [Shift Register - Seven segment controller](https://github.com/rwaldron/johnny-five/blob/master/docs/shift-register-seven-segment.md)
 - [Shift Register - Seven segments, daisy chained](https://github.com/rwaldron/johnny-five/blob/master/docs/shift-register-daisy-chain.md)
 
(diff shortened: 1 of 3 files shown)


rwaldron:
@lucasjmoore thanks for the contribution! In 0.9.x we removed all of those weird error params from event handler callback parameter lists.

frxnz:
@rwaldron It looks like a [couple of these were missed](https://github.com/rwaldron/johnny-five/blob/master/lib/wii.js#L274-L316) in the `Wii` class.

@lucasjmoore As @rwaldron mentioned the `err` params were removed from emitted events. The proper fix would be to remove the middle `null` param everywhere you see something like `emit(type, null, event)` in [wii.js](https://github.com/rwaldron/johnny-five/blob/master/lib/wii.js). If you want to submit a new PR for that, that would be awesome!

The tests should have caught this, so some updated tests would be amazing :smile:

rwaldron:
Weird, thanks for checking up on that.

lucasjmoore:
Ohhh I see, that explains that, thanks for the added explanation.

PR #943, opened 2015-10-28
If a controller is specified for a component class using an Expander, don't normalize or "request" pins

rwaldron:
cc @dtex 

This will pull into `controllers-can-use-expanders`


scottgonzalez:
Can `Board.Component` just do that check internally and not require each component to repeat the logic?

rwaldron:
> Can Board.Component just do that check internally and not require each component to repeat the logic?

Well, yes... I wanted it that way to begin with and wrote it as such, but... putting this in `board.js` requires adding: 

```js
var Expander = require("../lib/expander");
```

And inside `expander.js` there is: 

```js
var Board = require("../lib/board");
```

And some time later...

```js
var methods = Object.keys(Board.prototype);
```

Which causes: 

```
TypeError: Cannot convert undefined or null to object
```

Let me try to get around that.

lib/board.js
```diff
@@ -818,6 +819,17 @@ Board.Component = function(opts, componentOpts) {
     }
   }
 
+  if (!Expander) {
```
negherbon: with instruction inside, is better remove keys or is the same thing?

dtex:
When initializing a component on an expander we still need to run the expander's pin normalization on the component's opts.pin/opts.pins. I don't see that here but maybe I'm just missing it?

dtex:
Did some testing. Looks pretty good but doesn't warn about pin reuse on controllers. This should warn but does not:

````
  var servo2 = new five.Servo({
    controller: "PCA9685",
    address: 0x73,
    pin: 2
  });

  var servo3 = new five.Servo({
    controller: "PCA9685",
    address: 0x73,
    pin: 2
  });
````

rwaldron:
>  but doesn't warn about pin reuse on controllers. This should warn but does not:

I don't know what to do about this right now. TBH, pin tracking has been nothing but a nightmare.

PR #947, opened 2015-10-31
Controllers can use expanders

dtex:
* Switch PCA9685 Controllers to use Expanders
* Add support for Tessel 2 Servo Module
* Fix default PWM range for PCA9685


rwaldron:
As always, it's a pleasure working with you :)

rwaldron:
Can you do a news write up for johnny-five.io?

dtex:
@rwaldron Sure. The lead is the Expander class with a mention of support for T2+Servo Module right?

Issue #956, opened 2015-11-10
Implement HC-SR04-I2C-BACKPACK controller


ajfisher:
Implement a controller which works for the HC-SR04 module that works with the backpack at https://github.com/ajfisher/nodebots-hcsr04

@ajfisher to implement - this here as a note to finish it off.

ajfisher:
Just got to fix a test - but this is failing on the within test.

```javascript
exports["Proximity: HCSR04I2C"] = {
  setUp: function(done) {
    this.board = newBoard();
    this.clock = sinon.useFakeTimers();
    this.i2cConfig = sinon.spy(MockFirmata.prototype, "i2cConfig");
    this.i2cReadOnce = sinon.stub(MockFirmata.prototype, "i2cReadOnce", function(ADDRESS, BYTES_TO_READ, callback) {
      var pulseval = 1000;
      // Split to HIGH and LOW
      console.log(pulseval);
      callback(pulseval);
    });

    this.proximity = new Proximity({
      controller: "HCSR04I2C",
      freq: 100,
    });

    done();
  },

  tearDown: function(done) {
    this.i2cConfig.restore();
    this.i2cReadOnce.restore();
    this.clock.restore();
    done();
  },

  shape: function(test) {
    test.expect(proto.length + instance.length);

    proto.forEach(function(method) {
      test.equal(typeof this.proximity[method.name], "function");
    }, this);

    instance.forEach(function(property) {
      test.notEqual(typeof this.proximity[property.name], 0);
    }, this);

    test.done();
  },

  data: function(test) {
    var spy = sinon.spy();
    test.expect(1);

    this.proximity.on("data", spy);
    this.clock.tick(100);
    test.equal(spy.callCount, 1);
    test.done();
  },

  change: function(test) {
    test.expect(1);

    var spy = sinon.spy();

    this.proximity.on("change", spy);

    this.clock.tick(100);

    test.ok(spy.called);
    test.done();
  },

    // TODO FIX THIS
  within: function(test) {
    var spy = sinon.spy();
    test.expect(2);

    this.clock.tick(250);

    this.proximity.within([0, 120], "inches", function() {
      test.equal(this.inches, 5.85);
      spy();
    });

    this.clock.tick(100);
    test.ok(spy.calledOnce);
    test.done();
  }
};
```

Error:

```bash
>> Proximity: HCSR04I2C - within
>> Error: NaN == 5.85
>> at Proximity.<anonymous> (test/proximity.js:1124:12)
[...]

PR #961, opened 2015-11-11
HCSR04 I2C Backpack Support

ajfisher:
Support for a new proximity controller for the HCSR04 I2C backpack including docs, circuit diagrams and tests.

Addresses #956


docs/proximity-hcsr04-i2c.md
```diff
@@ -0,0 +1,67 @@
+<!--remove-start-->
+
+# Proximity - HC-SR04
+
+<!--remove-end-->
+
+
+Ping proximity example for use with an `HCSR04` sensor and I2C backpack. Standard
+firmata should be used or any other standard I2C capable IO Plugin.
```
frxnz: Should we add a note here and a link to https://github.com/ajfisher/nodebots-hcsr04?

frxnz:
You'll also need to add an entry to [programs.json](https://github.com/rwaldron/johnny-five/blob/master/tpl/programs.json) so the docs will be added to johnny-five.io

This is awesome! The first official Johnny-Five supported backpack :smile:

test/proximity.js
```diff
@@ -1049,6 +1049,86 @@ exports["Proximity: HCSR04"] = {
...
 };
 
+exports["Proximity: HCSR04I2C"] = {
+  setUp: function(done) {
+    this.board = newBoard();
+    this.clock = sinon.useFakeTimers();
+    this.i2cConfig = sinon.spy(MockFirmata.prototype, "i2cConfig");
+    this.i2cReadOnce = sinon.stub(MockFirmata.prototype, "i2cReadOnce", function(ADDRESS, BYTES_TO_READ, callback) {
+      // Use this for 1000 us as duration (HIGH and LOW)
+      callback([3, 232]);
```
dtex: The way this is handled in LidarLite makes it more clear what's happening. 
````
var val = 1000;
callback([ val >> 8, val & 0xff ])
````
Sorry for my bad suggestion earlier.

File: docs/proximity-hcsr04-i2c.md [hunk 3]
```diff
@@ -0,0 +1,67 @@
...
+```
+
+
+```javascript
+var five = require("johnny-five");
+var board = new five.Board();
+
+board.on("ready", function() {
+  var proximity = new five.Proximity({
+    controller: "HCSR04I2C",
```
dtex: Should we indicate that this is a back pack solution in the controller name?

lib/proximity.js
```diff
@@ -190,6 +190,36 @@ var Controllers = {
...
+					this.board.io.i2cReadOnce(address, BYTES_TO_READ, function(data) {
+						datahandler((data[0] << 8) + data[1]);
+                        read();
+					});
+				}.bind(this), msToNextRead);
+			}.bind(this);
+
+            read();
+        }
+    },
```
dtex: Nit: The indentation is inconsistent. Two spaces per tab is what most of the library has. 

@rwaldron I think jshint used to catch this. Can/should we add this to .jscsrc or is this a non-issue?

docs/proximity-hcsr04-i2c.md
frxnz: Now that you mention it I think the original plan was to call this `HCSR04BACKPACK`

dtex:
> This is awesome! The first official Johnny-Five supported backpack

Second that. Can't wait to see all the new classes/controllers and ideas.

ajfisher:
Answers to those various points.

The jshint stuff fixed it but modded 100+ files so probably needs to be
fixed to run automatically. I'll fix these manually for this pr then do
another one with all of the auto fixes to get it back up to date.

Re name, I kind of got sick of typing it. ;P happy to change it though if
it makes it more clear.

And will make those changes to the way the value is being created in the
test.

Will try to get those done this morning.

lib/proximity.js
```diff
@@ -190,6 +190,36 @@ var Controllers = {
...
+					  this.board.io.i2cReadOnce(address, BYTES_TO_READ, function(data) {
+						  datahandler((data[0] << 8) + data[1]);
+              read();
+					  });
+				  }.bind(this), msToNextRead);
+			  }.bind(this);
+
+        read();
+      }
+    },
```
rwaldron: Something funky with the indents here? Perhaps you're using hard tabs?

lib/proximity.js
```diff
@@ -190,6 +190,36 @@ var Controllers = {
...
+        var msToNextRead = 90;
+
+        opts.address = address;
+
+        // set up IO connection:
+        this.io.i2cConfig(opts);
+
+			  var read = function() {
+				  setTimeout(function() {
+					  this.board.io.i2cReadOnce(address, BYTES_TO_READ, function(data) {
```
rwaldron: Why `i2cReadOnce`? `i2cRead` is always more efficient—can that be done? Presumably the delays between actual reading on the backpack can behandled there. Even if the continuous read gets whatever is in the register twice in a row?

tpl/programs.json
```diff
@@ -629,6 +629,11 @@
         "description": "Basic ping Proximity example for any of the [`PING_PULSE_IN`](http://johnny-five.io/api/proximity/#controller-alias-table) components (eg. `HCSR04` or Parallax's Ultrasonic Ping). [PingFirmata](http://johnny-five.io/api/proximity/#pingfirmata) must be loaded onto Arduino-compatible boards to enable this component."
       },
       {
+        "file": "proximity-hcsr04-i2c.js",
+        "title": "Proximity - HC-SR04 I2C Backpack",
+        "description": "Ping Proximity example that uses an HCSR04 ultrasonic sensor and an I2C backpack ([see backpack documentation](https://github.com/ajfisher/nodebots-hcsr04).
```
rwaldron: This line is missing a double quote at the end. I think that will fix the build

lib/proximity.js
```diff
@@ -190,6 +190,36 @@ var Controllers = {
...
+					  });
+				  }.bind(this), msToNextRead);
+			  }.bind(this);
+
+        read();
+      }
+    },
+    toCm: {
+      value: function(raw) {
+        return +(raw / 29.1 / 2).toFixed(3);
```
rwaldron: If you have full control over the value that the backpack writes to its register, why not just write a cm value? Then there is no magic numbers on this end.

rwaldron:
A couple notes, but overall this is totally awesome!

lib/proximity.js
ajfisher: Mostly because it can do sub cm detection so passing a float back is a bit grotty. In theory also you are supposed to account for air pressure when you calculate this so a raw value would allow you to accommodate for that (not sure anyone other than me has ever *actually* done that with a nodebot yet though ;P )

tpl/programs.json
ajfisher: Yep - just discovered `grunt test-examples`

lib/proximity.js
ajfisher: Not sure what vim was doing here - trying to be too smart I think...

lib/proximity.js
rwaldron: Ah! Thanks for explanation

lib/proximity.js
```diff
@@ -190,6 +190,36 @@ var Controllers = {
...
+        var BYTES_TO_READ = 0x02;
+        var address = opts.address || 0x27;
+        var msToNextRead = 90;
+
+        opts.address = address;
+
+        // set up IO connection:
+        this.io.i2cConfig(opts);
+
+			  var read = function() {
```
rwaldron: This is funky, but I can fix it

PR #962, opened 2015-11-12
Update extended test to use external mock-firmata

dtex:


rwaldron:
Whoops, sorry about that!

PR #965, opened 2015-11-15
Increase PWM frequency for PCA9685

blaines:
This will reduce visible PWM flashing with the PCA9685. Mostly noticeable when recording video but also visible with the eye under certain circumstances.

### DIFF
--- a/lib/led/led.js
+++ b/lib/led/led.js
@@ -21,7 +21,7 @@ var Controllers = {
           controller: this.controller,
           bus: this.bus,
           pwmRange: this.pwmRange,
-          frequency: 50, // Hz
+          frequency: 100, // Hz
         });
 
         this.pin = typeof opts.pin === "undefined" ? 0 : opts.pin;


blaines:
It'd be nice to have this as an option for five.Led. Willing to implement this with some guidance.

lib/led/led.js
```diff
@@ -21,7 +21,7 @@ var Controllers = {
           controller: this.controller,
           bus: this.bus,
           pwmRange: this.pwmRange,
-          frequency: 50, // Hz
+          frequency: 100, // Hz
```
rwaldron: This should actually be 200Hz, can you try that and confirm?

rwaldron:
> It'd be nice to have this as an option for five.Led. Willing to implement this with some guidance.

Just change 

```diff
-          frequency: 50, // Hz
+          frequency: this.frequency || 200, // Hz
```

And then this should work: 

```js
new five.Led({ 
  controller: "PCA9685", 
  pin: 0, 
  frequency: ..., 
  // and so on
});
```

rwaldron:
Also here https://github.com/rwaldron/johnny-five/blob/master/lib/led/rgb.js#L45-L60

PR #975, opened 2015-12-03
Stop continuous rotation servo on instantiation

dtex:
I can think of no reason why we would not want this to happen. Servo state is a crap shoot otherwise.

### DIFF
--- a/lib/servo.js
+++ b/lib/servo.js
@@ -202,6 +202,11 @@ function Servo(opts) {
   if (opts.center) {
     this.center();
   }
+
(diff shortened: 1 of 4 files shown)


lib/servo.js
```diff
@@ -202,6 +202,11 @@ function Servo(opts) {
   if (opts.center) {
     this.center();
   }
+
+  if (opts.type === "continuous") {
+    this.stop();
+  }
```
rwaldron: :+1:

PR #980, opened 2015-12-10
changing this.value to this.scaled to reflect light sensitivity

lrlna:
Hi! 

`this.value` outputs a really small value which doesn't reflect on the led :|. 
`this.scaled` seems to work better with this example :star2: :tada:


nodebotanist:
++

rwaldron:
Awesome, thanks!

lrlna:
Woo! :sparkles:

PR #982, opened 2015-12-11
Add temporalFallback

dtex:
temporal uses 100% of the CPU so on long running animations we switch to setInterval instead of temporal (after five seconds). 

Does five seconds seem good? It's completely arbitrary.

Tests look good but I still want to try with hardware so don't land this yet.

### DIFF
--- a/lib/animation.js
+++ b/lib/animation.js
@@ -37,6 +37,20 @@ Animation.normalize = "@@normalize";
 Animation.render = "@@render";
 
 /**
+ * Temporal will run up the CPU. temporalFallback is used
+ * for long running animations.
+ */
+function TemporalFallback(animation) {
+  this.interval = setInterval(function() { animation.loopFunction({calledAt: Date.now()}); }, animation.rate);
+}
+
+TemporalFallback.prototype.stop = function() {
+  if (this.interval) {
+    clearInterval(this.interval);
+  }
+};
+
+/**
  * Animation
  * @constructor
  *
@@ -126,9 +140,7 @@ Animation.prototype.next = function() {
       this.onstart();
     }
 
-    this.normalizedKeyFrames = __.cloneDeep(this.keyFrames);
-    this.normalizedKeyFrames = this.target[Animation.normalize](this.normalizedKeyFrames);
-    this.normalizedKeyFrames = this.normalizeKeyframes(this.normalizedKeyFrames, this.cuePoints);
+    this.normalizeKeyframes();
 
     if (this.reverse) {
       this.currentSpeed *= -1;
@@ -156,7 +168,9 @@ Animation.prototype.pause = function() {
 
   this.emit("animation:pause");
 
-  this.playLoop.stop();
+  if (this.playLoop) {
+    this.playLoop.stop();
+  }
   this.paused = true;
 
   if (this.onpause) {
@@ -177,6 +191,9 @@ Animation.prototype.stop = function() {
   this.emit("animation:stop");
 
   this.segments = [];
+  if (this.playLoop) {
+    this.playLoop.stop();
+  }
   temporal.clear();
 
   if (this.onstop) {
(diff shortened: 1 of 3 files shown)


lib/animation.js
```diff
@@ -37,6 +37,19 @@ Animation.normalize = "@@normalize";
 Animation.render = "@@render";
 
 /**
+ * Temporal will run up the CPU. temporalFallback is used
+ * for long running animations.
+ */
+var temporalFallback = function(animation) {
+  this.interval = setInterval(function() { animation.loopFunction({calledAt: Date.now()}); }, animation.rate);
+  return this;
```
rwaldron: `return this;` is implicit

lib/animation.js
```diff
@@ -37,6 +37,19 @@ Animation.normalize = "@@normalize";
 Animation.render = "@@render";
 
 /**
+ * Temporal will run up the CPU. temporalFallback is used
+ * for long running animations.
+ */
+var temporalFallback = function(animation) {
```
rwaldron:
```js   
function TemporalFallback(...) {
```

lib/animation.js
```diff
@@ -37,6 +37,19 @@ Animation.normalize = "@@normalize";
...
+  this.interval = setInterval(function() { animation.loopFunction({calledAt: Date.now()}); }, animation.rate);
+  return this;
+};
+
+temporalFallback.prototype.stop = function() {
+  clearInterval(this.interval);
```
rwaldron: Check `if (this.interval) {...` first

lib/animation.js
```diff
@@ -210,6 +226,65 @@ Animation.prototype.speed = function(speed) {
...
+  // Get tweened value
+  var val = this.tweenedValue(indices, progress);
+
+  // call render function
+  this.target[Animation.render](val);
+
+  // If this animation has been running for more than 5 seconds
+  if (loop.calledAt > this.fallBackTime) {
+    this.fallBackTime = Infinity;
+    this.playLoop.stop();
```
rwaldron: There are other places where `this.playLoop` is checked before stop is called, should it be same here, or is this known to be safe?

lib/animation.js
```diff
@@ -210,6 +226,65 @@ Animation.prototype.speed = function(speed) {
...
+        this.onloop();
+      }
+
+      if (this.metronomic) {
+        this.reverse = this.reverse ? false : true;
+      }
+
+      this.normalizedKeyFrames = __.cloneDeep(this.keyFrames);
+      this.normalizedKeyFrames = this.target[Animation.normalize](this.normalizedKeyFrames);
+      this.normalizedKeyFrames = this.normalizeKeyframes();
```
rwaldron: While we're here, these three lines have always had a smell, maybe we can rethink this? I'd prefer  if `this.normalizeKeyFrames()` is modified to do the normalization and update `this.normalizedKeyFrames` directly, then just `return this;`. That way any animation object can have its `this.normalizeKeyFrames()` called at any time and the result is the animation object's `normalizedKeyFrames` property is normalized.

lib/animation.js
dtex: Ah, because I used the "new" keyword. I love learning new things.

lib/animation.js
dtex: playLoop invokes loopFunction so it's know to be running here but this caused me to check the other instances. The one inside pause() was at risk so I added the check there.

lib/animation.js
dtex: I had it in the back of my mind that users might want to use animation's normalize routine w/o the device's normalize routine though now that time has passed I can't imagine why.

I'll rework it.

rwaldron: Cool, ping me when ready :)

dtex:
Okay, I've made those changes. I accidentally deleted three lines without noticing and that cost me half a friggin' day but I'm a dingus, programming is hard and computers are stupid because they let me do stupid things.

We should be good to go now. The processor spins up to >120% according to Activity Monitor (not sure how that's possible), and then once the temporalFallback kicks in it drops back down to 15% or so (even on my hexapod).

rwaldron:
Sounds like my day!

rwaldron:
This is really great, nice work :)

Land at will

PR #990, opened 2015-12-16
Add Servo.home()

dtex:
Returns the servo to its startAt position. 

For a single servo it's just sugar for ```servo.to(servo.startAt)```

For a servo array it will return each servo to its own startAt value (handy for setting chains to their home positions).


rwaldron:
LGTM, land at will!

PR #993, opened 2015-12-16
Add two controller servos

dtex:
Sorry for the two PR's. Decided to add the second diagram after seeing the first in place.


rwaldron:
These are great!

PR #994, opened 2015-12-17
Wiki artwork additions and ESCs example

dtex:


dtex:
@rwaldron I've never actually worked with an ESC so would you double check my example here and the [ESCs wiki page](https://github.com/rwaldron/johnny-five/wiki/ESCs) The images will show up after this lands.

rwaldron:
Yep, these are correct—thanks!

PR #998, opened 2015-12-18
Motor enable pin

dtex:


lib/motor.js
```diff
@@ -597,7 +599,7 @@ util.inherits(Motor, events.EventEmitter);
 Motor.prototype.initialize = function() {
   this.io.pinMode(this.pins.pwm, this.io.MODES.PWM);
 
-  ["dir", "cdir", "brake"].forEach(function(pin) {
+  ["dir", "cdir", "brake", "enable"].forEach(function(pin) {
     if (this.pins[pin]) {
```
dtex: Are there any boards where dir, cdir, break or enable could be on pin 0?

rwaldron: Yeah, actually now that you mention it: Tessel 2.

I'd change this to `if (typeof this.pins[pin] !== "undefined") {`

lib/motor.js
```diff
@@ -727,12 +729,28 @@ Motor.prototype.release = function() {
...
-  this.dir(speed, this.direction);
+  this.dir(this.direction);
   this.start(speed);
 
   return this;
 };
 
+Motor.prototype.enable = function() {
+  var state = priv.get(this);
+  if (this.pins.enable && !this.enabled) {
```
rwaldron: I guess this would need to change as well: `if (typeof this.pins.enable !== "undefined"...`

lib/motor.js
```diff
@@ -727,12 +729,28 @@ Motor.prototype.release = function() {
...
+    this.setPin(this.pins.enable, 1);
+    state.enabled = true;
+  }
+};
+
+Motor.prototype.disable = function() {
+  var state = priv.get(this);
+  if (this.pins.enable && this.enabled) {
```
rwaldron: And here

rwaldron:
I'll make an example that shows using the sparkfun block thing.

dtex:
Did you get a chance to test it?

rwaldron:
Not yet, I'm _trying_ to be lazy today, but that's only half working

rwaldron:
@dtex can you update with the changes noted? Thanks!

dtex:
Made that change, but note that I have no way to test in the real world since I don't have that block so hopefully someone can test it out for me.

rwaldron:
I added an example that illustrates the test hardware that I used—thanks again @dtex 

cc @Legendaire

Legendaire:
I fried my motor block so I can't test until the new one arrives.

dtex:
Yikes, how'd that happen?

Legendaire:
Who knew the plus sign meant positive and not the other way around......

rwaldron:
Happens to the best

PR #1001, opened 2016-01-07
adafruit_MPL3115A2

kimio-kosaka:
Fixed connection error  Arduino UNO ---- Adafruit-MPL3115A2.
1) fixed I2C connection error.
2) connected 5v directly.
Adafruit-MPL3115A2 include voltage regulator(5v to 3v3) and I2C logic level converter. It can connect to 5V circuit directly.


rwaldron:
Thank you @kimio-kosaka!

rwaldron:
Updated: http://johnny-five.io/examples/multi-mpl3115a2/

Issue #1000, opened 2016-01-07
Support for HT1632C


ericandrewlewis:
This was [sorta asked previously](https://github.com/rwaldron/johnny-five/issues/577), but the conversation ended up about the `HT16K33` rather than the `HT1632C`.

Can we add support for `HT1632C` to Johnny Five? This drives a [16x24 Adafruit matrix](https://www.adafruit.com/product/555) I just got.

dtex:
Hi @ericandrewlewis ,

Johnny-Five contributors and maintainers are loathe to close issues where someone has a need. We don't want anyone to ever feel that we don't care, but we just haven't gotten to this feature yet. Rather than leave it languishing as an open issue we have created a [Requested Features](../wiki/Requested-Features) page and added your request for the HT1632C there.

Issue #665, opened 2015-03-23
refactor LedControl


boneskull:
After trying to tweak `LedControl.js` to work with a [24-segment, bi-color bar-graph](https://www.adafruit.com/products/1721), I've come to the conclusion that the code needs some refactoring, because it scales poorly.
- if you are attempting to use the `HT16K33` controller in the context of a bar graph (since that's the chip the bar graph backpack uses), certain functions should not be available, and others (or rather, one, `setBar()`) should be.  
- it is difficult to reuse code in a controller given the common pattern of defining controllers; no "subcontrollers" are allowed
- Adding new functions requires explicitly adding them to the context of the `LedControl` instance (which _could_ be done via something like `_.mixin()` or `_.assign()`)
- controllers are not sufficiently encapsulated as to offer unique implementations for a display using a unique IC; the HT16K33 chip can drive all sorts of displays, not just traditional matrices or arrays of matrices.
- Matrix dimension restrictions should be shelved except in the case of hardware-specific controllers.  I don't need the code to tell me that driving six LEDs with a 74595 is not allowed!
- I don't understand what a `DEFAULT` controller buys us, if `LedControl.prototype` is non-empty.

To solve this, I would expect to use either inheritance or composition to extend the `LedControl` object with the proper functions.

Three (3) solutions follow.

Using inheritance:
- The `DEFAULT` controller's general-purpose methods are folded into `LedControl.prototype`, if any.  For all intents & purposes, `LedControl` is abstract.
- The `HT16K33` pseudoclass would be a static property of `LedControl`, and extend `LedControl`, providing specific implementation for this chip.   It contains no code which assumes any given matrix configuration.
- `Matrix`, `Digit` (or `SevenSegment`), `BarGraph` and even `Alphanumeric` would be static, class-level psuedoclasses in `HT16K33`, containing code specific to each implementation.  The `Adafruit_LEDBackpack` lib has implementation for all of these.
- The `ShiftRegister74595` (for lack of a better name) pseudoclass would do likewise.
- The organization scheme here is simply for sake of organization and is not required.
- Use the factory pattern to grab the class you want:
  
  ``` js
  LedControl.create = function create(type, controller, opts) {
    return new LedControl[controller][type](opts);
  };
  
  // alternatively:
   LedControl.createMatrix = function createMatrix(controller, opts) {
     return new LedControl[controller].Matrix(opts);
  };
  
  LedControl.createBarGraph = function createBarGraph(controller, opts) {
[...]

PR #1006, opened 2016-01-10
[WIP] support for HT1632C LED control

ericandrewlewis:
This is a work in progress PR for #1000. More or less a JS port of [Adafruit's HT1632C library](https://github.com/adafruit/HT1632). 

Feedback welcome :smile: More soon.

### DIFF
--- a/lib/led/ledcontrol.js
+++ b/lib/led/ledcontrol.js
@@ -72,6 +72,16 @@ function LedControl(opts) {
   this.sendDigit = controller.sendDigit;
   this.initialize = controller.initialize;
 
+  // Functions from the HT1632C controller. Perhaps we can find a more declarative
+  // way to copy functions from controller to LEDcontrol (mixin?)
+  this.writeData = controller.writeData;
+  if (controller.draw) {
+    this.draw = controller.draw;
+  }
+
+  if (controller.brightness) {
+    this.brightness = controller.brightness;
+  }
   // controller specific op codes
   this.OP = controller.OP;
 
@@ -86,6 +96,10 @@ function LedControl(opts) {
   // extra functions for HT16K33 devices only
   if (controller.writeDisplay) {
     this.writeDisplay = controller.writeDisplay;
+  }
+
+  if (controller.clearBuffers) {
+    this.clearBuffers = controller.clearBuffers;
   }
   if (controller.blink) {
     this.blink = controller.blink;
@@ -122,8 +136,10 @@ function LedControl(opts) {
     columns: opts.dims.columns
   };
 
-  if (!(state.columns === 8 || state.columns === 16) || !(state.rows === 8 || state.rows === 16) || (state.columns + state.rows === 32)) {
-    throw new Error("Invalid matrix dimensions specified: must be 8x8, 16x8 or 8x16");
+  if ( opts.controller !== "HT1632C" ) {
+    if (!(state.columns === 8 || state.columns === 16) || !(state.rows === 8 || state.rows === 16) || (state.columns + state.rows === 32)) {
+      throw new Error("Invalid matrix dimensions specified: must be 8x8, 16x8 or 8x16");
+    }
   }
 
   Object.defineProperties(this, {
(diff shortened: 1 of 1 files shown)


lib/led/ledcontrol.js
```diff
@@ -704,6 +711,128 @@ Controllers = {
...
+      for ( var i = numberOfBits; i > 0; i-- ) {
+        this.io.digitalWrite(this.pins["write"], 0);
+        if ( data & ( 1 << (i-1) ) ) {
+          this.io.digitalWrite(this.pins["data"], 1);
+        } else {
+          this.io.digitalWrite(this.pins["data"], 0);
+        }
+        this.io.digitalWrite(this.pins["write"], 1);
+      }
+      this.io.pinMode(this.pins["data"], this.io.MODES.INPUT);
```
rwaldron: For all occurrences, just use dot access: `this.pins["write"]` => `this.pins.write`

lib/led/ledcontrol.js
```diff
@@ -704,6 +711,128 @@ Controllers = {
...
+     * @return {[type]} [description]
+     */
+    writeScreen: function() {
+      this.io.digitalWrite(this.pins["cs"], 0);
+      this.writeData(0x5, 3);
+
+      this.writeData(0x0, 7);
+
+      var width = 24, height = 16;
+      for (i=0; i <= (width * height / 8); i+=2) {
```
rwaldron: `i` is leaking here, needs `var i`

lib/led/ledcontrol.js
```diff
@@ -72,6 +72,13 @@ function LedControl(opts) {
   this.sendDigit = controller.sendDigit;
   this.initialize = controller.initialize;
 
+  // Functions from the HT1632C controller. Perhaps we can find a more declarative
+  // way to copy functions from controller to LEDcontrol (mixin?)
```
rwaldron: Yes, `LedControl` is a nightmare and I want to toss it for a complete overhaul that follows the Component/Controller model (as seen in just about every other class). Unfortunately, this is a huge undertaking and I don't have the time to do it and no one has stepped forward to tackle it.

lib/led/ledcontrol.js
```diff
@@ -72,6 +72,13 @@ function LedControl(opts) {
...
+  this.sendCommand = controller.sendCommand;
+  this.writeScreen = controller.writeScreen;
+  this.writeData = controller.writeData;
+  this.clearScreen = controller.clearScreen;
```
rwaldron: This should be defined as `clear`

lib/led/ledcontrol.js
```diff
@@ -704,6 +711,128 @@ Controllers = {
...
+      this.sendCommand(this.OP.LED_ON);
+      this.sendCommand(this.OP.BLINK_OFF);
+      this.sendCommand(this.OP.MASTER_MODE);
+      this.sendCommand(this.OP.INT_RC);
+      // hardcoding this for now. need to figure out a nice way of configuring this.
+      this.sendCommand(this.OP.COMMON_16NMOS);
+      this.sendCommand(this.OP.PWM_CONTROL | 0xF);
+    },
+
+    clearScreen: function() {
```
rwaldron: `clearScreen` => `clear`

lib/led/ledcontrol.js
```diff
@@ -704,6 +711,128 @@ Controllers = {
...
+      this.pins = {
+        data: opts.pins.data,
+        write: opts.pins.write,
+        cs: opts.pins.cs
+      };
+      ["data", "write", "cs"].forEach(function(pin) {
+        this.io.pinMode(this.pins[pin], this.io.MODES.OUTPUT);
+      }, this);
+
+      this.ledMatrix = [];
```
rwaldron: This looks like it's serving the same purpose as `this.displaybuffers`?

lib/led/ledcontrol.js
```diff
@@ -704,6 +711,128 @@ Controllers = {
...
+        this.ledMatrix[i] = Math.floor( Math.random() * 256 );
+      }
+      this.writeScreen();
+    },
+
+    /**
+     * Write the data in the buffers to the screen.
+     * @return {[type]} [description]
+     */
+    writeScreen: function() {
```
rwaldron: This looks like `writeDisplay`. https://github.com/ericandrewlewis/johnny-five/blob/issue-1000/lib/led/ledcontrol.js#L622-L630

lib/led/ledcontrol.js
```diff
@@ -704,6 +711,128 @@ Controllers = {
...
+
+      this.io.digitalWrite(this.pins["cs"], 1);
+    },
+
+    /**
+     * Send a command to the controller.
+     * @param  {Number} commandCode The command code.
+     * @return {[type]}             [description]
+     */
+    sendCommand: function(commandCode) {
```
rwaldron: `send`

ericandrewlewis:
Thanks for the feedback @rwaldron, I've fixed up accordingly.

Most API functionality has parity with the other matrices.

Here's some example code that would work with an `HT1632C`, hooked up the same way as in [this tutorial](https://learn.adafruit.com/16x24-led-matrix/wiring):

```javascript
var five = require("johnny-five");

var board = new five.Board();

  var matrix = new five.Led.Matrix({
    controller: "HT1632C",
    pins: {
      data: 2,
      write: 3,
      cs: 4
    },
    devices: 1,
    dims: {
      rows: 16,
      columns: 24
    }
  });

  matrix.led( 0, 0, true );
  matrix.led( 15, 23, true );
  matrix.writeDisplay();
  matrix.blink();
  matrix.brightness(0);
  // Maximum brightness.
  matrix.brightness(15);
  

  matrix.draw( "hi" );

  matrix.brightness(0);
  matrix.writeDisplay();
});
```

Some open questions:

Writing to the display seems expensive at the moment, so I've disabled `led()` from calling it directly at the moment. Might have to take a look at the C++ library again for optimizations if any are possible. For now, `writeDisplay()` must be called explicitly after `led()`. 

How would one expect `draw()` to work with this display? Passing in an array of row data makes sense. But what about passing in characters, since this display has space for 6 characters?

The HT16K33 supports bicolor displays. Are there bicolor HT1632C matrices?

The HT16K33 supports rotation. Should the HT1632C do this as well?

Any other feedback appreciated.

rwaldron:
> For now, writeDisplay() must be called explicitly after led().

This is a semantic deviation that definitely needs to be resolved, but I have a feeling that won't be possible considering...

> Writing to the display seems expensive at the moment,

Based on the [communication protocol requirements](https://github.com/adafruit/HT1632/blob/master/HT1632.cpp#L402-L441), I assumed it would be. It's not much different from the MAX7219 and MAX722 devices, which are written using `shiftOut` (which is also expensive, though not quite as much)

> How would one expect draw() to work with this display? Passing in an array of row data makes sense. But what about passing in characters, since this display has space for 6 characters?

I honestly have no idea. This display breaks a lot of assumptions made by existing LED matrix component controller implementations—that Johnny-Five just followed because that seemed to be a smart thing to do. Part of me wishes Johnny-Five could offload some of this stuff into plugins like the totally excellent [node-pixel](https://github.com/ajfisher/node-pixel), but that ship has probably sailed :\

Let's gather some other brains that have worked on plugin stuff and/or led component stuff...

@ajfisher @frxnz @boneskull 


The thing I'd like to think about and discuss: building a set of LED component (digit, matrix, bar graph) classes as a plugin similar to node-pixel. I think this would be a way to expand the capabilities in a smart way and most importantly starting _clean_. @dtex and @BrianGenisio are brilliant at identifying capability intersection.

boneskull:
what was the other issue I was talking about this in?  Had something to do with bar graphs...

boneskull:
I think it's worth mentioning that due to the nature of these displays, an inheritance model might be inappropriate. If we define individual capabilities such as "bi-color" or "matrix" and compose resulting objects that way, we won't fall into the trap of trying to classify hardware too hard or violate assumptions. 

I'm not sure a pairing of a public API and a controller object makes much sense though. You can't necessarily use any combination. Maybe the public API should be hardware-specific. I know this goes contrary to much of j5's API but it's worth discussing. The current abstraction for these displays leaks everywhere....

boneskull:
(that ticket was #665)

boneskull:
also, please let me know if the above comment(s) didn't make sense; I wrote them while half-awake

ericandrewlewis:
#665 is a nice read :) 

We could do a refactor there. After that adding support here should be nominal. (Famous last words...) 

I'm going to leave this open for now and see if I can help out on that ticket.

dtex:
Hi @ericandrewlewis ,

Since this has been open and on hold for a while I'm closing. Rather than leave it languishing as an open PR we have created a [Requested Features](../wiki/Requested-Features) page and I have added your request for the HT1632C there with links to this PR and the relevant issue.

PR #1008, opened 2016-01-12
GPS

dtex:
Note: I've edited this first comment so that it reflects the current state of the PR

### My MVPR (Minimum Viable Pull Request) for GPS.

First I should explain why I have the "Breakout", "Receivers" and "Chips" layers... 

GPS breakouts can have three physical "layers": the chip, the antenna/receiver module and the breakout board. Each of these layers are produced by different manufacturers, have their own documentation, and may or may not override and/or block functionality of the other layers by hot wiring pins or not passing them through. They can also add new functionality of their own.

We've just gotta make it all work ¯\_(ツ)_/¯

In GPS() chips are defined in the Chips object (which is analgous to Devices in our Device/Controller model. Breakout boards are in Breakouts and are analogous to Controllers. To handle the middle layer I've added the Receivers object. This object is applied is exactly like the Chips object. Generally, the user should just have to select a breakout board. We can figure out which module (if any) and which chip to use based on that.

These layers aren't fully utilized yet, but I've looked at many breakouts and it will be needed. Sadly I only have the Adafruit Ultimate GPS and the Tessel GPS Module. The latter is waiting on the ability to adjust the baud rate on an open port (coming soon to a firmata near you) so only the former is included in this PR.

````
/**
 * @constructor
 * @param {Object} opts Options: pin(s), device, module, controller, precision, serialport, freq
 *
 * Sample initialization
 *    new five.GPS({ pins: {rx: 10, tx: 11});
 */
````

### Methods

**GPS.sendCommand(String command)**
Sends a command to the gps chip. Line endings and checksum are automatically added.

**GPS.frequency(Number freq)**
Sets the frequency of updates in ms (default is 1000)

**GPS.restart(Boolean coldRestart)**
Restarts the GPS. If coldRestart === true the GPS will completely reset and all user configurations and previous readings will be lost. It can take the GPS 35 or more seconds to start emitting valid data again.

There are a lot of other commands we could add, but I'm not really sure they are needed. I'd rather add them when the need arises.

### Properties

**GPS.pos**
Latitude, longitude and altitude

**GPS.nav**
Course and speed

**GPS.sat**
Satellites used and precision data

### Events

**message**
Emitted when any valid message is received from the GPS.

**operations**
Emitted when the operational details change (i.e. the satellites being used)

**acknowledge**
Emitted when the GPS responds to a command

**unknown**
Emitted when an unknown message is received from the GPS

**data**
Emitted when any message is received, after the data has been updated

**change**
Emitted when lat, long, or alt has changed

**navigation**
Emitted when course, or speed has changed

Like I said, the Tessel GPS Module is coming. I've already got it working, just waiting on one new command in firmata. I know what the command API signature will be and I could go ahead and add it to tessel-io, but I'd feel more comfortable waiting for the firmata update first.

@makenai has a Spark Fun Venus GPS and he is going to look at adding support for that.

Tests are in place.

### CODE REVIEW

**Comment by rwaldron** [comment]
I'll dig into this more tomorrow, but currently "controller" generally corresponds to a chip. Sometimes it gets named for a breakout with a chip where the chip name is less obvious

File: lib/gps.js [hunk 1]
```diff
@@ -0,0 +1,367 @@
...
+ *
+ * Sample initialization
+ *
+ *    new five.GPS({ pins: {rx: 10, tx: 11});
+ *
+ */
+
+function GPS(opts) {
+
+  var controller, module, device, state;
```
rwaldron: Let's not use the word `module`, since it has too much baggage already

package.json
```diff
@@ -123,6 +123,7 @@
   },
   "optionalDependencies": {
     "firmata": "^0.7.2",
+    "firmata": "^0.8.0",
```
rwaldron: Just change the existing version

lib/gps.js
```diff
@@ -0,0 +1,367 @@
...
+}
+
+util.inherits(GPS, events.EventEmitter);
+
+/*
+ * Default intialization for serial GPS
+ */
+GPS.prototype.initialize = function(opts) {
+
+  this.serialPort = opts.serialPort || opts.portId || opts.port || opts.bus || this.io.SERIAL_PORT_IDs.SW_SERIAL0;
```
rwaldron: This belongs in a state object

lib/gps.js
```diff
@@ -0,0 +1,367 @@
...
+
+  var state = priv.get(this);
+  var cksum = sentence.split("*");
+
+  // Check for valid sentence
+  if (cksum[1] !== getNmeaChecksum(cksum[0].substring(1))) {
+    return;
+  }
+
+  process.nextTick(this.emit.bind(this, "message", sentence));
```
rwaldron: `parseNmeaSentence` is already called asynchronously, because `serialRead` is async. These events don't need to be explicitly scheduled for the next turn

lib/gps.js
```diff
@@ -0,0 +1,367 @@
...
+  var hexsum = getNmeaChecksum(inString.substring(1));
+  cc.push(42, hexsum.charCodeAt(0), hexsum.charCodeAt(1), 13, 10);
+
+  this.io.serialWrite(this.serialPort, cc);
+};
+
+GPS.prototype.listen = function() {
+  var input = "";
+  // Start the read loop
+  this.io.serialRead(this.serialPort, 0, function(data) {
```
soundanalogous: you can omit the 2nd param `0` here if you want

package.json
dtex: Fixed.

lib/gps.js
dtex: Done

lib/gps.js
dtex: Done

lib/gps.js
dtex: Done. Thanks for taking a look!

dtex:
Well the important thing is the order in which we decorate the instance with these different layers. The GPS chip must come first, then the antenna module, then the breakout board. We can change the labels all we want and I guess that makes more sense then forcing our existing labels to be something they're not.

How about:

Chips
AntennaModules
Breakouts

?

dtex:
Another question: It looks like I need to add the serialCondig, serialWrite, etc methods to board-io since that's where we get our methods for mock-firmata. Did I follow that trail correctly?

rwaldron:
> Another question: It looks like I need to add the serialConfig, serialWrite, etc methods to board-io since that's where we get our methods for mock-firmata. Did I follow that trail correctly?

Yessir

rwaldron:
> Chips
> AntennaModules
> Breakouts

Why not 

- Chip
- Antenna
- Breakout

?

dtex:
I was just being consistent with the pluralization [for example](https://github.com/rwaldron/johnny-five/blob/master/lib/expander.js#L24) and I guess I was just trying to make the AntennaModule label as generic as I could since it can do more than provide a link to an antenna. Module really is a better label when the scope of concern is limited to GPS's but I understand the confusion it would generate in a JS project.

~googling... googling... googling...~ AHA!

Assuming Pololu used the right taxonomy [here](https://www.pololu.com/product/2138)

Chips
Receivers
Breakouts

dtex:
Submitted a PR to board-io: https://github.com/achingbrain/board-io/pull/9

rwaldron:
Oh sorry, I didn't mean to de-pluralize there...whoops!

dtex:
I added tests for GPS but until achingbrain/board-io#9 lands they will error out.

dtex:
Appveyor seems to be using cached modules and is not grabbing the latest board-io. I submitted a PR to mock-firmata to require the latest board-io: https://github.com/rwaldron/mock-firmata/pull/1

rwaldron:
@dtex can you do a rebase—thank you sir :)

dtex:
Rodger dodger. That other GPS and the I2C adapter came in, I'm just swamped right now. Two of my team are busy zooming around Malaysia on mopeds right now (jerks :-P ).

dtex:
Woah, what happened? This was like three commits a couple of hours ago. Do I need to reset this branch and try again?

lib/gps.js
```diff
@@ -0,0 +1,394 @@
...
+var Pin = require("./pin");
+
+var priv = new Map();
+
+var Breakouts = {
+
+  /*
+   * https://www.adafruit.com/products/746
+   */
+  adafruitUltimateGPS: {
```
rwaldron: `adafruitUltimateGPS` => `ADAFRUIT_ULTIMATE_GPS`

lib/gps.js
```diff
@@ -0,0 +1,394 @@
...
+
+  if (typeof opts.pins === "undefined") {
+    opts.pins = {};
+  }
+
+  Board.Component.call(
+    this, this.opts = Board.Options(opts)
+  );
+
+  breakout = opts.breakout || "adafruitUltimateGPS";
```
rwaldron: ADAFRUIT_ULTIMATE_GPS

lib/gps.js
```diff
@@ -0,0 +1,394 @@
...
+  }
+
+  // If a receiver was defined or derived but chip was not
+  if (receiver && !chip && Receivers[receiver].chip) {
+    chip = Receivers[receiver].chip.value;
+  }
+
+  // Allow users to pass in custom chip types
+  chip = typeof chip === "string" ?
+    Chips[chip] : this.opts.chip;
```
rwaldron: Please don't do this: `this.opts`. `chip` should be a property directly on `this` (and same for all others)

lib/gps.js
```diff
@@ -0,0 +1,394 @@
...
+  if (Array.isArray(opts)) {
+    opts = { pins: {tx: opts[0], rx: opts[1], onOff: opts[2] }};
+  }
+
+  if (typeof opts.pins === "undefined") {
+    opts.pins = {};
+  }
+
+  Board.Component.call(
+    this, this.opts = Board.Options(opts)
```
rwaldron:
```
  Board.Component.call(
  );
```

lib/gps.js
rwaldron: In `Board.Options`: 

```js
Object.assign(this, opts);
```

lib/gps.js
```diff
@@ -0,0 +1,394 @@
...
+
+util.inherits(GPS, events.EventEmitter);
+
+/*
+ * Default intialization for serial GPS
+ */
+GPS.prototype.initialize = function(opts) {
+
+  var state = priv.get(this);
+  state.serialPort = opts.serialPort || opts.portId || opts.port || opts.bus || this.io.SERIAL_PORT_IDs.SW_SERIAL0;
```
rwaldron:
```
this.io.SERIAL_PORT_IDs.SW_SERIAL0;
```

This likely doesn't exist on any other platforms aside from Firmata.js.

dtex: Won't it have to be added into any IO plugin that supports software serial?

lib/gps.js
```diff
@@ -0,0 +1,394 @@
...
+
+  var state = priv.get(this);
+  var cksum = sentence.split("*");
+
+  // Check for valid sentence
+  if (cksum[1] !== getNmeaChecksum(cksum[0].substring(1))) {
+    return;
+  }
+
+  this.emit("message", sentence);
```
rwaldron: "message" => "sentence"?

lib/gps.js
rwaldron: - `this.io.SERIAL_PORT_IDs` -> yes
- `SW_SERIAL0` -> no

"SW_SERIAL0" is "software serial 0"

rwaldron:
Needs rebasing :P

lib/gps.js
dtex: Well, this is a pickle... How can we pick a default serial port ID? Should we always require it from the user? Something like this in each of the IO's would make it easy:
````
// ids for hardware and software serial ports on the board
  this.SERIAL_PORT_IDs = {
    DEFAULT: 0x08,
    HW_SERIAL0: 0x00,
    HW_SERIAL1: 0x01,
    HW_SERIAL2: 0x02,
    HW_SERIAL3: 0x03,
    SW_SERIAL0: 0x08,
    SW_SERIAL1: 0x09,
    SW_SERIAL2: 0x10,
    SW_SERIAL3: 0x11
  };
````

File: lib/gps.js (hunk 11 above)
Comment (dtex) [inline lib/gps.js]: I don't understand this comment.

File: lib/gps.js (hunk 10 above)
Comment (rwaldron) [inline lib/gps.js]: I don't know yet... This is all relatively new to me as well. We require a pin for basically everything, this is similar to a pin. Most platforms will only have one serial/UART available... I think we can safely say "each platform must define a default" (per your other comment which seems to have disappeared)

Comment (dtex) [inline lib/gps.js]: Ya, I axed it. It seems we could use ```this.io.SERIAL_PORT_IDs[0]``` as the default which would require re-ordering them in firmata.js or add ```this.io.DEFAULT_SERIAL_PORT_ID```

Comment (rwaldron) [inline lib/gps.js]: I think the latter is reasonable

Comment (soundanalogous) [inline lib/gps.js]: For firmata you can determine the default because there is a Serial pin type (but only for UARTs). If the configuration query indicates a board has Serial pins, it has one or more available UARTS (total number of UARTS is Serial pins / 2). Currently Serial0 (which is `Serial` rather than `Serial1`, etc) in Arduino is never reported. Otherwise you'd have to use SW serial as the default.

Comment (dtex) [inline lib/gps.js]: I was leaning toward SW serial because I've always considered the Uno to be our prototypical use case. So if the number of detected Serial pins === 0 I could assume I'm talking to an Uno or other board with just the one UART and should use SW_SERIAL0. That way Mega users would get HW_SERIAL0 by default and ...

You know what, this is just going to confuse people. I should go back to requiring the port to be specified. It's a little more configuration, but it will be more clear to users what's going on.

Comment (soundanalogous) [inline lib/gps.js]: When defaulting to SW serial, it may also be wise to recommend specific pins. There are all sorts of tricky issues with SW serial because it relies on pin change interrupts. For example is you specify SW serial pins on the same port (the AVR port - not the Firmata port) that you are using for any digital input or output, it will mess up the SW serial transmission but not throw an error. Super frustrating to debug if the user doesn't understand the cause.

**Comment by dtex** [comment]
Finally got this rebased and squashed into one commit. I've got a wiki page ready to go and I'll write up an announcement tomorrow.

File: lib/gps.js [hunk 12]
```diff
@@ -0,0 +1,418 @@
...
+
+util.inherits(GPS, events.EventEmitter);
+
+/*
+ * Default intialization for serial GPS
+ */
+GPS.prototype.initialize = function(opts) {
+
+  var state = priv.get(this);
+  state.serialPort = opts.serialPort || opts.portId || opts.port || opts.bus || this.io.SERIAL_PORT_IDs.SW_SERIAL0 || this.io.SERIAL_PORT_IDs[0];
```
rwaldron: `this.io.SERIAL_PORT_IDs[0]` => `this.io.SERIAL_PORT_IDs.DEFAULT`

I'm going to write that into the IO Plugin spec.

dtex: I think that will require a change in firmata because in firmata io.SERIAL_PORT_IDs[0] ===  io.SERIAL_PORT_IDs.HW_SERIAL0 which we would never want as the default

rwaldron: Naming nit pick: drop `serialPort` entirely:

```js
state.portId = opts.portId || opts.port || opts.bus || this.io.SERIAL_PORT_IDs.DEFAULT;
```

rwaldron: > I think that will require a breaking* change in firmata because in firmata io.SERIAL_PORT_IDs[0] === io.SERIAL_PORT_IDs.HW_SERIAL0 which we would never want as the default


But there is no such thing as `io.SERIAL_PORT_IDs[0] ` :P 

I've just published Firmata v0.10.0, which has `this.io.SERIAL_PORT_IDs.DEFAULT` that has the same value as `this.io.SERIAL_PORT_IDs.SW_SERIAL0`

rwaldron:
Where is the GPS wiki page that you mentioned?

eg/gps-hardware-serial.js
```diff
@@ -0,0 +1,18 @@
+var five = require("../lib/johnny-five.js");
+var board = new five.Board();
+
+board.on("ready", function() {
+
+  /*
+   * Explicitly setting HW_SERIAL1 for the port
+   */
+  var gps = new five.GPS({
+    port: "HW_SERIAL1"
```
rwaldron: This won't work because `port` is expecting the value of `SERIAL_PORT_IDs.HW_SERIAL1`, not the key itself. I'd like to make this possible, but would prefer to wait until @nebrius has settled his implementation of the serial extensions in Raspi-IO (to see how the patterns will intersect). 

For now let's drop `gps-hardware-serial.js` from `tpl/programs.json`. I don't want to discard it, but I don't want to mislead users

lib/gps.js
```diff
@@ -0,0 +1,418 @@
...
+    baud: this.baud,
+    rxPin: this.pins.rx,
+    txPin: this.pins.tx
+  });
+
+  if (this.configure) {
+    this.configure(function() {
+      this.listen();
+      if (opts.freq) {
+        this.frequency(opts.freq);
```
rwaldron: Just noticed this... I made a mistake 4 years ago in using `freq` as the property name for values that are milliseconds and not hertz. This use is hertz, but has the same name as the mis-named property in every other class. Eventually, I want to fix this throughout, but in the meantime I want to reserve the word "frequency" to be used only as a data property of either the constructor options or a component instance, so it can't be a function (sorry). I will volunteer to make this change if you're not sure what I'm driving at, but let me try to run it down: 

1. Currently, `frequency` is a function, change this to an accessor property. 
2. Currently, `freq` is a property of `opts`, change this to `frequency`
3. Currently, `this.frequency(opts.freq)` is called to send the frequency command and set the initial value of `state.frequency`, change this to `this.frequency = opts.frequency;`

rwaldron:
I see that instances have `gps.pos.lat, gps.pos.lng`, but I really want to be able to write: 

```js
gps.on("change", function() {
   console.log(this.latitude, this.longitude);
});
```

test/gps.js
```diff
@@ -0,0 +1,235 @@
...
+
+    this.serialConfig.reset();
+    this.serialWrite.reset();
+    test.done();
+  },
+
+  sendCommand: function(test) {
+    test.expect(2);
+
+    this.gps.sendCommand("Hey,laser,lips,your,mama,was,a,snow,blower\r\n");
```
rwaldron: Best test command ever? Yep.

lib/gps.js
dtex: I get what you're saying re: changing frequency to an accessor property and can make that change. 

Won't hertz be the inverse on frequency? (1hz = 1000ms frequency, 1000hz = 1ms frequency)

eg/gps-hardware-serial.js
dtex: I'll remove this from program.json. That should have been ```this.io.SERIAL_PORT_IDs.HW_SERIAL1```... I'm starting to rush things and am making mistakes :-(

dtex:
I chose this shape for pos because it can just be passed into the Google Maps Javascript API for setting a location, marker or whatever (it matches the Maps LatLng class). No problem changing it though.

dtex:
Here's that wiki page draft:

https://github.com/dtex/johnny-five/wiki/GPS

soundanalogous:
You may want to add a note in the GPS wiki that some Arduino boards do not support SoftwareSerial - namely the SAM an SAMD architectures (Due and Zero).

eg/gps-hardware-serial.js
```diff
@@ -0,0 +1,18 @@
+var five = require("../lib/johnny-five.js");
+var board = new five.Board();
+
+board.on("ready", function() {
+
+  /*
+   * Explicitly setting HW_SERIAL1 for the port
+   */
+  var gps = new five.GPS({
+    port: this.io.SERIAL_PORT_IDs.HW_SERIAL1
```
rwaldron: I'm not going to block on this for now, but I'm going to think about how we can make this less icky

lib/gps.js
```diff
@@ -0,0 +1,439 @@
...
+    longitude: {
+      get: function() {
+        return state.longitude;
+      }
+    },
+    altitude: {
+      get: function() {
+        return state.altitude;
+      }
+    },
```
rwaldron: Yesssss!

lib/gps.js
```diff
@@ -0,0 +1,439 @@
...
+
+        // Enforce maximum frequency of 10hz
+        if (freq < 100) {
+          freq = 100;
+        }
+
+        this.sendCommand("$PMTK220,"+String(freq));
+        state.frequency = freq;
+      }
+    },
```
rwaldron: Perfect

lib/gps.js
```diff
@@ -0,0 +1,439 @@
...
+    return;
+  }
+
+  this.emit("message", sentence);
+
+  var sentenceArr = cksum[0].split(",");
+  var last = {
+    lat: state.latitude,
+    lng: state.longitude,
+    alt: state.altitude,
```
rwaldron: These should be `latitude`, `longitude` and `altitude`

lib/gps.js
```diff
@@ -0,0 +1,439 @@
...
+
+// Convert Lat or Lng to decimal degrees
+function degToDec(degrees, degLength, dir, precision) {
+
+  if (degrees) {
+    var decimalDegrees = Number(degrees.substring(0, degLength)) + Number(degrees.substring(degLength))/60;
+    if (dir === "S" || dir === "W") {
+      decimalDegrees = decimalDegrees * -1;
+    }
+    return decimalDegrees.toPrecision(precision);
```
rwaldron: `toPrecision` returns a string, so you'll want to wrap this in `Number(...)`

lib/gps.js
rwaldron: Sorry, I just meant that if the data is a "sentence", maybe the event should be "sentence". It's easier to think about it like "the `"sentence"` event emits `sentence` data"—what do you think?

lib/gps.js
```diff
@@ -0,0 +1,439 @@
...
+   */
+  MT3339: {
+    baud: {
+      value: 9600
+    },
+    configure: {
+      value: function(callback) {
+        callback();
+      }
+    },
```
rwaldron: I forgot to mention this before—I really like what you're doing here, really excellent foresight. One thing I'm thinking of now, should this be async?

```js   
process.nextTick(callback);
```

It seems like future configuration operations may very well be the kind that require some degree of async behavior—we might as well enforce async now, even if nothing is actually happening

lib/gps.js
```diff
@@ -0,0 +1,439 @@
...
+  // Create a "state" entry for privately
+  // storing the state of the instance
+  state = {
+    sat: {},
+    latitude: 0.0,
+    longitude: 0.0,
+    altitude: 0.0,
+    speed: 0.0,
+    course: 0.0,
+    frequency: 1000,
```
rwaldron: 1000 => 1

lib/gps.js
```diff
@@ -0,0 +1,439 @@
...
+
+  var breakout, receiver, chip, state;
+
+  if (!(this instanceof GPS)) {
+    return new GPS(opts);
+  }
+
+  // Allow users to pass in a 2 element array for tx and rx pins
+  if (Array.isArray(opts)) {
+    opts = { pins: {tx: opts[0], rx: opts[1], onOff: opts[2] }};
```
rwaldron: Boards always print these in the order: RX, TX. 

Also: 

```c
SoftwareSerial mySerial(10, 11); // RX, TX
```

Therefore, `GPS` should expect them in this order: 

```js
five.GPS([ RX, TX, onOff ]);
```

lib/gps.js
```diff
@@ -0,0 +1,439 @@
...
+
+  // If a receiver was defined or derived but chip was not
+  if (receiver && !chip && Receivers[receiver].chip) {
+    chip = Receivers[receiver].chip.value;
+  }
+
+  // Default case
+  if (!chip) {
+    chip = "DEFAULT";
+  }
```
rwaldron: Lines 152-159 can be consolidated: 

```js
// If still no chip...
if (!chip) {
  
  // If receiver defined and has chip...
  if (receiver && Receivers[receiver].chip) {
    chip = Receivers[receiver].chip.value;
  } else {
  // Default case
    chip = "DEFAULT";
  }
}
```

lib/gps.js
dtex: Also toFixed makes more sense.

PR #1011, opened 2016-01-13
set version of lodash

nloding:
lodash 4.0.0 dropped `pluck` method, so until it can be updated to use `map` drop to 3.10.1


rwaldron:
Thanks for catching this!!

PR #1015, opened 2016-01-17
Max31850

BrianGenisio:
Add support for the MAX31850K thermometer component: https://www.adafruit.com/products/1727

This component requires OneWire support to be turned on in the ConfigurableFirmata.  

@bstein5 please review


rwaldron:
How did I miss this 0_o? I will get one of these to test asap

bstein5:
Fixed the freeze-up error, just a small addition. 
Basically during the isConversionAvailable function the device is supposed to reset and spit back all 0's when ready, but occasionally (still not entirely sure why, but it generally occurs after heating up the thermocouple) it doesn't reset correctly and the program freezes up. So this checks if the second data point is a zero, if not it returns. There's prob ably a more elegant solution, but this at least works.

` isConversionAvailable = function(done) {
              var nextDevice;

              if (devicesToWait.length === 0) {
                return done();
              }
              nextDevice = devicesToWait.pop();
              board.io.sendOneWireReset(pin);
              board.io.sendOneWireWriteAndRead(pin, nextDevice, CONSTANTS.READ_SCRATCHPAD_COMMAND, CONSTANTS.READ_COUNT, function(err, data) {
                if (!data[0]) {
                  devicesToWait.push(nextDevice);
                  if (data[1] != 0) {  //*****checks if second data bit is 0, if not its an error and gets thrown out
                    return done();
                  }
                }
              isConversionAvailable(done);

          });
            }.bind(this);`

BrianGenisio:
@rwaldron This PR is working, and has been tested by myself and @bstein5.  

I can't figure out why Travis is unhappy.  When I run `grunt examples` on my branch, I don't get any modified files, and `grunt test-examples` is happy on my machine as well.  I'm perplexed by this.

test/thermometer.js
```diff
@@ -377,6 +377,169 @@ exports["Thermometer -- ANALOG"] = {
...
+    this.sendOneWireSearch = this.sandbox.spy(MockFirmata.prototype, "sendOneWireSearch");
+    this.sendOneWireDelay = this.sandbox.spy(MockFirmata.prototype, "sendOneWireDelay");
+    this.sendOneWireReset = this.sandbox.spy(MockFirmata.prototype, "sendOneWireReset");
+    this.sendOneWireWrite = this.sandbox.spy(MockFirmata.prototype, "sendOneWireWrite");
+    this.sendOneWireWriteAndRead = this.sandbox.spy(MockFirmata.prototype, "sendOneWireWriteAndRead");
+
+    done();
+  },
+
+  tearDown: function(done) {
```
rwaldron: Need to call `Board.purge()` (I can add this in after I land it)

rwaldron: Just kidding, I forgot this file had been renovated to include a global tearDown

rwaldron:
Great work everyone!

rwaldron:
@BrianGenisio or @bstein5 can either of you write up a "release" announcement for this controller? https://github.com/j5js/johnny-five.io/tree/master/src/news

(if @bstein5 wants to do this, don't forget to add yourself here: https://github.com/j5js/johnny-five.io/blob/master/src/authors.json )

PR #1016, opened 2016-01-18
WIP: Refactor LEDControl into Controllers and Devices

### LINKED ISSUE #665: refactor LedControl

ericandrewlewis:
Work in progress to fix #665. Feedback welcome :smile:. 

I'm following patterns existing in the Motor module, separating logic out into "devices" (e.g. 8x8 bicolor) and "controllers" (e.g. `HT16K33`). 

@dtex @rwaldron @boneskull can one of you confirm this is a good path forward? Happy to continue!

### DIFF
--- a/lib/led/ledcontrol.js
+++ b/lib/led/ledcontrol.js
@@ -59,7 +59,6 @@ var priv = new Map(),
  *                                    Only for use with the HT16K33 controller.
  */
 function LedControl(opts) {
-
   Board.Component.call(
     this, opts = Board.Options(opts)
   );
@@ -76,22 +75,24 @@ function LedControl(opts) {
     controller = opts.controller;
   }
 
+  var device = null;
+  // Allow users to pass in custom device types
+  if ( typeof opts.device === "undefined" ) {
+    // figure out device implictly if it's not supplied.
+    device = Devices["AdafruitSingleColorMatrix"];
+  } else if ( typeof opts.device === "string" ) {
+    device = Devices[opts.device];
+  } else {
+    device = opts.device;
+  }
+  // Mixin properties from the device (e.g. seven-segment display).
+  Object.defineProperties(this, device);
+
   if (typeof controller === "undefined") {
     controller = Controllers.DEFAULT;
   }
 
-  // functions from Controller interface
-
-  this.clear = controller.clear;
-  this.led = controller.led;
-  this.row = controller.row;
-  this.scanLimit = controller.scanLimit;
-  this.send = controller.send;
-  this.sendDigit = controller.sendDigit;
-  this.initialize = controller.initialize;
-
-  // controller specific op codes
-  this.OP = controller.OP;
+  Board.Controller.call(this, controller, opts);
 
   // digit indexes may be ordered left to right (1) or reversed (-1)
   this.digitOrder = 1;
@@ -101,13 +102,17 @@ function LedControl(opts) {
     this.colon = opts.colon || false;
   }
 
-  // extra functions for HT16K33 devices only
-  if (controller.writeDisplay) {
-    this.writeDisplay = controller.writeDisplay;
+  opts.dims = opts.dims || LedControl.MATRIX_DIMENSIONS["8x8"];
+  if (typeof opts.dims === "string") {
+    opts.dims = LedControl.MATRIX_DIMENSIONS[opts.dims];
   }
-  if (controller.blink) {
-    this.blink = controller.blink;
+  if (Array.isArray(opts.dims)) {
+    opts.dims = {
+      rows: opts.dims[0],
+      columns: opts.dims[1],
+    };
   }
+
   /*
     devices variable indicates number of connected LED devices
     Here's an example of multiple devices:
@@ -121,16 +126,6 @@ function LedControl(opts) {
   for (var i = 0; i < 64; i++) {
     this.status[i] = 0x00;
   }
-  opts.dims = opts.dims || LedControl.MATRIX_DIMENSIONS["8x8"];
-  if (typeof opts.dims === "string") {
-    opts.dims = LedControl.MATRIX_DIMENSIONS[opts.dims];
-  }
-  if (Array.isArray(opts.dims)) {
-    opts.dims = {
-      rows: opts.dims[0],
-      columns: opts.dims[1],
-    };
-  }
   var state = {
     devices: devices,
     digits: opts.digits || 8,
@@ -140,6 +135,7 @@ function LedControl(opts) {
     columns: opts.dims.columns
   };
 
+  // @maybetodo Move this to LED device initializers?
   if (!(state.columns === 8 || state.columns === 16) || !(state.rows === 8 || state.rows === 16) || (state.columns + state.rows === 32)) {
     throw new Error("Invalid matrix dimensions specified: must be 8x8, 16x8 or 8x16");
   }
@@ -178,7 +174,8 @@ function LedControl(opts) {
   });
 
   priv.set(this, state);
-  controller.initialize.call(this, opts);
+
+  this.initialize(opts);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ericandrewlewis:
Going to continue with this ;) Will post here with updates, any feedback as progress developers appreciated!

rwaldron:
I have not forgotten about this, just super busy. I will put together a test rig hopefully this weekend

lib/led/ledcontrol.js
```diff
@@ -59,9 +59,8 @@ var priv = new Map(),
  *                                    Only for use with the HT16K33 controller.
  */
 function LedControl(opts) {
-
   Board.Component.call(
-    this, opts = Board.Options(opts)
+    this, this.opts = opts = Board.Options(opts)
```
rwaldron: Don't assign `opts` to `this.opts`. Some classes have this and it's a mistake that needs to be rectified

lib/led/ledcontrol.js
```diff
@@ -80,18 +79,7 @@ function LedControl(opts) {
...
-  this.led = controller.led;
-  this.row = controller.row;
-  this.scanLimit = controller.scanLimit;
-  this.send = controller.send;
-  this.sendDigit = controller.sendDigit;
-  this.initialize = controller.initialize;
-
-  // controller specific op codes
-  this.OP = controller.OP;
+  Board.Controller.call(this, controller, opts);
```
rwaldron: I believe Controllers are setup after Device? cc @dtex

ericandrewlewis:
@rwaldron no worries! I've been meaning to give a plain-text status update since this is such a huge refactor. Waiting for a component order to come in so I can test all the devices.

lib/led/ledcontrol.js
ericandrewlewis: Yep, that's [the order the motor module takes](https://github.com/rwaldron/johnny-five/blob/master/lib/motor.js#L529-L544), I'll update accordingly.

dtex: Sorry, totally missed the mention. Yes, controller overrides device so :+1:

lib/led/ledcontrol.js
dtex: reflectanceArray and motor are the offenders. I will fix the latter.

ericandrewlewis:
Adjusted the commit based on feedback.

To recap what's going on before going further in this PR: what were previously called "controllers" (e.g. HT16K33) are being split into devices ("matrix display" and "seven-segment display") and controllers (HT16K33). Both of these will be mixins on an LedControl instance that utilizes `Object.defineProperties()`. 

Devices should track state of themselves. e.g. the matrix display tracks which pixels are on and off. 

The role of the controller is to write to the physical controller. 

This generally means that a matrix device will contain methods like `led()` (to turn on an led in a matrix) and `row()` (to set the state of a row in a matrix).

I notice that the 7219 is [written to one row at a time](https://github.com/rwaldron/johnny-five/blob/fd560084c3a150e5d08f50766d7fba10cf425e5c/lib/led/ledcontrol.js#L965-L991) and the HT16K33 is [written to all at once](https://github.com/rwaldron/johnny-five/blob/fd560084c3a150e5d08f50766d7fba10cf425e5c/lib/led/ledcontrol.js#L680-L688). To extract the `led()` method from both of these controllers into a single method on the "matrix" device will require some thought, since in the `led()` method we'll want to set state and then write that to the display. Should we create a new `writeDisplay()` method on the 7219 controller?

PR #1023, opened 2016-01-29
Bno055

yakmoose:
This is a series of updates to the BNO module.
Includes some documentation, bug fixes, additional configuration options (crystal and chip orientation) and introduces a new orientation class that presents the absolute orientation of the sensor as Euler angles and quaternions.

Does not yet include unit configurations.


yakmoose:
Will go and fix the jshint errors

PR #1028, opened 2016-02-03
Add browser-serialport dependency for NW.js support via Firmata

cotejp:
I submitted a very similar and related pull request for the Firmata module (https://github.com/jgautier/firmata/pull/108). If both PRs are merged in, NW.js v0.13 and Johnny-Five would work together right out of the box. The huge gain is that it would no longer be necessary to custom-compile node-serialport for NW.js (which is prohibitively complicated for beginners).

With the small changes I'm suggesting in this PR and the one for Firmata, I was able to successfully use Arduino + browser-serialport + Firmata + Johnny-Five + NW.js. This is an amazingly powerful combination for physical computing and robotics.

### DIFF
--- a/lib/board.js
+++ b/lib/board.js
@@ -68,9 +68,16 @@ var Serial = {
   attempts: [],
 
   detect: function(callback) {
-    var serialport = IS_TEST_MODE ?
-      require("mock-firmata").SerialPort :
-      require("serialport");
+      
+    var serialport;
+    
+    if (IS_TEST_MODE) {
+      serialport = require("mock-firmata").SerialPort;
+    } else if (parseFloat(process.versions.nw) >= 0.13) {
+      serialport = require("browser-serialport");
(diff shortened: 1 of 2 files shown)


cotejp:
According to AppVeyor, my tiny change is not compatible with version 0.12 but is with version 0.10. The `Thermometer` test cases are not happy. Does anybody know if it's actually related to this pull request or if it's something else entirely? I tried looking into it but I'm not sure what the problem is.

rwaldron:
> According to AppVeyor, my tiny change is not compatible with version 0.12 but is with version 0.10. The Thermometer test cases are not happy. Does anybody know if it's actually related to this pull request or if it's something else entirely? I tried looking into it but I'm not sure what the problem is.


The failure is unrelated, I just re-ran this to get it green (because I knew it would pass on re-run)

PR #1030, opened 2016-02-04
Add support for CSS color names to RGB.color()

byronhulcher:
After  a discussion with @rwaldron at the recent Nodebots NYC meetup, I wanted to add support for CSS color names when setting the color of an RGB LED.  As it turns out, one of the already available dependencies, `color-convert`, can do that for us.  This could be useful for people not comfortable with hex color codes.  

Example usage:
```
 // Initialize the RGB LED
  var led = new five.Led.RGB({
    pins: {
      red: 6,
      green: 5,
      blue: 3
    }
  });

  // Turn it on and set the initial color
  led.on();
  led.color("green");
```

### DIFF
--- a/lib/led/rgb.js
+++ b/lib/led/rgb.js
@@ -311,18 +311,7 @@ RGB.prototype.color = function(red, green, blue) {
         
       } else {
         // color name
-        input = converter.keyword2rgb(input.toLowerCase());
-
-        // we are unable to match the name
-        if (typeof input === "undefined") {
-          throw new Error("Led.RGB.color: invalid color (" + red + ")");
-        }
-
-        update = {
-          red: input[0],
-          green: input[1],
-          blue: input[2]
-        };
+        return this.color(converter.keyword2rgb(input.toLowerCase()));
       }
     }
   } else {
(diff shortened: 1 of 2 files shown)


lib/led/rgb.js
```diff
@@ -292,21 +293,37 @@ RGB.prototype.color = function(red, green, blue) {
...
+          red: parseInt(input.slice(0, 2), 16),
+          green: parseInt(input.slice(2, 4), 16),
+          blue: parseInt(input.slice(4, 6), 16)
+        };
+
+        
+      } else {
+        // color name
+        input = converter.keyword2rgb(input.toLowerCase());
+
```
rwaldron: What do you think about this: 

```js
return this.color(converter.keyword2rgb(input.toLowerCase()));
```

That way we avoid the duplication on lines 317-325 and make use of the existing byte array handling.

byronhulcher: I like it! Was thinking that while copy/pasting but was hesitant to add recursion.  Will do

rwaldron:
Awesome!

PR #1032, opened 2016-02-10
fixed 74HC595 expander

tronsha:


lib/expander.js
```diff
@@ -1392,12 +1392,12 @@ var Controllers = {
...
-          throw new Error("Expected pins.data");
+        if (!opts.pins.clock) {
+          throw new Error("Expected pins.clock");
         }
 
-        if (!opts.pins.data) {
-          throw new Error("Expected pins.data");
+        if (!opts.pins.latch) {
+          throw new Error("Expected pins.latch");
         }
```
rwaldron: Copy pasta :(

Thanks for fixing that <3

PR #1041, opened 2016-02-23
BNO055 calibration events and configuration options

yakmoose:
Hi,

This is just a small update that improves the calibration state of the BNO055, it now emits calibration events from the readCalibration() steps and allows a calibration mask to be supplied that allows you to alter what you consider a fully calibrated system. 

I've also updated the test as there is now an additional data event that is emitted due to the calibration now emitting. Took a while to get my head around the testing, if there is a preferred way of doing this let me know.

Cheers
John


lib/imu.js
```diff
@@ -368,20 +373,25 @@ var Drivers = {
...
 
-                // if (state === 0xFF) {
-                  // change this so that we wait for the system
-                  // to be ready, not each component.
-                  // as per adafruit example suggestions
-                  // wonder if the calibration should be checked here, or where we are consuming?
-                if (s === 3) {
+                // emit the calibration state so we can work out in our userspace if
+                // we are good to go, and for when we are performing the calibration steps
+                this.emit("data", computed);
```
rwaldron: The `Motion` class has sensors that emit a "calibrated" event, let's make this match: `"data"` => `"calibrated"` (and then update the tests)

yakmoose: Cool, I'll look at how this works and  update this and the various subcomponents handling of this.

rwaldron:
Ping me when ready for another review :)

lib/imu.js
yakmoose: I'm pretty close to having it sorted, did a bunch of real world testing (bolted it to my telescope while looking at things). I've just committed where I'm up to, but I will need to update the tests to reflect the new events being emitted. 

The main thing that I'm trying to do here is make the calibration state changes visible more than yes I'm calibrated. The device is quite autonomous and will move between calibration states so you kind of need to know when the state changes.

yakmoose:
I'm pretty close to having it sorted, did a bunch of real world testing (bolted it to my telescope while looking at things). I've just committed where I'm up to, but I will need to update the tests to reflect the new events being emitted.

The main thing that I'm trying to do here is make the calibration state changes visible more than yes I'm calibrated. The device is quite autonomous and will move between calibration states so you kind of need to know when the state changes.

yakmoose:
Hi,

I've cleaned up the bno tests, added in the calibration events, and rebased against master.

Cheers
John

rwaldron:
@yakmoose awesome, I will take a look and get this landed asap.

PR #1046, opened 2016-02-29
Fix button's Fritzing schematics

sayanee:
All the 4 button pins were not aligned to the breadboard holes and was slightly confusing for beginners during a workshop. Hence, I amended both the Fritzing file and the exported `*.png` image.

Before [as seen on the documentation](http://johnny-five.io/examples/button/):

before

After - all 4 pins are connected on the breadboard:

button


rwaldron:
Ha! Great catch :)

Issue #815, opened 2015-06-01
Piezo update to controller 


rwaldron:
Suppport for PCA9685

https://github.com/Reinbert/pca9685/blob/master/examples/piezo.c

PR #1047, opened 2016-03-02
Mix in controller properties for `Piezo`

bcjordan:
Allow for custom piezo controller following [led controller](https://github.com/rwaldron/johnny-five/blob/master/lib/led/led.js#L142) property mix-in pattern. Mostly to support overriding of `setTone`/`noTone` per https://github.com/rwaldron/playground-io/pull/1, but may be of interest for #815.

@rwaldron let me know if you'd prefer I stub out more or use a different approach for mixing in, could do any of:

1. Stub out default controller behavior (add a `Controllers` map with an empty `DEFAULT`)
1. Move default piezo tone/no-tone behavior into the `DEFAULT` controller
1. Use some sort of different controller overriding pattern

TODO:

- [x] 1, 2, 3 above
- [x] Write test covering controller behavior

### DIFF
--- a/lib/piezo.js
+++ b/lib/piezo.js
@@ -8,7 +8,7 @@ var Controllers = {
 
   DEFAULT: {
     initialize: {
-      value: function(opts) {
+      value: function() {
         this.io.pinMode(this.pin, this.io.MODES.OUTPUT);
       }
     }
@@ -25,6 +25,8 @@ function Piezo(opts) {
   Board.Component.call(
     this, opts = Board.Options(opts)
   );
+
+  var controller;
 
   if (opts.controller && typeof opts.controller === "string") {
     controller = Controllers[opts.controller.toUpperCase()];
@@ -55,7 +57,7 @@ function Piezo(opts) {
   });
 
   if (typeof this.initialize === "function") {
-    this.initialize(opts);
+    this.initialize();
   }
 }
 
--- a/lib/piezo.js
+++ b/lib/piezo.js
@@ -5,11 +5,60 @@ var priv = new Map();
 
 
 var Controllers = {
-
+  /**
+   * Timer-based tone generator using digital high/low piezo.
+   */
   DEFAULT: {
     initialize: {
+      writable: true,
       value: function() {
         this.io.pinMode(this.pin, this.io.MODES.OUTPUT);
+      }
+    },
+    tone: {
+      writable: true,
+      value: function(tone, duration) {
+        if (isNaN(tone) || isNaN(duration)) {
+          // Very Bad Things happen if one tries to play a NaN tone
+          throw new Error(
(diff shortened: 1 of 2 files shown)


rwaldron:
I'd like to see the complete migration, if you're up for doing it :)

rwaldron:
This will get you started: https://github.com/rwaldron/johnny-five/tree/piezo-controller I've left all the nitty-gritty for you to do :)

Here's how to test: `grunt nodeunit:just:piezo && grunt nodeunit:just:extended/piezo`

bcjordan:
Perfect, on this tomorrow morning, thanks!

lib/piezo.js
```diff
@@ -355,4 +423,9 @@ Piezo.prototype.stop = function() {
   return this;
 };
 
+if (IS_TEST_MODE) {
+  Piezo.frequencyToTone = frequencyToTone;
```
bcjordan: Any preferred way to expose helper util methods for testing?

lib/piezo.js
```diff
@@ -229,38 +301,39 @@ Piezo.prototype.note = function(note, duration) {
...
-    value = value === 1 ? 0 : 1;
-    this.io.digitalWrite(this.pin, value);
+  return this.frequency(toneToFrequency(tone), duration);
+};
 
-    if ((timer.difTime / 1000000) > duration) {
-      clearTimer.call(this);
-    }
-  }.bind(this), null, tone + "u", function() {});
+Piezo.prototype.frequency = function(frequency, duration) {
```
bcjordan: One quirky part of this implementation I'd like thoughts on / alternatives, this requires the given controller to define at least one of `frequency` and `tone`. The Circuit Playground's sysex API, for example, works with `Piezo.prototype.frequency`-equivalent values (called `tone` in [Arduino land](https://www.arduino.cc/en/Reference/Tone)), while the internals of the default/digital piezo implementation work at the `Piezo.prototype.tone` level and treat frequency as the helper.

The intent with this approach is to reconcile controller implementations wanting to work with either tone OR frequency, without implementations needing to re-define or import the conversion methods.

Alternatively we could treat `Piezo.prototype.tone` as the only controller override-able method, just not sure whether that's more likely to be cumbersome for future controllers than `Piezo.prototype.frequency`.

bcjordan:
@rwaldron alright, ready for another look! & added a couple of line note comments with design choice Qs.

lib/piezo.js
rwaldron: You can use this same silliness if you want. It's "meh", but works and stays out of the way of non-test-runtime API

rwaldron:
@bcjordan Landed!

bcjordan:
@rwaldron awesome, thanks for your help on this!

PR #1048, opened 2016-03-02
Added sensitivity option for MPR121QR2 (to replace hardcoded value)

cotejp:


rwaldron:
I've landed the version that we worked together to define today :) Thanks for being patient and working through each step with me. Now let's tackle this weird component identity problem!

cotejp:
Thanks to you for making all this! Anytime you need, just ping me!

PR #1050, opened 2016-03-04
added node-open-pilot project

darioodiaz:
Node Open Pilot is an opensource software and openhardware project yo build your own drone using a firmata-board compatible and johnny-five


TSC21:
I though, when reading it was called Open Pilot, you have made a bridge JF<->UAVTalk, which is the communication protocol used on (OpenPilot)[https://www.openpilot.org/]. But seems it has nothing to do with it. I recommend you change your project name, @darioodiaz, as it may lead to this kind of wrong interpretations...

darioodiaz:
Hi @TSC21 I name it "Node Open Pilot", because is opensource, a pilot software and written in Nodejs :p, but now that you mention I'll change the project name to avoid confusion, thanks for the suggestion

PR #1059, opened 2016-03-14
Fix BMP085 file name

dtex:


rwaldron:
You don't need to make a PR for this—I definitely trust you ;P

PR #1066, opened 2016-03-22
The within() function now properly works with floating point numbers.

cotejp:


rwaldron:
The failure shown in appveyor is unrelated to this fix.

PR #1074, opened 2016-03-29
Add spark fun gps

makenai:
This adds support for the SparkFun Venus GPS module:  https://www.sparkfun.com/products/11058


makenai:
I'm not sure what's up with the serialport test failing. Is there anything I need to do to fix that?

rwaldron:
No, it's some strange thing that appeared recently :(

lib/gps.js
```diff
@@ -308,6 +397,47 @@ GPS.prototype.sendCommand = function(string) {
   this.io.serialWrite(state.portId, cc);
 };
 
+/*
+ * SkyTraq Binary Message Protocol
+ * http://cdn.sparkfun.com/datasheets/Sensors/GPS/Venus/638/doc/AN0003_v1.4.19.pdf
+ */
+GPS.prototype.sendSkyTraqCommand = function(payload, callback) {
```
rwaldron: Is this specific to _just_ this model? There shouldn't be component specific methods like this, so I think you and @dtex should coordinate on this implementation.

lib/gps.js
```diff
@@ -432,6 +562,40 @@ GPS.prototype.parseNmeaSentence = function(sentence) {
 
 };
 
+/*
+ * SkyTraq Binary Message Protocol
+ * http://cdn.sparkfun.com/datasheets/Sensors/GPS/Venus/638/doc/AN0003_v1.4.19.pdf
+ */
+GPS.prototype.parseSkyTraqSentence = function(sentence) {
```
rwaldron: I guess the same question as above, with same suggestion. Basically: what would happen if this method were called on a GPS that wasn't this model? Probably nothing useful? Maybe an error? 

Maybe GPS class design needs to be revisited? I really wanted this stuff sort out before it was landed, and I thought it had been :\

makenai: Those methods will only be called on that specific model due to a broken out 'listen' handle for this chipset, and chipset specific commands. Is the issue just with having the utility functions available?

rwaldron: But by putting them here, they are defined for all GPS instances, which I don't want. I think that the whole GPS class needs to be re-evaluated such that it can handle disparate sources in a uniform way. 

> Is the issue just with having the utility functions available?

I don't want component-specific things on prototypes. That's why we have controller and breakout definitions, but I think something isn't lining up in GPS. I want @dtex to take a look at this and also @BrianGenisio

dtex: If you have a broken out listen method you can just put the custom parser method on the CHIP object but I don't think that's a good idea...

I'm sure there will be other chips that use Skytraq (and multiple chips that use SiRF, Rockwell and Sony formats) but DRY is good so I have a proposition:

Let's create a PARSER object that wraps each of the parsing methods, and then the chip can specify which method on PARSER to use. i.e.:

````
PARSER = {
  NMEA: function(sentence) {},
  SKYTRAQ: function(sentence) { ... },
  etc...
}

// and then on CHIP

Venus638FLPx: {
  parse: PARSER.SKYTRAQ,
  baud: {
    value: 9600
  },
  ...
````

rwaldron:
@dtex @makenai what's the word?

dtex:
It looks like the heavy lifting is done, just some refactoring. How do you (@rwaldron) feel about that proposed PARSER object?

rwaldron:
@dtex but there is still both a `sendCommand` and `sendSkyTraqCommand`? Is that going to be fixed with the new parser thing you're describing? I think that sounds like a good solution.

rwaldron:
Also, needs one of these: `git pull --rebase upstream master`

dtex:
sendCommand will vary with the chip but parsers will not so just leave sendCommand on the prototype and override that with a custom sendCommand on CHIP.VENUS638FLPX.

rwaldron:
> so just leave sendCommand on the prototype and override that with a custom sendCommand on CHIP.VENUS638FLPX.

Yeah, this is what I wanted. I don't want another sort-of-like-sendCommand-thing.

rwaldron:
@dtex @makenai what's the next step?

makenai:
@rwaldron I can definitely it a shot at that, but it would take me a couple of weeks to get around to it right now. If @dtex or anyone with a stronger idea of how it should work would like to get it started, I'm open to that too!

dtex:
I can work on those changes. I've got a pretty good idea of what needs to happen. Is the best way to get this in is to fork makenai:addSparkFunGPS and make a PR to that, or should I just open a new PR here?

rwaldron:
> Is the best way to get this in is to fork makenai:addSparkFunGPS and make a PR to that

That's what I would recommend doing :) it will be trivial for @makenai to land it when you're ready

dtex:
Hi @makenai,

Are you back home? NBD is behind us and I wanted to get Tessel GPS support added but we should land this first. Any plans to work on it soon? If not, I could probably get my hands on a Venus and apply the finishing touches.

Your trip looked awesome BTW. I so want to visit there.

makenai:
Yup, I'm back! And this has been weighing heavily on my conscience. :) I'm at CascadiaFest this week, but I'm going to plan to take a look at it during / possibly a bit after depending on my motivation level.

dtex:
Okay, have fun at Cascadia. I was there last year and had a really good time.

rwaldron:
🔔

dtex:
@makenai I'm anxious to land support for the Tessel GPS module. I want [this ball of cups](https://www.youtube.com/watch?v=56h60ul2fnA&feature=youtu.be&t=34) to be able to drive itself around the block (Halloween is coming and how fun would it be to add some lights and let it go?). I've been holding off because I refactored a lot in my PR to your repo and I don't want to lose that. Also, I'm worried that if I submit my own PR with the Tessel GPS stuff it will make merging a ginormous PIA for you. Do you think you'll have it in the next couple of weeks?

dtex:
Hi @makenai, 

I hope all is well.

This PR has sat for a while and neither one of us have gotten around to it. I want to close it for cleanliness but I don't want the addition of the Venus to fall off the radar. To handle cases like this I've created a new [Requested Features](../wiki/Requested-Features) page in the wiki and added it to that list.

PR #1076, opened 2016-04-01
Implement basic safety mechanism

jugglinmike:
Use well-researched heuristic to ensure safe operation for all
Johhny-Five instances.

### DIFF
--- a/lib/fn.js
+++ b/lib/fn.js
@@ -219,4 +219,25 @@ Fn.uint24 = function(high, low, xlow) {
 };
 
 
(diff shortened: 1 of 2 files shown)


dtex:
Running this on process.nextTick will max out CPU usage. I want my safety to be at the frontmost of my robot's thoughts but I don't want it to be all consuming. It presumably has other things it needs to do :-P

jugglinmike:
My concern is that "other things it needs to do" may include "dismembering Mike."

rwaldron:
Thanks for the contribution <3

PR #1078, opened 2016-04-03
Add filtering to Proximity

frxnz:
Submitted for your approval, filtered data output for the Proximity class.

This adds a `filter` property, which should look like one of these:

``` js
var proximity = new five.Proximity({
  controller: "HCSR04",
  pin: 7,
  filter: true
});

var proximity = new five.Proximity({
  controller: "HCSR04",
  pin: 7,
  filter: {
    samples: 5,
    soften: 2,
    deviation: 2
  }
});

var proximity = new five.Proximity({
  controller: "HCSR04",
  pin: 7,
  filter: function (raw) {
    return Math.floor(raw);
  }
});
```

`samples` defines the size of an array which is used to compute a running average
`soften` defines the number of decimal points to round the output values
`deviation` defines the amount of acceptable variation in values

Once we're happy with the details I'll add tests.

### DIFF
--- a/lib/proximity.js
+++ b/lib/proximity.js
@@ -402,8 +402,8 @@ function Proximity(opts) {
   var controller = null;
   var state = {};
   var raw = 0;
-  var last = 0;
   var freq = opts.freq || 25;
+  var last = 0;
   var pinValue = typeof opts === "object" ? opts.pin : opts;
 
   Board.Device.call(
@@ -468,7 +468,7 @@ function Proximity(opts) {
     opts.pinValue = pinValue;
     this.initialize(opts, function(data) {
       raw = data;
-    }.bind(this));
+    });
   }
 
   setInterval(function() {
@@ -490,7 +490,6 @@ function Proximity(opts) {
       last = raw;
       this.emit("change", data);
     }
-
   }.bind(this), freq);
 
   Filtration.call(this, opts);
--- a/lib/proximity.js
+++ b/lib/proximity.js
@@ -468,7 +468,7 @@ function Proximity(opts) {
     opts.pinValue = pinValue;
     this.initialize(opts, function(data) {
       raw = data;
-    });
+    }.bind(this));
   }
 
   setInterval(function() {
--- a/lib/proximity.js
+++ b/lib/proximity.js
@@ -22,7 +22,7 @@ var Controllers = {
     },
     toCm: {
       value: function(raw) {
-        return +(12343.85 * Math.pow(raw, -1.15)).toFixed(2);
+        return +(12343.85 * Math.pow(raw, -1.15)).toFixed(3);
       }
     }
   },
@@ -33,7 +33,7 @@ var Controllers = {
     },
     toCm: {
       value: function(raw) {
-        return +((2914 / (raw + 5)) - 1).toFixed(2);
+        return +((2914 / (raw + 5)) - 1).toFixed(3);
       }
     }
   },
@@ -45,7 +45,7 @@ var Controllers = {
     },
     toCm: {
       value: function(raw) {
-        return +(10650.08 * Math.pow(raw, -0.935) - 10).toFixed(2);
+        return +(10650.08 * Math.pow(raw, -0.935) - 10).toFixed(3);
       }
     }
   },
@@ -57,7 +57,7 @@ var Controllers = {
     },
     toCm: {
       value: function(raw) {
-        return +(2076 / (raw - 11)).toFixed(2);
+        return +(2076 / (raw - 11)).toFixed(3);
       }
     }
   },
@@ -401,9 +401,20 @@ function Proximity(opts) {
   var controller = null;
   var state = {};
   var raw = 0;
-  var freq = opts.freq || 25;
   var last = 0;
+  var rawFiltered = 0;
+  var lastFiltered = 0;
+  var freq = opts.freq || 25;
   var pinValue = typeof opts === "object" ? opts.pin : opts;
+  var isFiltered = opts.filter;
+  var filterDefaults = {
+    samples: 5,
+    soften: 2,
+    deviation: 2
+  };
+  var filterOpts = typeof opts.filter === "object" ? Object.assign(filterDefaults, opts.filter) : filterDefaults;
+
+  this.samples = [];
 
   Board.Device.call(
     this, opts = Board.Options(opts)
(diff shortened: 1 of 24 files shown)


lib/proximity.js
```diff
@@ -401,9 +401,20 @@ function Proximity(opts) {
...
   var pinValue = typeof opts === "object" ? opts.pin : opts;
+  var isFiltered = opts.filter;
+  var filterDefaults = {
+    samples: 5,
+    soften: 2,
+    deviation: 2
+  };
+  var filterOpts = typeof opts.filter === "object" ? Object.assign(filterDefaults, opts.filter) : filterDefaults;
+
+  this.samples = [];
```
rwaldron: Once we extend this capability beyond Proximity, will the property name `samples` be problematic? I haven't looked

rwaldron:
I really like this feature and the minimal-yet-flexible surface that it exposes. I would like you to continue iterating on this, with the end goal to be something like a mixin or decorator that can be easily required and applied to all other input components.

lib/proximity.js
```diff
@@ -448,41 +459,80 @@ function Proximity(opts) {
...
+      }
+    }
   });
 
   if (typeof this.initialize === "function") {
     opts.pinValue = pinValue;
     this.initialize(opts, function(data) {
+
+      if (isFiltered && typeof filterOpts.deviation === "number") {
+        data = raw && data > raw * (filterOpts.deviation) ? raw : data;
```
rwaldron: Nits: 

- no parens around `filterOpts.deviation`
- put the parens around `data > raw * filterOpts.deviation`: `(data > raw * filterOpts.deviation)`

lib/proximity.js
```diff
@@ -448,41 +459,80 @@ function Proximity(opts) {
...
     if (raw === undefined) {
       return;
     }
 
-    var data = {
+    var data = isFiltered ? this.filtered : {
       cm: this.cm,
       centimeters: this.centimeters,
-      in: this.in,
+      in : this.in,
```
rwaldron: This space can be removed ;)

lib/proximity.js
```diff
@@ -448,41 +459,80 @@ function Proximity(opts) {
...
+        raw = average;
+
+        return {
+          cm: isSoften ? soften(this.cm) : this.cm,
+          centimeters: isSoften ? soften(this.centimeters) : this.centimeters,
+          in : isSoften ? soften(this.in) : this.in,
+          inches: isSoften ? soften(this.inches) : this.inches,
+        };
+
+        function soften(num) {
```
rwaldron: No functions after return

lib/proximity.js
frxnz: @rwaldron Problematic in what way? As far as naming I think it's the most appropriate of the 3 filter properties. `soften` I think is a little unintuitive, and `deviation` could use a little more though

lib/proximity.js
```diff
@@ -448,41 +459,80 @@ function Proximity(opts) {
...
     in: {
       get: function() {
         return this.inches;
       }
     },
+    filtered: {
+      get: typeof opts.filter === "function" ? opts.filter : function() {
+        var average = this.samples.reduce(function(a, b) {
+          return a + b;
+        }) / this.samples.length;
```
rwaldron: Use `Fn.sum(this.samples) / this.samples.length`

rwaldron: Or better yet... add `Fn.average(values)` ;)

lib/proximity.js
rwaldron: I just realized something: I'm not sure this is the data that should be operated on because `data` could be _very_ different depending on the source peripheral. Shouldn't the filtering be applied to the appropriate data for the component? For example, the `Proximity` class computes centimeter values, so centimeter values should be what is filtered (instead of time in μs).

rwaldron:
I put together a _super_ rough outline of what I'm thinking: 

- it's untested
- it preserves your exposed API (`filter: {... }`)
- it needs to be filled in with your filtering logic

```js
var Fn = require("../fn"); // assuming this file is /lib/mixins/filtration.js

var filterDefaults = {
  samples: 5,
  soften: 2,
  deviation: 2
};

module.exports = function(options) {
  var filterOpts = options.filter;
  var isFiltered = !!filterOpts;
  
  if (!isFiltered) {
    return;
  }

  filterOpts = typeof filterOpts === "object" ? 
    Object.assign({}, filterDefaults, opts.filter) : filterDefaults;

  var state = {
    previous: 0,
    samples: []
  };
  
  function soften() {
    // ... using filterOpts and so on...
  }

  Object.defineProperties(this, {
    filtered: {
      get: function() {
        // ...
        var average = Fn.average(state.samples);

        // other stuff...
      }
    }
  });

  this.on("data", function(data) {
    if (isFiltered && typeof filterOpts.deviation === "number") {
      data = state.previous && (data > state.previous * filterOpts.deviation) ? 
        state.previous : data;
    }
    
    if (!isFiltered || state.samples.length === filterOpts.samples) {
      state.samples.shift();
    }

    state.samples.push(data);    
  });
};

/*

  Component modules can add this like: 

  var Filtration = require("./mixins/filtration");


  function Component(opts) {


    // I think this needs to be called very last in the constructor
    Filtration.call(this, opts);
  }

 */
```

lib/proximity.js
frxnz: `deviation` is just a multiplier so applying to raw values should be fine, and it needs to happen where the samples are being collected. The rest of the filtration is indeed done on transformed values.

lib/proximity.js
frxnz: Mind reader. I was looking through Fn this morning, `Fn.average` forthcoming :)

frxnz:
That looks perfect, thanks for putting that together. I definitely wanted this to be something we could use for other classes, so the mixin approach is great. If we're happy with this as a proof of concept I can get started on that.

A couple things I'm questioning:

* The name `soften`. Should this just be `round`? It's the number of decimal places to round to
* The value for `deviation`. Ideally this would be a percent value, is using a float intuitive? (2 == 200%)
* Should there be a separate `data:filtered` event that is always available? Or is that adding unnecessary overhead?

I'd also love to get some feedback from @tralves since this was largely his idea :)

lib/proximity.js
rwaldron: Ok... I think that's going to be problematic. We're trying to hide the raw values because they are usually nonsensical as far as the user is concerned. Doing this here is also super invasive and makes it basically impossible to abstract this feature into something we can apply to all input things, which is definitely the goal. Maybe we just need more time to think about it...

lib/proximity.js
rwaldron: Just that I thought some other classes might have `samples` properties, but I can't find any that do.

rwaldron:
> That looks perfect, thanks for putting that together. I definitely wanted this to be something we could use for other classes, so the mixin approach is great. If we're happy with this as a proof of concept I can get started on that.

Yeah, I think it encapsulates your implementation well... _except_ for the deviation part (unless I've misunderstood... see my response in the changes tab?). Also... I realized that `data`, in this context: 

```js
  this.on("data", function(data) {
    if (isFiltered && typeof filterOpts.deviation === "number") {
      data = state.previous && (data > state.previous * filterOpts.deviation) ? 
        state.previous : data;
    }
```

Isn't always "just a number", for example... gyro, accelerometer, joystick

lib/proximity.js
frxnz: Ok good point. Thinking about it some more there's no reason it _has_ to be in the dataHandler callback. We'll just need to filter the `samples` array.

frxnz:
Just so I'm clear, getting the values from the `Filtration` class above would look like this:

```js
var prox = new five.Proximity({
    filter: true
});

prox.on('data', function () {
    console.log(this.filtered);
});
```

And ideally `this.filtered` would be on object like
```js
{
    cm: 2,
    centimeters: 2,
    in: 0.787,
    inches: 0.787
}
```

I need to think about this some more, but this definitely isn't going to be as simple as `Filtration.call(this, opts);`

rwaldron:
>  from the Filtration class above would look like this:
> ...

Yes, I think that's how I interpreted your vision. My concern is how to define filtering for things that have `x`, `y`, `z`. In reality, Proximity only has `cm`, right? But other things have more than one value

frxnz:
@rwaldron I added a `Filtration` mixin which eliminates all of the filtering logic from the component class. I ended up adding an array of filterable properties, which I know you were hoping to avoid. My thought is with the mixin in place, we can see how it all comes together, and hopefully come up with a better solution.

rwaldron:
Needs rebase

PR #1082, opened 2016-04-05
Make baud property settable.

### LINKED ISSUE #1075: GPS class conversion problem at gps.js line 317,318
- I'm trying to receive the NMEA massage from my custom GPS board and it works fine with [my Arduino IDE code](https://github.com/pacozaa/Final-Project-ArduinoCode/blob/master/TestGPS/TestGPS.ino).

gpsj5

- When come to johnny-five event 

> change or sentences not trigger. 

So I go to the gps.js class and debug it if the data come. Yes it is, but not properly parse.
gps1
gps2

and I try to change something on the gps.js but still not working for me
gpsraw1
gpsraw2

- Maybe because I use level shifter to convert the signal from GPS board to Arduino MEGA, but Why Arduino IDE can print it out?

dtex:
Fixes #1075

### DIFF
--- a/lib/gps.js
+++ b/lib/gps.js
@@ -39,7 +39,8 @@ var Chips = {
 
   DEFAULT: {
     baud: {
-      value: 9600
+      value: 9600,
+      writable: true
     },
     configure: {
       value: function(callback) {
@@ -53,7 +54,8 @@ var Chips = {
    */
   MT3339: {
     baud: {
-      value: 9600
+      value: 9600,
+      writable: true
     },
     configure: {
       value: function(callback) {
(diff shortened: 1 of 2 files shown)


dtex:
@pacozaa tested and this fix works, but I don't want to merge without knowing what's going on with Travis and Appveyor

pacozaa:
@dtex No problem :)

lib/gps.js
```diff
@@ -189,7 +189,7 @@ function GPS(opts) {
 
   // If necessary set default property values
   this.fixed = opts.fixed || 6;
-  this.baud = opts.baud || breakout.baud || chip.baud;
+  this.baud = opts.baud || this.defaultBaud;
```
rwaldron: Instead of making a new property, just do: 

```js
  DEFAULT: {
    baud: {
      writable: true,
      value: 9600
    },
```
and

```js
  MT3339: {
    baud: {
      writable: true,
      value: 9600
    },
```

rwaldron:
:+1: nice work dudes

Issue #1083, opened 2016-04-05
Should sensor's change event be better documented?


bartmichu:
Hi, I'm a bit confused with examples used across the site. Consider the following code:

```
var five = require("johnny-five");
var board = new five.Board();

board.on("ready", function () {
  var pot = new five.Sensor({
    pin: "A0",
    threshold: 10
  });

  pot.on("change", function () {
    console.log(this.value);
  });
});
```

I would assume that each consecutive entry will differ by at least 10, am I right? Turns out that's not the case because `this.value` returns new reading which is often different than the actual value read on change. 

Looking at the source code I see that function has a parameter, so I can use:

```
pot.on("change", function (value) {
  console.log(value);
});
```

No inconsistencies now. Not a big deal, but gives a wrong example in my opinion.
I didn't look into similar events on other sensors yet.

PR #1084, opened 2016-04-05
Use value read instead of a new one

bartmichu:
Addresses #1083


dtex:
@bartmichu Each of the eg files has a corresponding docs file that is generated by running ```grunt examples```. Here [more info](https://github.com/rwaldron/johnny-five/blob/master/CONTRIBUTING.md#writing-documentation) on that process.

bartmichu:
@dtex I will look into that, thanks.

PR #1089, opened 2016-04-09
Default custom property to object literal instead of null

bartmichu:
Currently, adding custom properties like that is not supported:
```
mydevice.custom.foo = "bar";
```
and it fails with ```TypeError: Cannot set property 'foo' of null```

Changing default value from ```null``` to a real object would fix this.


rwaldron:
I will agree to this, but I'd like it stated that I don't agree with assigning empty objects "just in case".

PR #1091, opened 2016-04-10
Pin aliasing

HipsterBrown:
Worked on this a bit with @rwaldron.

The labels for pins on development boards don't always match up with the property names for the pins object that is passed into the constructor for various classes, i.e. [the `Led.Digits` class usage](http://johnny-five.io/api/led.digits/#usage). Because of the non-standard naming for these pin labels, it's difficult for folks who don't know how their board's pins match up with the Johnny-Five API. 

This PR allows any of the known SPI aliases for `miso`, `clk`, `ss`, `mosi` to be used in for the properties of that pins object. 

**Concerns:**
I'm not sure how to reflect this in the various docs that show `data`, `clock`, and `cs` being used for those properties. 

This will also break any uses of `ShiftRegister.pins`.

### DIFF
--- a/lib/board.options.js
+++ b/lib/board.options.js
@@ -4,16 +4,17 @@
 
 var MISO = "miso";
 var MOSI = "mosi";
-var SCLK = "sclk";
+var CLK = "clk";
 var SS = "ss";
 
 // jshint unused:false
 var aliases = {
 
-  // SCLK
-  clk: SCLK,
-  clock: SCLK,
-  sclk: SCLK,
+  // CLK
+  clk: CLK,
+  clock: CLK,
+  sclk: CLK,
+  sck: CLK,
 
   // MISO
   somi: MISO,
@@ -29,6 +30,7 @@ var aliases = {
   sdi: MOSI,
   data: MOSI,
   di: MOSI,
+  dn: MOSI,
   din: MOSI,
   si: MOSI,
   mtst: MOSI,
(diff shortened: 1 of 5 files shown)


lib/board.options.js
```diff
@@ -74,7 +76,15 @@ function Options(arg) {
   } else {
     opts = arg;
 
+    if (opts && opts.pins != null) {
```
rwaldron: Something we overlooked: this should also check that `typeof opts.pins === "object"`

lib/led/ledcontrol.js
```diff
@@ -772,14 +772,9 @@ Controllers = {
...
   DEFAULT: {
     OP: {},
-    initialize: function(opts) {
+    initialize: function() {
 
-      this.pins = {
-        data: opts.pins.data,
-        clock: opts.pins.clock,
-        cs: opts.pins.cs || opts.pins.latch
-      };
```
rwaldron: For anyone wondering why this is removed, @HipsterBrown and I discovered that it's unnecessarily duplicating the efforts of `Board.Component`

lib/board.options.js
HipsterBrown: Will fix. Thanks for catching it.

rwaldron:
Doesn't this need to change? https://github.com/rwaldron/johnny-five/blob/master/lib/motor.js#L16

HipsterBrown:
@rwaldron Not necessarily since the Motor class passes the pin names through the `register` object to the ShiftRegister class, which handles the aliasing internally. But you are correct that it doesn't follow the new pin naming scheme. 

[That array](https://github.com/rwaldron/johnny-five/blob/master/lib/motor.js#L16) could be updated with our SPI naming, i.e. `["clk", "mosi", "latch"]`, but then we would need to alias the options passed to the Motor class, repeating the logic we have in the Options class. This makes me think we could pull that logic into a separate class/module for re-use. Do you think that would be reasonable for this build?

rwaldron:
> This makes me think we could pull that logic into a separate class/module for re-use. Do you think that would be reasonable for this build?

I'd rather not, because these should be applied to all uses, all the time.

rwaldron:
Thinking more about this and I just read http://www.appelsiini.net/2012/driving-595-shift-registers and now I'm not sure that we sould apply this all the time.

PR #1095, opened 2016-04-11
I think it is cool to add more event and show more value

pacozaa:
I add more event to example and more value to show in change event.
And correct wrong comment above change event which is not involve to speed and course.

### DIFF
--- a/docs/gps.md
+++ b/docs/gps.md
@@ -47,11 +47,19 @@ board.on("ready", function() {
     }
   });
 
-  // If latitude, longitude, course or speed change log it
+  // If latitude, longitude change log it
   gps.on("change", function() {
     console.log("position");
     console.log("  latitude   : ", this.latitude);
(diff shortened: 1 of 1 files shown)


dtex:
Hi @pacozaa ,

I think this is a fine change, but the docs files are generated from the eg files, so you'll want to edit those. Here is a little more info on how that works:

https://github.com/rwaldron/johnny-five/blob/master/CONTRIBUTING.md#writing-documentation

pacozaa:
@dtex 

1. Ok. Thank you very much.
2. If I want to edit [this gps's api doc](http://johnny-five.io/api/gps/) to mention about using GPS class with Arduino MEGA 2560. Where should I take a look at it?

dtex:
Those are generated from this repo's WIKI https://github.com/rwaldron/johnny-five/wiki/GPS

PR #1103, opened 2016-04-15
Thermo-humidity sensor SHT31D

burgo4:
Added support for SHT31D sensor (thermo and humidity).


rwaldron:
The fritzing diagram should show the breakout attached to an Arduino UNO, similar to this: http://johnny-five.io/examples/multi-htu21d/ (don't worry about resizing, just commit it as-is from Fritzing)

rwaldron:
Actually, don't worry about that—I'm going to land this so I can figure out why the examples generator is failing.

GREAT WORK!!! <3

rwaldron:
Ah! That was easy, I just needed to run `grunt examples`

Issue #1113, opened 2016-04-27
array-includes-shim.js considered harmful


mwittig:
This shim adds the "includes" method to the Array.prototype (used if Array.includes is undefined) as an enumerable property. I think this is bad behaviour as sloppy code in other packages might be affected. For example, if one enumerates keys of an array (not a good idea, I know, but a common mistake) and forgets to check for hasOwnProperty() this may be troublesome. Thus, I think it is better to introduce the method as a non-enumerateble property (or use a third-party shim like [array-includes](https://www.npmjs.com/package/array-includes)  if you prefer). I'll come up with a PR shortly.

rwaldron:
That was admittedly a sloppy, hacked-in-spite, response to the size of the `array-incudes` module. 

>  Thus, I think it is better to introduce the method as a non-enumerateble property 

I agree. 

> (or use a third-party shim like array-includes if you prefer).

Naw, getting away from this was intentional, but you're very right to call out the sloppy-ness of my replacement.

rwaldron:
Landed! https://github.com/rwaldron/johnny-five/commit/4e1379778ef7ab1b51b71beaf16af9b9ee18ad9e

mwittig:
Thanks for looking into this so quickly

PR #1114, opened 2016-04-27
Suggested fix for #1113

mwittig:
Add "includes" Array.prototype function as a non-enumerable member, https://github.com/rwaldron/johnny-five/issues/1113


rwaldron:
👍 Nice one, thank you

PR #1153, opened 2016-06-10
make gitter badge image valid common mark

ashleygwilliams:
fixes issue where gitter badge is not displaying on npmjs.com package page
spaces in image paths are not valid common mark, replaced with `%20`


rwaldron:
Sorry I missed this :\ after I read your tweet I went in and changed it to the shields.io version and pushed it out. Either way, I really appreciate your time and attention to identifying the problematic piece of markdown :)

PR #1170, opened 2016-07-01
Restore handling of 0 value

dtex:
Was preparing for NBD and found that blink() did not work on Tessel 2 because pin mode was PWM and we weren't handling cases of ```state.value === 0``` for that.

This situation arises in experiment 1 of the guide for J5IK.

### DIFF
--- a/lib/led/led.js
+++ b/lib/led/led.js
@@ -230,11 +230,10 @@ Led.prototype.on = function() {
     }
 
     // ...the last value was 0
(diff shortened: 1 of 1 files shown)


rwaldron:
Ah! That's super strange because I couldn't figure out any test that would go through this path. Excellent catch though! Thanks for addressing this so fast :)

rwaldron:
I'm going to expedite this since I know these will all pass :)

PR #1171, opened 2016-07-05
add custom frequency method for piezo backpack

deathbearbrown:
Doing QA for the new Piezo backpack firmware (yay!)

Here's the fritzing diagram for setting up the backpack:
piezo-test-rig

Here's the firmware for the backpack:
https://github.com/rwaldron/johnny-five/blob/master/firmwares/piezo_i2c_backpack.ino

Test code:

```
var five = require("johnny-five"),
  board = new five.Board();

  var piezo = new five.Piezo({
    controller: "I2C_BACKPACK",
    pin: 3
  });

  piezo.play({
    song: [
      ['D4', 1/4], ['F4', 1/4], ['D4',1/6], ['D4', 1/8], ['G4', 1/4], ['D4', 1/6], ['C4', 1/6],
      ['D4', 1/4], ['A4', 1/4], ['D4',1/6], ['D4', 1/8], ['A#4', 1/4], ['A4', 1/6], ['F4', 1/6],
      ['D4', 1/6], ['A4', 1/6], ['D5', 1/8], ['D4', 1/6], ['C4', 1/6], ['C4', 1/6], ['A3', 1/8], ['E4', 1/4], ['D4', 1/6]
    ],
    tempo: 45
  });
});

```

If successful, you should hear the Beverly Hills Cop theme song and not a high pitched fax machine sound.


deathbearbrown:
I added the backpack usage to the docs. I included the breadboard image in this PR which I'll use to update the wiki
https://github.com/rwaldron/johnny-five/wiki/Piezo

deathbearbrown:
I also added a delay in the firmware so that the notes are more crisp.  I'm going to stop touching this now... maybe. Probably not. :D

firmwares/piezo_i2c_backpack.ino
```diff
@@ -81,18 +81,23 @@ void onReceive(int howMany) {
     Serial.println(ms);
     Serial.println("--------------------");
   #endif
-
+  Serial.println(OUTPUT);
```
rwaldron: Move this into the DEBUG_MODE block

firmwares/piezo_i2c_backpack.ino
```diff
@@ -81,18 +81,23 @@ void onReceive(int howMany) {
...
   pinMode(pin, OUTPUT);
 
   if (command == NO_TONE) {
     noTone(pin);
   }
 
+  // delay between notes
+  int pauseBetweenNotes = ms * 1.3;
+  delay(pauseBetweenNotes);
+  noTone(pin);
```
rwaldron: This should be inside the `if (command == TONE) {` body, otherwise it's unnecessarily called for `NO_TONE` commands as well

lib/piezo.js
```diff
@@ -123,6 +123,12 @@ var Controllers = {
         return this;
       },
     },
+    frequency: {
+      value: function(frequency, duration) {
+        // toTone() calculation is not required when using the backpack
+        return this.tone(frequency, duration);
+      }
+    }
```
rwaldron: Nice—and thanks for the useful comment here :)

test/piezo.js
```diff
@@ -785,6 +785,16 @@ exports["Piezo - I2C Backpack"] = {
     test.done();
   },
 
+  hasCustomFrequencyMethody: function(test) {
+    test.expect(1);
+    // I hope this is enough
```
rwaldron: It's basically perfect.

test/piezo.js
```diff
@@ -785,6 +785,16 @@ exports["Piezo - I2C Backpack"] = {
...
+    var piezoFrequency = this.piezo.frequency;
+    var frequencyPrototype = Piezo.prototype.frequency;
+    test.notEqual(piezoFrequency, frequencyPrototype);
```
rwaldron: Just a nit, and by no means a blocker, but this could be written as 

```js
test.notEqual(this.piezo.frequency, Piezo.prototype.frequency);
```

test/piezo.js
```diff
@@ -796,7 +806,7 @@ exports["Piezo - I2C Backpack"] = {
   },
 
   noTone: function(test) {
-    // test.expect(2);
+    test.expect(3);
```
rwaldron: Whoops! Thanks for fixing that

firmwares/piezo_i2c_backpack.ino
deathbearbrown: oops, actually I didn't mean to keep that

firmwares/piezo_i2c_backpack.ino
deathbearbrown: So I looked at the mario script and my old piezo scripts and to keep the tempo, I think you'd always have it even if there was no tone?

test/piezo.js
```diff
@@ -785,6 +785,16 @@ exports["Piezo - I2C Backpack"] = {
...
 
+  hasCustomFrequencyMethody: function(test) {
+    test.expect(1);
+    // I hope this is enough
+    var piezoFrequency = this.piezo.frequency;
+    var frequencyPrototype = Piezo.prototype.frequency;
+    test.notEqual(piezoFrequency, frequencyPrototype);
+
+    test.done();
+  },
```
rwaldron: If I could suggest an _additional_ test... Make a test that does something like this: 

```js
var tone = this.sandbox.stub(this.piezo, "tone");
this.piezo.frequency(100, 100);
test.deepEqual(tone.lastCall.args, [100, 100]);
```

Which will assert that `frequency(...)` doesn't mess with the value before forwarding to `tone(...)`

test/piezo.js
```diff
@@ -795,8 +803,18 @@ exports["Piezo - I2C Backpack"] = {
...
 
+  frequencyArgsPersistToTone: function(test) {
+    test.expect(1);
+
+    var tone = this.sandbox.stub(this.piezo, "tone");
+    this.piezo.frequency(100, 100);
+    test.deepEqual(tone.lastCall.args, [100, 100]);
+
+    test.done();
+  },
```
rwaldron: 👍

test/piezo.js
deathbearbrown: I added this and made the backpack tone `writable: true` to be able to stub it

deathbearbrown:
So... I'm starting to think with i2c the delay isn't doing anything.  
From the Mario script it works like this:
- loop over notes
  - turn on tone
  - delay is tone-ms * 1.3  (so wait for the duration of the tone  with a little extra room)
  - turn off tone

With the i2cBackpack, if I added a delay and noTone() at the end, no sound comes out of the piezo.

BrianGenisio:
@deathbearbrown I'm going to wire this up when I get the next chance... but I'm confused about the last comment.  Are you saying the delay IS necessary?  Or it IS NOT necessary?

And should I test with the delay (your branch)?  Or without it (`master`)?  

The instructions above tell me to use the `master` backpack firmware, but the delay is in the firmware on your branch.

Sorry for the confusion!

BrianGenisio:
@deathbearbrown and @rwaldron  I just tested this with and without the delay and I can't say that I can tell the difference.  I'm not sure which one is meant to go in, but otherwise, this all worked for me!  :)

rwaldron:
Nice! I should revisit this myself one last time (probably tomorrow) and then we'll get it landed (probably also tomorrow)

saejinpark:
sorry I'm not english
I want piezo buzzer

how in delayMicroseconds?

void buzz(int targetPin, long frequency, long length) {
  long delayValue = 1000000 / frequency / 2; // calculate the delay value between transitions
  //// 1 second's worth of microseconds, divided by the frequency, then split in half since
  //// there are two phases to each cycle
  long numCycles = frequency * length / 1000; // calculate the number of cycles for proper timing
  //// multiply frequency, which is really cycles per second, by the number of seconds to
  //// get the total number of cycles to produce
  for (long i = 0; i < numCycles; i++) { // for the calculated length of time...
    digitalWrite(targetPin, HIGH); // write the buzzer pin high to push out the diaphram
    delayMicroseconds(delayValue); // wait for the calculated delay value
    digitalWrite(targetPin, LOW); // write the buzzer pin low to pull back the diaphram
    delayMicroseconds(delayValue); // wait again or the calculated delay value
  }

Issue #326, opened 2014-04-21
LCD outputting garbage


dazld:
Arduino Leonardo
OSX 10.9.2
FTDI 32bit drivers

LCD1602A

Node 0.10.25

I can run and setup the LCD using the arduino IDE (1.5.6r2), everything appears to be fine. It's a simple circuit, pot for brightness, and 4bit controls wired up to the pins as in the example, 7-12.

When I try to output any text or commands to the LCD after installing firmata and running the J5 script, it either does nothing or just outputs garbage. Swapping back to the arduino IDE and running a sketch there works fine. 

I'm just using the eg/lcd.js example, nothing else for now.

Is this a known issue, am I doing something wrong...?

Looks like:

![screen shot 2014-04-21 at 18 44 57](https://cloud.githubusercontent.com/assets/201036/2756259/5c15edaa-c974-11e3-8c12-65e5b9ecee75.png)

sephallen:
Sorry for bringing up an old issue, but was this ever resolved?
I have the same LCD screen and as with the OP is works fine in the Arduino IDE, but I just get a completely blank display with Johnny Five.
I've tried on an Uno and Mega, but both have the same results.

rwaldron:
@sephallen are you using a parallel LCD (as shown in the image)? I find this is reproducible only when I have the LCD incorrectly wired, so that's the first thing to double check. Can you also paste the exact code that you're using?

sephallen:
Yes, the LCD is almost identical.
The code I'm using with the Arduino IDE is the LiquidCrystal HelloWorld example which works perfectly.
The node code is as follows:

```javascript
var five = require("johnny-five"),
    board = new five.Board();
board.on("ready", function() {
    var lcd = new five.LCD({
        // LCD pin name  RS  EN  DB4 DB5 DB6 DB7
        // Arduino pin # 12  11   5   4  3  2
        pins: [ 12, 11, 5, 4, 3, 2 ],
        rows: 2,
        cols: 16
    });
    lcd.on("ready", function() {
        lcd.clear();
        lcd.setCursor(0,0);
        lcd.print("hello world");
    });
});
```

rwaldron:
> The code I'm using with the Arduino IDE is the LiquidCrystal HelloWorld example which works perfectly.


Sorry for the dumb question: you made sure to upload StandardFirmata to the board before using the Johnny-Five program?

sephallen:
Yes, of course.
[...]

PR #1174, opened 2016-07-10
Improve LCD millisecond delay handling

fivdi:
The LCD sleep function is currently implemented as follows:
```js
function sleep(ms) {
  var start = Date.now();
  while (Date.now() < start + ms) {}
}
```
As discussed [here](https://github.com/rwaldron/johnny-five/issues/326#issuecomment-230125813), the current implementation will often sleep for less than the requested number of milliseconds.

The modification to the sleep function introduced by this PR ensures that the sleep function sleeps for at least the requested number of milliseconds.

### DIFF
--- a/lib/lcd.js
+++ b/lib/lcd.js
@@ -12,7 +12,9 @@ var priv = new Map();
  */
 function sleep(ms) {
   var start = Date.now();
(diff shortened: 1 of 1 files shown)


rwaldron:
Rad :)

PR #1189, opened 2016-07-27
Prevent `Piezo#play` from mutating `song` array

### LINKED ISSUE #1186: `Piezo#play` mutates input array.
The following code will only play a song once:

```js
const tune = {
  song: [['A4', 1], ['B4', 1]],
  tempo: 120
};

board.on('ready', () => {
    var piezo = new five.Piezo(3);
    setInterval(() => piezo.play(tune), 500);
});
```

Problem is that `Piezo#play` uses the input array as a queue.

Takes array [here](https://github.com/rwaldron/johnny-five/blob/642db0416143484655f33011576a23c819e9f45d/lib/piezo.js#L454) without copy. Then it pops off values [here](https://github.com/rwaldron/johnny-five/blob/642db0416143484655f33011576a23c819e9f45d/lib/piezo.js#L468).

Happy to contribute a PR to fix or document this. It's totally unnecessary as a for loop could be used.

rhys-vdw:
Closes #1186

### DIFF
--- a/lib/piezo.js
+++ b/lib/piezo.js
@@ -453,9 +453,10 @@ Piezo.prototype.play = function(tune, callback) {
   var beatDuration = Math.round(60000 / tempo);
   var song = tune.song || [];
   var duration;
+  var nextNoteIndex = 0;
 
   var next = function() {
-    if (song.length === 0) {
+    if (nextNoteIndex === song.length) {
       // No more notes in song:
       // Song is over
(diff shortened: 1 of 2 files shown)


rhys-vdw:
That test failure doesn't appear to be related to this PR. I saw those errors a couple of times locally with Node 5 too.

rwaldron:
Nice and clean.

PR #1194, opened 2016-07-31
fix pin 7->8 in button-options.

haru01:
button-options.fzz and button-options.png:   pin '8'
but source code and doc:  pin '7'

## see also

* docs/breadboard/button-options.png
* docs/breadboard/button-options.fzz


rwaldron:
Excellent!

PR #1225, opened 2016-09-17
Add enable pin for stepper

dtex:
Will update the wiki when this lands.

### DIFF
--- a/lib/stepper.js
+++ b/lib/stepper.js
@@ -147,8 +147,9 @@ function Stepper(opts) {
 
   // Convert an array of pins to the appropriate named pin
   if (Array.isArray(this.pins)) {
-    if (this.pins.length === 2) {
-      // Using an array of 2 pins requres a TYPE
+    var pinArray = this.pins;
+    if (pinArray.length === 2 || pinArray.length === 3) {
+      // Using an array of 2 or 3 pins requres a TYPE
       // to disambiguate DRIVER and TWO_WIRE
       if (!opts.type) {
         throw new Error(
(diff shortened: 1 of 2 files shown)


dtex:
Has a bug. Will fix and resubmit.

lib/stepper.js
```diff
@@ -157,13 +158,17 @@ function Stepper(opts) {
...
     if (opts.type === Stepper.TYPE.DRIVER) {
-      this.pins = {
-        step: this.pins[0],
-        dir: this.pins[1]
-      };
+      this.pins.step = pinArray[0];
+      this.pins.dir = pinArray[1];
     } else {
-      this.pins = new MotorPins(this.pins);
+      this.pins = new MotorPins(pinArray);
```
rwaldron: Blah... `Stepper` is such a mess in the way it handles `this.pins` :(

dtex: Ya, that's what made me close my PR. I missed something in the convolution.

dtex:
Okay, so this branch works fine for the HRC robot so my urgency has abated. If I go ahead and refactor this class I am going to use this [boilerplate for Johnny-Five classes](https://gist.github.com/dtex/1ef97480209efe621cf9). Would you like to suggest any changes before I do?

rwaldron:
I like all of those comments :)


Using `this.opts.anything` has to go (should've never happened in motor.js, which it only appears there iirc).


Change https://gist.github.com/dtex/1ef97480209efe621cf9#file-classboilerplate-js-L91 to: 

```js
  Board.Component.call(
    // opts gets normalized to opts, not `this.opts`
  );
```

Change https://gist.github.com/dtex/1ef97480209efe621cf9#file-classboilerplate-js-L97-L103 to: 

```js
  // If this class has multiple devices we need to figure out which one to use
  if (typeof opts.device === "undefined") {
    // Derive the device type if not explicitly passed
  }

  // Allow users to pass in custom device types
  device = typeof opts.device === "string" ?
    Devices[opts.device] : opts.device;
```

Change https://gist.github.com/dtex/1ef97480209efe621cf9#file-classboilerplate-js-L112-L118 to:


```js
  // Then the controller decorates the instance
  var controller = null;

  if (opts.controller && typeof opts.controller === "string") {
    controller = Controllers[opts.controller.toUpperCase()];
  } else {
    controller = opts.controller;
  }

  if (controller == null) {
    controller = Controllers.ANALOG;
  }

  Board.Controller.call(this, controller, opts);
```


All occurrences of `this.opts.someValue` => `this.someValue`

rwaldron:
So.. this: 

```js
  if (controller == null) {
    controller = Controllers.ANALOG;
  }
```

You should choose an intuitive controller as a default (or if you decide "no default" it needs to throw): 

```js
  if (controller == null) {
    throw new Error("Stepper expects a valid controller");
  }
```

PR #1238, opened 2016-10-14
remove controller alias for HD44780

nullvariable:
HD44780 is not actually a controller but a type of screen


rwaldron:
Thanks!!

PR #1239, opened 2016-10-16
Led.RGB: Fix example "LED - Rainbow"

jsoltysiak:
Loop didn't go through entire rainbow array. The last element was
omitted.


rwaldron:
Nice catch and thanks for the fix

PR #1240, opened 2016-10-19
Update lcd-enumeratechars.md

shimnex:
Small bug in exemple


shimnex:
Does anyone can test the original example? For me, it do not work at all.

rwaldron:
We both spotted this ;) I made a fix earlier today.

PR #1241, opened 2016-10-19
Update lcd.js, bug in useChar function

shimnex:
state.characters[name] = 0 is treaty as undefined, must use better comparason.


shimnex:
So ... will you merge this fix? Do a have to do something else?

Issue #1242, opened 2016-10-23
Support for pull-down resistors on platforms that have them


andrewn:
On platforms such as Raspberry Pi, [some pins have both pull-up and pull-down resistors](http://wiringpi.com/pins/special-pin-functions/) but currently there's no way to enable pull-down resistors using Johnny-Five.

Enabling a pull-up resistor works by writing `HIGH` to the pin in `INPUT` mode. I'm proposing that writing `LOW` should enable the built-in pull-down resistor.

I've created an [issue](https://github.com/nebrius/raspi-io/issues/71) and [pull request](https://github.com/nebrius/raspi-io/issues/72) to support this behaviour in [Raspi-IO](https://github.com/nebrius/raspi-io).

I'm going to open a pull request with a patch enabling this but I think an issue is useful for discussing if this is the right approach and something you'd support.

revolunet:
looks like its been merged 👍

fivdi:
Now that #1243 has been merged I think this issue can be closed.

rwaldron:
@fivdi thanks for the follow up

PR #1243, opened 2016-10-23
Button pull-down support

andrewn:
As suggested in #1242 this adds support to the `Button` for enabling a pull-down resistor when `LOW` is written to an input pin on platforms that have them like the Raspberry Pi.

The corresponding [pull request](https://github.com/nebrius/raspi-io/issues/72) adds this behaviour to [Raspi-IO](https://github.com/nebrius/raspi-io).


rwaldron:
This is supposed to be the default behavior of inputs, but I'm not opposed to allowing a mechanism that gives explicit control. We'll need to add a test—let me know if you'd like help with that

andrewn:
Ah great, I didn't know that was default behaviour. On the Pi it needs to be explicitly enabled on a pin.

I'm happy to write a test but I can't quite figure out what it should be and where it should go so pointers would be useful please!

rwaldron:
No worries, I don't mind writing the basic tests—that way you can see an example of what needs to be written for next time :)

rwaldron:
Thanks!!

andrewn:
No, thank you! 😄

PR #1244, opened 2016-10-23
Use more then 8 predefined characters

shimnex:
Actually, only the first 8 predefined characters will be load. After 8 "useChar", it is not possible to change/replace character in LCD memory.

This fix allow to use the 8 LAST predefined characters defined with "useChar".


shimnex:
Well, I thing this version is good. What do you thing.

rwaldron:
Sorry, I made a git command mistake locally and didn't mean to merge that. I still need to review

rwaldron:
This is great, a nice memory leak fix!

PR #1256, opened 2016-12-02
Prevent HTU21D from freezing node

bbx10:
https://github.com/sandeepmistry/node-chip-io/issues/28


fivdi:
@bbx10 now that chip-io uses i2c-bus for I2C device access this fix is no longer needed.

rwaldron:
> now that chip-io uses i2c-bus for I2C device access this fix is no longer needed.

Thanks @fivdi <3

PR #1261, opened 2016-12-13
Relay: Fix typo

scottgonzalez:


rwaldron:
Derp. Thanks for catching and correcting that ❤️

PR #1270, opened 2017-01-09
Updating the copyright to 2017

BrianGenisio:
Running `grunt examples` causes ALL docs to update because of the copyright.  I'm sending this as a separate PR so it doesn't come with other feature-level PRs.

### DIFF
--- a/docs/accelerometer-LIS3DH.md
+++ b/docs/accelerometer-LIS3DH.md
@@ -77,7 +77,7 @@ board.on("ready", function() {
 ## License
 Copyright (c) 2012, 2013, 2014 Rick Waldron <waldron.rick@gmail.com>
 Licensed under the MIT license.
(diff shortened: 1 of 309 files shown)


rwaldron:
I think these need to be "2016, 2017", otherwise 2016 is just unaccounted for

BrianGenisio:
This is really just an update to [tpl/.eg.md](tpl/.eg.md) which currently doesn't consider the year 2015.  It covers "2012, 2013, 2014 Rick Waldron" and then "<%= theYear %> The Johnny-Five Contributors".  How would you like it to read?

rwaldron:
After revisiting a discussion with "my" (he's my next door neighbor) pro-bono IP lawyer, all we need is the year that the content was created. So adjusting tpl/.eg.md is right because it should reflect the year that the file is regenerated. Sorry for the confusion.

PR #1284, opened 2017-01-30
Fixed scaling factors

CzechUavGuy:
Should be scaled by 2^10, not by 2^16. See datasheet https://cdn-shop.adafruit.com/datasheets/TSL2561.pdf page 24

### DIFF
--- a/lib/light.js
+++ b/lib/light.js
@@ -327,8 +327,8 @@ var Controllers = {
 
               // Page 27
               // CalculateLux(...)
-              ch0 = (ch0 * scale) >> 0x10;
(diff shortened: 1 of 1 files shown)


lyzadanger:
As a second eye here, I do see that `CH_SCALE` is set to `10` on p24 of datasheet:

image

And subsequent arithmetic, page 27:

image

This looks good to _me_ but I'd like to get someone who knows this area of the codebase better than I to take a peek...

rwaldron:
That looks like I just made a typing mistake. This fix is right. Will land when I return from parental leave in a week and a half, or someone else can merge before.

PR #1285, opened 2017-01-30
Pcf8591 analog out

CzechUavGuy:
Added possibility for PCF8591 to output analog value. Needs code review - see comments in code.

### DIFF
--- a/lib/expander.js
+++ b/lib/expander.js
@@ -728,7 +728,7 @@ var Controllers = {
     }
   },
   // http://www.nxp.com/documents/data_sheet/PCF8591.pdf
-  PCF8591: {
+PCF8591: {
     ADDRESSES: {
       value: [0x48]
     },
@@ -758,8 +758,17 @@ var Controllers = {
             analogChannel: i
           });
         }
+	this.pins.push({
+	  supportedModes: [
+	    this.MODES.PWM
+	  ],
+	  mode: 0,	//no idea about this line
+	  value: 0,	//no idea about this line
+	  report: 0,	//no idea about this line
+	  analogChannel: 4
+	});
 
-        this.analogPins.push(0, 1, 2, 3);
+        this.analogPins.push(0, 1, 2, 3, 4);
 
         this.io.i2cWrite(this.address, state.control);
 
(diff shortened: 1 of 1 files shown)


rwaldron:
LGTM, but needs tests. I might just write the test myself to ease back into work (I've been away for 8 weeks). Let me know if you want to write the tests or if I should. Here's where they'd go: https://github.com/rwaldron/johnny-five/blob/master/test/expander.js#L1982-L2156

fivdi:
Is this really a good idea? The [PCF8591](https://www.nxp.com/documents/data_sheet/PCF8591.pdf) has a D/A converter that converts digital values to their corresponding analog voltage. The analog voltage can be more or less any voltage. It's not the same as PWM where the output voltage is either 0V or VDD where VDD is typically 5V.

Although the technique used by this PR will almost certainly work, reusing the PWM mode for DAC will likely result in confusion. For example, it's easy to set the brightness of an LED using PWM. Setting the brightness of an LED using DAC is more complex. If the PWM mode is reused for DAC, Johnny-Five will think that the pin supports PWM, but it doesn't.

soundanalogous:
I'm planning to add DAC support to Firmata (at the protocol level) at some point: https://github.com/firmata/protocol/issues/32.

fivdi:
@soundanalogous That sounds like a better way to go to me.

soundanalogous:
This brings up another interesting question however which is how do we handle pin numbering with expander ICs?

CzechUavGuy:
OK lets wait for soundanalogous for proper implementation. Then I can use his framework to rewrite my patch.

rwaldron:
@fivdi thanks for intervening with that info

@soundanalogous re: pin numbering on ICs, if I understand your question correctly (which I might not), I generally stick to "as seen on the pcb"


I agree that once a "dacWrite" (or whatever) is designed, then this can follow.

dtex:
The fix will have to wait on DAC support in firmata so I'm closing this PR. I've added the feature request to the [Requests Features](https://github.com/rwaldron/johnny-five/wiki/Requested-Features) wiki page and linked back to this PR.

Issue #1298, opened 2017-03-05
Max. raw value for an adc assumed to be 1024 but it's 1023


fivdi:
When the LM35, LM335 and TMP36 thermometers calculate the mV output of a device [here](https://github.com/rwaldron/johnny-five/blob/v0.10.4/lib/thermometer.js#L311), [here](https://github.com/rwaldron/johnny-five/blob/v0.10.4/lib/thermometer.js#L330) and [here](https://github.com/rwaldron/johnny-five/blob/v0.10.4/lib/thermometer.js#L346), the calculation assumes that the maximum raw value for an analog to digital conversion is 1024 but it's 1023.

This is what the code currently looks like in all three cases:
```js
var mV = this.aref * 1000 * raw / 1024;
```
Assuming that `this.aref` is 5 and `raw` is 1023 (which is the maximum value of an adc) this code will set `mV` to 4995.1171875 which is incorrect, `mV` should be set to 5000.

This can be fixed by replacing 1024 with 1023:
```js
var mV = this.aref * 1000 * raw / 1023;
```

fivdi:
https://github.com/rwaldron/johnny-five/pull/1299 is a PR for fixing this issue.

PR #1299, opened 2017-03-05
Thermometer: LM35/LM335/TMP36 max raw value 1023

fivdi:
Fixes https://github.com/rwaldron/johnny-five/issues/1298

### DIFF
--- a/lib/thermometer.js
+++ b/lib/thermometer.js
@@ -308,7 +308,7 @@ var Controllers = {
         // VOUT = 250 mV at 25°C
         // VOUT = –550 mV at –55°C
 
-        var mV = this.aref * 1000 * raw / 1024;
+        var mV = this.aref * 1000 * raw / 1023;
 
         // 10mV = 1°C
         //
@@ -327,7 +327,7 @@ var Controllers = {
       value: function(raw) {
         // OUTPUT 10mV/°K
 
-        var mV = this.aref * 1000 * raw / 1024;
+        var mV = this.aref * 1000 * raw / 1023;
 
         // Page 1
         return Math.round((mV / 10) - CELSIUS_TO_KELVIN);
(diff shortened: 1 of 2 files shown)


fivdi:
The test that's failing here is related to animations on OSX with Node.js v0.10:
```
Testing animation.js.......F..............................
>> Animation -- Servo - keyframeEasing
>> Error: 0 == true
>> at Object.exports.Animation -- Servo.keyframeEasing (test/animation.js:263:10)
>> at Object.exports.Animation -- Servo.setUp (test/animation.js:97:5)
```
The thermometer tests are fine:
```
Testing thermometer.js................................................................................................OK
```
Note that it wasn't necessary to modify the thermometer tests as they all passed after changing the max. raw value of an adc conversion from 1024 to 1023. This is a relatively small change in the max. raw value and only results in a relatively small change in the temperature reading. The tests use [math.round](https://github.com/rwaldron/johnny-five/blob/v0.10.4/test/thermometer.js#L46-L81) which means small changes in temperature are not relevant for the tests.

lib/thermometer.js
```diff
@@ -343,7 +343,7 @@ var Controllers = {
     toCelsius: {
       value: function(raw) {
         // Analog Reference Voltage
-        var mV = this.aref * 1000 * raw / 1024;
```
rwaldron: lol. WHOOPS!

rwaldron:
How do you feel about writing some additional tests to cover this against future regression? wink-wink, nudge-nudge

fivdi:
Ok, I'll take a look.

fivdi:
> How do you feel about writing some additional tests to cover this against future regression?

Regression tests added with 0ffe5fc8fd502e060704248e44df4a8ce036f74f

fivdi:
The AppVeyor build times out after 60 minutes but I have no idea why. The timeout doesn't appear to be related to this PR.

rwaldron:
Looks like bogus fails? It's passing fine now https://ci.appveyor.com/project/rwaldron/johnny-five/build/756

rwaldron:
👏

PR #1302, opened 2017-03-10
Add support for BH1750 light sensor

CzechUavGuy:
This adds support for BH1750 light sensor, see datasheet info at http://rohmfs.rohm.com/en/products/databook/datasheet/ic/sensor/light/bh1750fvi-e.pdf

### DIFF
--- a/lib/light.js
+++ b/lib/light.js
@@ -394,6 +394,40 @@ var Controllers = {
       }
     }
   },
+  BH1750: {
+    //http://rohmfs.rohm.com/en/products/databook/datasheet/ic/sensor/light/bh1750fvi-e.pdf
+    //code based on Arduino library https://github.com/claws/BH1750
+    //currently only "continuous H-resolution" mode supported
+    ADDRESSES: {
+      value: [0x23, 0x5C]
+    },
+    initialize: {
+      value: function(opts, dataHandler) {
+        var address = opts.address || 0x23;
+        var mode = opts.mode || 0x10;
+        opts.address = address;
+        this.io.i2cConfig(opts);
+        this.io.i2cWrite(address, mode);
+        var read = function() {
+          setTimeout(function() {
+            this.io.i2cReadOnce(address, 2, function(data) {
+              var raw = data[0];
+              raw <<= 8;
+              raw |= data[1];
+              dataHandler(raw);
+              read();
+            });
+          }.bind(this), 120);
+        }.bind(this);
+        read();
+      }
+    },
+    toIntensityLevel: {
+      value: function(raw) {
+        return toFixed(raw / 1.2, 2);
+      }
+    }
+  },
 };
 
 Controllers.ALSPT19 = Controllers["ALS-PT19"] = Controllers.DEFAULT;


lib/light.js
```diff
@@ -394,6 +394,40 @@ var Controllers = {
...
+        }.bind(this);
+        read();
+      }
+    },
+    toIntensityLevel: {
+      value: function(raw) {
+        return toFixed(raw / 1.2, 2);
+      }
+    }
+  },
```
rwaldron: This is clean—I love it

rwaldron:
LGTM, but needs some tests. Let me know if you're up for it, otherwise I can write them :)

CzechUavGuy:
I'm a little busy right now, but when I'll get back to my Firmata project I'll do them.

rwaldron:
I can also help write them, just to help with momentum :) I need to order one of these

CzechUavGuy:
OK go ahead. Thanks!

rwaldron:
Quick note: I get "access denied" when trying to open that url for the datasheet. I will update the link in the comments to an accessible URL.

rwaldron:
The BH1750 I ordered isn't being detected for some reason, so minor delay until tomorrow at least.

pataltechnologies:
@rwaldron I'm trying to use this, and I'm having trouble getting things to read. Is it possible you could open another PR and add an example/documentation markdown file to the `/docs` folder?

rwaldron:
I'll try to get that done this week

rwaldron:
@pataltechnologies I've added the example: https://github.com/rwaldron/johnny-five/blob/master/docs/light-ambient-BH1750.md

pataltechnologies:
@rwaldron Woah nice. You took it one step further with the diagrams. Thank you for doing this.

jhoughjr:
The example does not work on a raspberry pi. Simply says invalid args in j5-io/index/js line 477.
Been googling all day to figure it out.

dtex:
This error is specific to raspi-io, but I'll see if I can help. The example is for an Arduino. Can you share your code? 

p.s. One of the reason's it's hard to Google is that raspi-io has very recently undergone a significant re-write so this error message might be relatively new.

cc @nebrius

nebrius:
@jhoughjr yep, you're literally the first person to ask this! (the rewrite was launched less than a month ago).

Can you copy-paste the full error along with the stack trace?

nebrius:
Doing a bit of digging, it seems that J5 is calling Raspi IO with some unexpected arguments.

Here's the signature for the method throwing the exception (written in TypeScript):

```typescript
  public i2cWrite(address: number, inBytes: number[]): void;
  public i2cWrite(address: number, register: number, inBytes: number[]): void;
```

Notice how there are two ways of calling them (with a register or without one). This is the only variant though, and when I did the rewrite I tightened up the type checking (go figure) so that it throws the exception you get if this method is called in a way I wasn't expecting.

My _guess_ is that J5 is calling this method with a third signature type that isn't documented (maybe a buffer instead of array of numbers?), and my rewrite exposed this undocumented method signature. @dtex, does this ring any bells?

dtex:
I'm only aware of the two. I did some digging of my own in the light class and I see [several places](https://github.com/rwaldron/johnny-five/blob/master/lib/light.js#L409) where the inBytes param is being passed as a number instead of an array. Looking at firmata.js [I don't think that case is handled](https://github.com/firmata/firmata.js/blob/master/packages/firmata-io/lib/firmata.js#L1239-L1246) at all. I am confused.

Issue #1291, opened 2017-02-15
If two 74HC595 shift registers are used they need unique addresses


fivdi:
This issue was reported on the J5 gitter channel. 

If two 74HC595 shift registers are used like this:

```js
  var expander_1 = new five.Expander({
    controller: "74HC595",
    pins: {
      data: 6,
      clock: 7,
      latch: 8
    }
  });

  var expander_2 = new five.Expander({
    controller: "74HC595",
    pins: {
      data: 9,
      clock: 10,
      latch: 11
    }
  });

  var virtual_1 = new five.Board.Virtual(expander_1);
  var virtual_2 = new five.Board.Virtual(expander_2);

  var ledsGroup_1 = new five.Leds({
    pins: [0, 1, 2, 3, 4, 5, 6, 7],
    board: virtual_1
  });

  var ledsGroup_2 = new five.Leds({
    pins: [0, 1, 2, 3, 4, 5, 6, 7],
    board: virtual_2
  });
```

The following error will be thrown:

<img src="https://files.gitter.im/rwaldron/johnny-five/psfR/blob">

Although a 74HC595 shift register isn't an I2C device, the Expander class treat it as one. This only becomes visible when two or more 74HC595 shift registers are used as they all end up having the same I2C address. It's possible to work around the issue by giving each shift register a unique address:

```js
var expander_1 = new five.Expander({
    controller: "74HC595",
    pins: {
      data: 6,
      clock: 7,
      latch: 8
    },
    address: 0x01
  });

var expander_2 = new five.Expander({
    controller: "74HC595",
    pins: {
      data: 9,
      clock: 10,
      latch: 11
    },
    address: 0x02
  });
```

Perhaps this is only a documentation issue.

rwaldron:
Yikes, that's a totally real bug. Expander controller definitions will need to include some flag that indicates that they are I2C and therefore subject to I2C address constraints, or "normalization" step that automatically adds an address for those that _shouldn't_ require an address from the user. I'll add this to my more immediate "to do" list

PR #1312, opened 2017-03-30
Expander: don't require an address when unnecessary. Fixes gh-1291

rwaldron:
@fivdi can you try your code with this branch and without your explicit addresses? Thanks!

Signed-off-by: Rick Waldron <waldron.rick@gmail.com>

### DIFF
--- a/lib/expander.js
+++ b/lib/expander.js
@@ -1812,6 +1812,10 @@ Object.keys(Controllers).forEach(function(name) {
   });
 });
 
+var nonAddressable = [
+  "74HC595"
+];
+
 function Expander(opts) {
   if (!(this instanceof Expander)) {
     return new Expander(opts);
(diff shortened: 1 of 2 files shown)


fivdi:
@rwaldron I gave this branch a try and the Error is no longer thrown. However, I tried without having any 74HC595 shift registers connected as I don't have any. The issue was originally reported by @kulagowskim on the gitter channel.

@kulagowskim The [issue](https://github.com/rwaldron/johnny-five/issues/1291) that you originally reported on the Johnny-Five gitter channel has been fixed. Can you try your code with this branch and without the explicit addresses that were added to workaround the issue? The branch can be installed with `npm install "rwaldron/johnny-five#1291"`.

rwaldron:
Thanks @fivdi, we'll wait to see what @kulagowskim reports. In the meantime, I will look at the these failures :|

PR #1329, opened 2017-04-20
Make unit tests less flaky on Travis CI

islemaster:
I believe I've identified and fixed the two top causes of flaky failures in the Travis CI builds. I've provided inline comments for each fix.

### Existing failure rate
For a baseline, I looked at the last five builds on `master`: [2574](https://travis-ci.org/rwaldron/johnny-five/builds/222835946), [2573](https://travis-ci.org/rwaldron/johnny-five/builds/222109930), [2569](https://travis-ci.org/rwaldron/johnny-five/builds/221795739), [2568](https://travis-ci.org/rwaldron/johnny-five/builds/221793037), and [2567](https://travis-ci.org/rwaldron/johnny-five/builds/221778389).

- Total jobs: 50 (10 per build)
- Failures: 4 jobs / 8%
  - `Animation -- Servo - keyframeEasing`
    - [2574.6](https://travis-ci.org/rwaldron/johnny-five/jobs/222835952)
  - `Cannot read property 'setUp' of undefined`
    - [2574.10](https://travis-ci.org/rwaldron/johnny-five/jobs/222835956)
    - [2568.9](https://travis-ci.org/rwaldron/johnny-five/jobs/221793046)
    - [2567.9](https://travis-ci.org/rwaldron/johnny-five/jobs/221778398)

I've seen both of these on my own pull requests recently too - they seem to be fairly common.

### New failure rate
Considering builds: [2578 (PR)](https://travis-ci.org/rwaldron/johnny-five/builds/224164000), [2579 (PR)](https://travis-ci.org/rwaldron/johnny-five/builds/224169626), [2580 (PR)](https://travis-ci.org/rwaldron/johnny-five/builds/224170521), [41](https://travis-ci.org/code-dot-org/johnny-five/builds/224162626), [42](https://travis-ci.org/code-dot-org/johnny-five/builds/224169608), and [43](https://travis-ci.org/code-dot-org/johnny-five/builds/224170493).

- Total jobs: 60 (10 per build)
- Failures: None!

### DIFF
--- a/test/animation.js
+++ b/test/animation.js
@@ -243,7 +243,10 @@ exports["Animation -- Servo"] = {
   },
 
   keyframeEasing: function(test) {
-
+    // Don't allow time to advance in this test.  Otherwise it's possible for
+    // the animation to progress between setup and the assertion, causing flaky
+    // failures.
+    this.sandbox.useFakeTimers();
     this.animation = new Animation(this.a);
     test.expect(1);
 
@@ -260,7 +263,8 @@ exports["Animation -- Servo"] = {
     var indices = this.animation.findIndices(progress);
     var val = this.animation.tweenedValue(indices, progress);
 
-    test.ok(Math.abs(val - 74.843 < 0.01));
+    test.ok(Math.abs(val - 74.843) < 0.01,
+      "Expected " + val + " to be within 0.01 of 74.843");
     test.done();
   },
 
--- a/test/repl.js
+++ b/test/repl.js
@@ -44,9 +44,16 @@ exports["Repl"] = {
       debug: false
     });
 
+    // Extra-careful guard against calling test.done() twice here.
+    // This was causing "Cannot read property 'setUp' of undefined" errors
+    // See https://github.com/caolan/nodeunit/issues/234
+    var calledTestDone = false;
     var reallyExit = this.sandbox.stub(process, "reallyExit", function() {
       reallyExit.restore();
-      test.done();
+      if (!calledTestDone) {
+        calledTestDone = true;
+        test.done();
+      }
     });
 
     board.on("ready", function() {


test/animation.js
```diff
@@ -243,7 +243,10 @@ exports["Animation -- Servo"] = {
   },
 
   keyframeEasing: function(test) {
-
+    // Don't allow time to advance in this test.  Otherwise it's possible for
+    // the animation to progress between setup and the assertion, causing flaky
+    // failures.
+    this.sandbox.useFakeTimers();
```
islemaster: On a separate branch I added a bunch of logging in this test, and I found that when it fails, the animation's `progress` value has advanced slightly from 0.9 between calling `this.animation.enqueue(segment)` and capturing the return value of `this.animation.calculateProgress(this.animation.playLoop.calledAt)`.  [For example, this run reports progress of 0.902](https://travis-ci.org/code-dot-org/johnny-five/jobs/224152325#L1371).

I eventually realized that this test assumes no significant time will pass in that space - that's why we're manually starting the progress at 0.9 - but on very slow machines that might not be true.  The solution is to use sinon's fake clocks to effectively stop time during this test.  The fake timers are implicitly cleaned up by the sandbox when the test is done.

test/animation.js
```diff
@@ -260,7 +263,8 @@ exports["Animation -- Servo"] = {
     var indices = this.animation.findIndices(progress);
     var val = this.animation.tweenedValue(indices, progress);
 
-    test.ok(Math.abs(val - 74.843 < 0.01));
+    test.ok(Math.abs(val - 74.843) < 0.01,
+      "Expected " + val + " to be within 0.01 of 74.843");
```
islemaster: I'm surprised that the original assertion here ever worked, due to the misplaced paren; turns out
```javascript
(Math.abs(true) === 1) == true;
(Math.abs(false) === 0) == false;
```
I figured it wouldn't hurt to leave a nicer assertion behind. :grin:

test/repl.js
```diff
@@ -44,9 +44,16 @@ exports["Repl"] = {
...
+    // This was causing "Cannot read property 'setUp' of undefined" errors
+    // See https://github.com/caolan/nodeunit/issues/234
+    var calledTestDone = false;
     var reallyExit = this.sandbox.stub(process, "reallyExit", function() {
       reallyExit.restore();
-      test.done();
+      if (!calledTestDone) {
+        calledTestDone = true;
+        test.done();
+      }
```
islemaster: This was a little harder to track down, but my first clue was [this issue](https://github.com/caolan/nodeunit/issues/198) on the nodeunit repo, followed by [this issue](https://github.com/caolan/nodeunit/issues/234) that is still open and summarizes the problem well; nodeunit raises this really opaque error if you happen to call `test.done()` twice in a single test.

I scoured rgb.collection.js for places where this might happen but eventually noticed that it always failed _right_ at the top of that file... So I checked the test file before it, and sure enough, there's some tricky and system-dependent callback stuff going on here.  I did some more extra logging on a separate branch and discovered that the `"reallyExit"` stub was getting called twice in failure cases (examples [here](https://travis-ci.org/code-dot-org/johnny-five/jobs/224146923#L1135) and [here](https://travis-ci.org/code-dot-org/johnny-five/jobs/224141898#L1134)).  I'm not sure how - that doesn't seem possible, since we restore the stub when it's called - but I figured a second guard against calling `test.done()` twice might be in order.

If anyone has a better idea of how the `reallyExit` stub is getting called a second time, I'm all ears.  Maybe something in the previous test?

rwaldron:
👏  Nice!

dtex:
Thank you so much @islemaster !!!

There are not enough emojis and smileys in the world to convey how happy (and relieved) this makes me.

PR #1336, opened 2017-05-03
Fix tests on Node v0.10.x

islemaster:
We [discussed here](https://github.com/rwaldron/johnny-five/commit/be3e8bd6211c6c1c1c25613d2bb7cc487466a581#commitcomment-21991716) a safe check for whether `Symbol` is available in the current node version (it's not, in node v0.10).  Three cases [were fixed here](https://github.com/rwaldron/johnny-five/commit/2884c2e5120b6eb1b8e91de806b238825ff68ba1) but a fourth was missed, causing tests to continue failing on node v0.10.x.  This fixes that fourth case.

### DIFF
--- a/test/collections.js
+++ b/test/collections.js
@@ -332,7 +332,7 @@ exports["Collection.Emitter"] = {
   },
 
   "Symbol.iterator": function(test) {
-    if (Symbol && Symbol.iterator) {
+    if (typeof Symbol !== "undefined" && Symbol.iterator) {
       test.expect(1);
       test.equal(Collection.prototype[Symbol.iterator], Array.prototype[Symbol.iterator]);
     }


rwaldron:
That's what I get for rushing. Thanks for following up on this.

rwaldron:
@islemaster q: are you relying on these older versions of Node, or just being diligent? I'm asking because we're no longer held back by BeagleBone-IO's dependencies (previous version was depending on bonescript, which was stuck with node-i2c, which appears abandoned)

islemaster:
Oh! I'm just being diligent - I'd like to keep our fork in lockstep with
upstream at least until release, but will skip versions with failing tests.

Since we're targeting browsers my constraint is what Babel can transpile,
and node is just a build tool. I believe my team is all on node 6 now. No
objection here if you're ready to drop support for node 0.10.

PR #1338, opened 2017-05-07
LCD: Wait for fast instructions to complete

### LINKED ISSUE #1295: An LCD displays garbage if an IO plugin is fast at digital output
A parallel LCD displays garbage if the IO plugin being used can perform fast digital output.

The execution times for the various LCD instructions are specified on pages 24 and 25 of the [HD44780 datasheet](https://www.sparkfun.com/datasheets/LCD/HD44780.pdf). If the execution time for an instruction is a small number of milliseconds, Johnny-Five takes this time into account. An example of such an instruction is the "return home" instruction. As can be seen in the code [here](https://github.com/rwaldron/johnny-five/blob/v0.10.4/lib/lcd.js#L843-L847), there is a 2 millisecond delay after sending the home instruction to the LCD. Waiting for 2 milliseconds gives the LCD time to complete the operation before the next instruction is sent to the LCD.

However, most instructions have an execution time of 37 microseconds that Johnny-Five doesn't take into account. While this may not have been an issue for most IO plugins in the past, it appears to be an issue for [pi-io](https://github.com/fivdi/pi-io). pi-io is fairly fast at digital output and can perform almost 2,000,000 digital writes per second which means that a single digital write can be performed in approximately 500 nanoseconds. Because Johnny-Five isn't waiting for the instructions with an execution time of 37 microseconds to complete, it's relatively easy for pi-io to begin executing an instruction before the previous instruction has completed.

The issue can be reproduced with [lcd-stress-test.js](https://github.com/fivdi/pi-io/blob/v0.1.1/test/lcd-stress-test.js) on a Raspberry Pi 3. When run, this program will display garbage on the LCD immediately.

It looks like the issue can be resolved by modifying [this code](https://github.com/rwaldron/johnny-five/blob/v0.10.4/lib/lcd.js#L755-L775) from:
```js
LCD.prototype.send = function(value) {
  var pin = 0;
  var mask = {
    4: 8,
    8: 128
  }[this.bitMode];

  for (; mask > 0; mask = mask >> 1) {
    this.io.digitalWrite(
      this.pins.data[pin],
      this.io[value & mask ? "HIGH" : "LOW"]
    );
    pin++;
  }

  ["LOW", "HIGH", "LOW"].forEach(function(val) {
    this.io.digitalWrite(this.pins.en, this.io[val]);
  }, this);

  return this;
};
```
to:
```js
// Waits for the specified number of microseconds.
function sleepus(usDelay) {
  var startTime = process.hrtime(),
    deltaTime,
    usWaited = 0;

  while (usDelay > usWaited) {
    deltaTime = process.hrtime(startTime);
    usWaited = (deltaTime[0] * 1E9 + deltaTime[1]) / 1000;
  }
}

LCD.prototype.send = function(value) {
  var pin = 0;
  var mask = {
    4: 8,
    8: 128
  }[this.bitMode];

  for (; mask > 0; mask = mask >> 1) {
    this.io.digitalWrite(
      this.pins.data[pin],
      this.io[value & mask ? "HIGH" : "LOW"]
    );
    pin++;
  }

  ["LOW", "HIGH", "LOW"].forEach(function(val) {
    this.io.digitalWrite(this.pins.en, this.io[val]);
  }, this);

  sleepus(37); // <<<<<<<<<<========== NEW NEW NEW

  return this;
};
```
Note that the following snippet of code from above is also likely to become critical at some stage:
```js
  ["LOW", "HIGH", "LOW"].forEach(function(val) {
    this.io.digitalWrite(this.pins.en, this.io[val]);
  }, this);
```
This code generates a short high level pulse on the enable pin. If pi-io is being used, the high level pulse width will be approximately 500 nanoseconds. At VCC = 3.3V, the minimum enable pulse width is specified as 450 nanoseconds on page 49 of  the [HD44780 datasheet](https://www.sparkfun.com/datasheets/LCD/HD44780.pdf).

fivdi:
Fixes #1295

### DIFF
--- a/lib/lcd.js
+++ b/lib/lcd.js
@@ -10,11 +10,24 @@ var priv = new Map();
  * If any other approach can be found, patches
  * will gratefully be accepted.
  */
+function sleepus(usDelay) {
+  var startTime = process.hrtime();
+  var deltaTime;
+  var usWaited = 0;
+
+  while (usDelay > usWaited) {
+    deltaTime = process.hrtime(startTime);
+    usWaited = (deltaTime[0] * 1E9 + deltaTime[1]) / 1000;
+  }
+}
+
+/**
+ * This atrocity is unfortunately necessary.
+ * If any other approach can be found, patches
+ * will gratefully be accepted.
+ */
 function sleep(ms) {
-  var start = Date.now();
-  // Wait 1ms longer than requested. For further information see:
-  // https://github.com/rwaldron/johnny-five/issues/326#issuecomment-230125813
-  while (Date.now() < start + ms + 1) {}
+  sleepus(ms * 1000);
 }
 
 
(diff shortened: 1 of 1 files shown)


lib/lcd.js
```diff
@@ -809,9 +822,23 @@ LCD.prototype.send = function(value) {
...
+  sleepus(1);
+  this.io.digitalWrite(this.pins.en, this.io.LOW);
+
+  // The execution time for the vast majority of instructions is at least
+  // 37 microseconds. See datasheet pages 24 and 25.
+  // https://www.sparkfun.com/datasheets/LCD/HD44780.pdf
+  // It's important to wait 37 microseconds here to prevent fast IO plugins
+  // like Pi-IO from executing the next instruction before the current
+  // instruction has completed.
+  sleepus(37);
```
rwaldron: I just need to test this with a mountain of LCDs 😩

fivdi: If it's any help, I tested with a 2004 LCD on a Pi 3 and a 1602 LCD on an Arduino UNO and there were no problems. I'd say the `sleepus(1)` and `sleepus(37)` calls are low risk modifications. The modification to `sleep` is potentially riskier. `sleep` used to wait 1 millisecond longer than asked but now that the high resolution timer is being used it no longer waits 1 millisecond longer than asked.

rwaldron: Ok, thanks for the follow up.

Issue #1297, opened 2017-03-02
five.Thermometer sets an unclearable interval when constructed


islemaster:
When constructed, `five.Thermometer` [calls `setInterval()` with the initial sampling frequency](https://github.com/rwaldron/johnny-five/blob/7ec4648249a73a90776e90fd0f13f71acbac35a0/lib/thermometer.js#L985-L1001) but does not track the interval key, and thus does not expose any way to pause or stop the component from emitting `'data'` events.

It would be nice to expose `enable()` and `disable()` methods [like `five.Sensor` does](https://github.com/rwaldron/johnny-five/blob/master/lib/sensor.js#L303-L319), even if we want the Thermometer component to start reading immediately.  This would allow the component to be cleaned up and removed if necessary.

rwaldron:
Hi! Sorry for the delay, I was on parental leave from January 20th until March 20th and I'm still trying to catch up with everything. I appreciate your patience ❤️

islemaster:
No problem!  We're making steady progress on our end - I figured I'd get back to submitting a PR for this myself soon-ish anyway.  Thanks! (And congratulations?)

islemaster:
I've proposed a fix for this.  Please take a look!

islemaster:
This is resolved as of [v0.11.0](https://github.com/rwaldron/johnny-five/releases/tag/v0.11.0) by the changes in #1346.  Thanks!

PR #1346, opened 2017-05-25
Give Thermometer enable/disable methods

islemaster:
Addresses https://github.com/rwaldron/johnny-five/issues/1297 by giving the `Thermometer` component `enable()` and `disable()` methods.

I've added tests but I'm a little confused by the structure of the Thermometer unit tests - this feature lives on the prototype and should be the same for all controllers, but it looks like it has to be tested per-controller?  But similar tests aren't applied to all controllers either.  Let me know if I should make sure these tests get added to every controller, or if we can run them just once on the default controller, etc.

Thanks!

### DIFF
--- a/lib/thermometer.js
+++ b/lib/thermometer.js
@@ -928,8 +928,6 @@ function Thermometer(opts) {
     this, opts = Board.Options(opts)
   );
 
-  var freq = opts.freq || 25;
-
   // Analog Reference Voltage (default to board.io.aref || 5)
   this.aref = opts.aref || this.io.aref || 5;
 
@@ -943,7 +941,13 @@ function Thermometer(opts) {
     controller = Controllers.ANALOG;
   }
 
-  priv.set(this, {});
+  var state = {
+    enabled: typeof opts.enabled === "undefined" ? true : opts.enabled,
+    intervalId: null,
+    freq: opts.freq || 25,
+    previousFreq: opts.freq || 25,
+  };
+  priv.set(this, state);
 
   Board.Controller.call(this, controller, opts);
 
@@ -952,6 +956,25 @@ function Thermometer(opts) {
       return x;
     };
   }
+
+  // TODO: Move this out of the constructor
+  var eventProcessing = function() {
+    if (raw == null) {
+      return;
+    }
+
+    var data = {};
+    data.C = data.celsius = this.celsius;
+    data.F = data.fahrenheit = this.fahrenheit;
+    data.K = data.kelvin = this.kelvin;
+
+    this.emit("data", data);
+
+    if (this.celsius !== last) {
+      last = this.celsius;
+      this.emit("change", data);
+    }
+  }.bind(this);
 
   var descriptors = {
     celsius: {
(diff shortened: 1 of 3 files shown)


rwaldron:
> I've added tests but I'm a little confused by the structure of the Thermometer unit tests - this feature lives on the prototype and should be the same for all controllers, but it looks like it has to be tested per-controller? 

This is what I _try_ to do, but not what always ends up happening, for a variety of reasons. 

> But similar tests aren't applied to all controllers either. 

Such is the result of a relaxed contribution policy, but I'm ok with the incurred technical debt (for now).

> Let me know if I should make sure these tests get added to every controller,

Don't worry about that for now :)

> or if we can run them just once on the default controller, etc.

This should be fine for now.

lib/thermometer.js
```diff
@@ -943,7 +941,13 @@ function Thermometer(opts) {
...
   }
 
-  priv.set(this, {});
+  var state = {
+    enabled: typeof opts.enabled === "undefined" ? true : opts.enabled,
+    intervalId: null,
+    freq: opts.freq || 25,
+    previousFreq: opts.freq || 25,
+  };
+  priv.set(this, state);
```
rwaldron: 👍

lib/thermometer.js
```diff
@@ -968,7 +991,22 @@ function Thermometer(opts) {
...
+        state.freq = newFreq;
+        if (state.intervalId) {
+          clearInterval(state.intervalId);
+        }
+
+        if (state.freq !== null) {
+          state.intervalId = setInterval(eventProcessing, newFreq);
+        }
+      }
+    },
```
rwaldron: 👍

lib/thermometer.js
```diff
@@ -983,26 +1021,50 @@ function Thermometer(opts) {
...
-  setInterval(function() {
-    if (raw == null) {
-      return;
-    }
+  // Set the freq property only after the get and set functions are defined
+  // and only if the sensor is not `enabled: false`
+  if (state.enabled) {
+    this.freq = state.freq;
+  }
+}
```
rwaldron: 👍

lib/thermometer.js
```diff
@@ -983,26 +1021,50 @@ function Thermometer(opts) {
...
+  }
 
-    if (this.celsius !== last) {
-      last = this.celsius;
-      this.emit("change", data);
-    }
-  }.bind(this), freq);
-}
+  return this;
+};
```
rwaldron: 👍

lib/thermometer.js
```diff
@@ -983,26 +1021,50 @@ function Thermometer(opts) {
...
+
+  /* istanbul ignore else */
+  if (state.enabled) {
+    state.enabled = false;
+    state.previousFreq = state.freq;
+    this.freq = null;
+  }
+
+  return this;
+};
```
rwaldron: 👍

test/thermometer.js
```diff
@@ -450,7 +547,10 @@ exports["Thermometer -- ANALOG"] = {
       test.expect(1);
       test.equal(digits.fractional(this.thermometer.C), 0);
       test.done();
-    }
+    },
+    enable: testEnable,
+    disable: testDisable,
+    constructDisabled: testConstructDisabled,
```
rwaldron: 👍

islemaster:
Done: Documented `enabled`, `freq`, `enable()` and `disable()` on [the Thermometer wiki page](https://github.com/rwaldron/johnny-five/wiki/Thermometer) ([diff](https://github.com/rwaldron/johnny-five/wiki/Thermometer/_compare/215522e4b5ba4c2f25a7a265ded4dcb175737967...246f4c5810249234748a376dd9d6322c8053b524)).  Also documented the same on [the Sensor wiki page](https://github.com/rwaldron/johnny-five/wiki/Sensor) ([diff](https://github.com/rwaldron/johnny-five/wiki/Sensor/_compare/8e6a89add419e42aa89e7ddd4be1940a15d8bb05...d04788c2ff11e9fc001ddedabb3054ad6c2d0513)) which was missing these.

PR #1363, opened 2017-07-11
Multi-pin DS18B20

ruswerner:
This adds support for having multiple driver instances for the DS18B20 on different pins (1wire bus) concurrently. This has been tested in real-life using 6 thermos, 3 on each of 2 different pins. Works great!


rwaldron:
Nice work!

PR #1375, opened 2017-08-16
Handle number as first value of servo keyframes

### LINKED ISSUE #1374: Issue with moving the servo
My Edison board is connected to one servo (Grove). When I use the below code -
keyFrames: [ {degrees: 45}, {degrees: 90}, {degrees: 135}, {degrees: 180}, {degrees: 0}]
it works fine

but it does not moves when I use the below -
keyFrames : [ 45, 45, 45, 45, -180 ]
instead the servo makes a buzzing sound and stops after sometime.

Also, when I use -
 keyFrames: [{degrees: 90}, { value: 180, easing: "inQuad" }, { value: 90, easing: "outQuad" }, {degrees: 0}]
it works perfectly

but when I use - 
keyFrames: [90, { value: 180, easing: "inQuad" }, { value: 90, easing: "outQuad" }, -90],
the servo arm never touches 180.

It seems I have to use degrees explicitly. Just putting the values (45, 90 etc) does not works.
I am new to this framework and maybe I am doing something stupid but I could use some help :)

[Note - in all above code before execution, the servo arm is at 0 degree]

dtex:
When number is passed as the first keyframe of a servo animation it should be treated as a step from the most recent position.

Also adds tests for led to ensure that number values in first keyframe are still treated as explicit values.

Will close #1374 

Note that when this lands I need to update the animation wiki to better describe all devices that can be animated.

### DIFF
--- a/lib/servo.js
+++ b/lib/servo.js
@@ -320,6 +320,14 @@ Servo.prototype[Animation.normalize] = function(keyFrames) {
       value: last
     };
   }
+
+  // If user passes a step as the first element in keyFrames use current position + step
+  if (typeof keyFrames[0] === "number") {
+    keyFrames[0] = {
+      value: last + keyFrames[0]
+    };
+  }
+
   return keyFrames.map(function(frame) {
     var value = frame;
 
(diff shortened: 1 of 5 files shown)


rwaldron:
Cleared to update docs. Ping me on gitter when you're ready for me to deploy the site

PR #1391, opened 2017-09-20
fix: Switch defaults to NO

er1x:


lib/switch.js
```diff
@@ -41,6 +41,7 @@ function Switch(opts) {
   }, 5);
 
   // Resolve the default type to Normally Open
+  opts = typeof opts === "object" ? opts : {pin: opts};
```
rwaldron: This logic is already handled in `Board.Options`, so you should move this: https://github.com/er1x/johnny-five/blob/23558b98e11ae80fca46cf4391652f9b3da59096/lib/switch.js#L62-L64 to line 43-ish.

rwaldron:
As soon as travis is green, this is good to go.

Issue #1026, opened 2016-01-30
Add isAnode to Led Docs


reconbot:
It's missing as a possible property to the options.

scottgonzalez:
`isAnode` doesn't apply to `Led`. This is used to denote whether the common pin on a multi-LED component (e.g., RGB LEDs, seven segment displays, etc.) is anode or cathode. On a single LED, there is no common pin because it's just one diode.

reconbot:
It does work for leds that are wired in reverse however.

frxnz:
This topic comes up about every 6 months. We should definitely do _something_. I'd opt for removing it from `Led` altogether.

* `isAnode` is in `Led` because originally `Led.RGB` was just 3 instances of `Led`
* `Led.RGB` no longer extends from `Led`, so it really doesn't need to exist in `Led` anymore

I suppose it's true that you could use `isAnode` for an LED wired in reverse, but is there any practical reason to do that?

soundanalogous:
There are 2 different ways to wire an LED to a microcontroller. The more common way is to wire the LED annode to the microcontroller pin and the cathode to ground (with a resistor between the mcu pin and the annode or the cathode and gnd). This is known as a "source drive" circuit since the mcu pin is sourcing the current to power the LED. The other way is "sync drive" in which the cathode is wired to the mcu pin and the anode is wired to power and current flows into the mcu (which can be more power efficient in some designs but likely makes little difference for prototypes and hobby applications). When using sync drive, you set the pin value to 0 rather than 1 to turn the LED on. Therefore it's necessary to provide a means to differentiate between the two configurations in order to determine if `led.on` should set the pin value to 0 or to 1.

soundanalogous:
Here's a slightly more detailed explanation on source vs sync: https://startingelectronics.org/articles/current-sourcing-sinking/

frxnz:
That's really interesting, thanks for detailed explanation @soundanalogous. If Johnny-Five wants to support that configuration, `isAnode` still probably isn't the best name for that property.

dtex:
button, joystick, motor and servo all have an ```invert``` property that does a similar (conceptually) thing. Not saying it's the right property name (no hablo LED), just one to consider.

reconbot:
I actually thought invert existed and had to look at the source to figure
out isAnode.

Both kinda make sense here, for both led and rgbled. I'm happy as long as
there are docs. And happier if the APIs are consistent.

Resseguie:
[...]

PR #1398, opened 2017-10-11
LED.isAnode does not work for OUTPUT mode

rudacs:
I fixed a problem in isAnode in the component of the LED that I just implemented in PWM mode.
In this item is discussed to change the name to invert (yet to be done).
[#1026](https://github.com/rwaldron/johnny-five/issues/1026)

### DIFF
--- a/lib/led/led.js
+++ b/lib/led/led.js
@@ -34,8 +34,14 @@ var Controllers = {
       writable: true,
       value: function(input) {
         var state = priv.get(this);
-        var output = typeof input !== "undefined" ? input : state.value;
-        var value = state.isAnode ? 255 - Board.constrain(output, 0, 255) : output;
+        var value = typeof input !== "undefined" ? input : state.value;
+
+        if (this.mode == this.io.MODES.PWM) {
+          value = state.isAnode ? 255 - Board.constrain(value, 0, 255) : value;
+        } else if (state.isAnode) {
+          value = value == this.io.LOW ? this.io.HIGH : this.io.LOW;
+        }
+
         this.write(value);
       }
     },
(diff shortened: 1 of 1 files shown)


lib/led/led.js
```diff
@@ -156,8 +163,16 @@ function Led(opts) {
...
+    * @deprecated Will be deleted in version 0.11.8. Use 'inverted' instead.
+  */
+  if (typeof opts.isAnode !== 'undefined' && opts.isAnode !== null) {
+    console.log("isAnode is deprecated. Use 'inverted' instead");
+    opts.inverted = opts.isAnode;
+  }
+
   var state = {
-    isAnode: opts.isAnode,
+    inverted: opts.inverted,
```
soundanalogous: We should not use "inverted" because it is ambiguous. An LED is either common anode or common cathode. There is no such state as "inverted" because there is no default state to invert.

dtex: Linking this [issue](https://github.com/rwaldron/johnny-five/issues/1026) wherein @soundanalogous gives us the labels ```sourcedrive``` and ```syncdrive``` which makes sense to me.

rudacs: So will it continue to be isAnode?

dtex: As I understand it "common anode" and "common cathode" are RGB LED properties. This change is specifically for single color LED's. So I would think @soundanalogous's suggestion of "syncdrive" (from #1026) would be the appropriate property here.

I also propose that instead of adding this condition to every controller we add a "syncdrive" device type that has a new ```getValue()``` or maybe just ```value()``` method.

Lastly, there is a [new property](https://github.com/firmata/firmata.js/pull/149) on the io object (in firmata at least) that allows us to know the resolution of pwm writes (i.e. 10-bit instead of 8-bit). We should leverage that here rather than assuming 8-bit.

dtex: This is still an issue, but it's sat for a while. Perhaps we can settle on a label for the property at JSCONF this year?

dtex:
This is still an issue, but it's sat for a while. Perhaps we can settle on a label for the property at JSCONF this year?

PR #1399, opened 2017-10-12
Add expander ADS1115

rudacs:
I added the ADS1115 component, it is also necessary to do the ADS1015, which is almost the same thing, but I do not have to be able to test.
If anyone has it I can add the component and test it.


eg/expander-ADS1115.js
```diff
@@ -0,0 +1,27 @@
+var Barcli = require("barcli");
+var five = require("../lib/johnny-five");
+var board = new five.Board();
+
+board.on("ready", function() {
+  var virtual = new five.Board.Virtual(
+    new five.Expander("ADS1115")
+  );
+
+  virtual.io.analogPins.forEach(function(input) {
```
rwaldron: Oh, clever!

lib/expander.js
```diff
@@ -1785,6 +1785,114 @@ var Controllers = {
...
+
+              pin = newPin;
+
+              delay();
+            }
+          }.bind(this));
+        }
+      }
+    },
+  }
```
rwaldron: I wish this had tests, but I think I would like to write them so I can get familiar with this chip. Thanks for this _excellent work_

PR #1400, opened 2017-10-13
Fix sycnchronous animation queues

dtex:
#1376 only fixed the problem when animation segments were enqueued in seperate passes through the event loop.  When two animation segments were enqueued on the same pass through the event loop we were starting both segments because the isRunning state was not being set until nextTick.

This should fix all that. I've updated some of the tests which I believe were written in a non-intuitive manner because of this bug. Instead of really understanding the problem I had just written the tests in a way that made them pass. There's probably a name for this sort of behavior. If not, we should invent a name.

### DIFF
--- a/test/animation.js
+++ b/test/animation.js
@@ -360,7 +360,32 @@ exports["Animation -- Servo"] = {
     );
 
     test.done();
-  }
+  },
+  
+    enqueueOnSameTick: function(test) {
+      
+      test.expect(7);
+  
+      this.animation = new Animation(this.a);
+  
+      this.normalizeKeyframes = this.sandbox.spy(this.animation, "normalizeKeyframes");
+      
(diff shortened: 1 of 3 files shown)


dtex:
This is ready for nits

test/animation.js
```diff
@@ -360,7 +360,32 @@ exports["Animation -- Servo"] = {
     );
 
     test.done();
-  }
+  },
+  
+    enqueueOnSameTick: function(test) {
```
rwaldron: Strange indent?

rwaldron:
@dtex I think the funky indent is my only nit. Extended tests are your turf, as is Animation ;)

rwaldron:
I did the fix for you, since I took so long to review. Sorry about that

PR #1403, opened 2017-10-21
Add option longPress on Button

rudacs:
I added an event on the button (longPress) to facilitate this action.

### DIFF
--- a/docs/button-options.md
+++ b/docs/button-options.md
@@ -41,7 +41,8 @@ board.on("ready", function() {
     board: board,
     pin: 8,
     holdtime: 1000,
-    invert: false // Default: "false".  Set to "true" if button is Active-Low
+    invert: false, // Default: "false".  Set to "true" if button is Active-Low
+    longPress: 1500 // Default: "false".  Set to "true" to wait 1000 ms to trigger
   });
 
   // Inject the `button` hardware into
@@ -68,6 +69,12 @@ board.on("ready", function() {
   // "up" the button is released
   button.on("up", function() {
     console.log("up");
+  });
+
+  // "longPress" the button is pressed for specified time (once trigger).
+  //        defaults to 1000ms (1 second)
+  button.on("longPress", function() {
+    console.log("longPress");
   });
 });
 
--- a/eg/button-options.js
+++ b/eg/button-options.js
@@ -10,7 +10,8 @@ board.on("ready", function() {
     board: board,
     pin: 8,
     holdtime: 1000,
-    invert: false // Default: "false".  Set to "true" if button is Active-Low
+    invert: false, // Default: "false".  Set to "true" if button is Active-Low
+    longPress: 1500 // Default: "false".  Set to "true" to wait 1000 ms to trigger
   });
 
   // Inject the `button` hardware into
@@ -37,5 +38,11 @@ board.on("ready", function() {
   // "up" the button is released
   button.on("up", function() {
     console.log("up");
+  });
+
+  // "longPress" the button is pressed for specified time (once trigger).
+  //        defaults to 1000ms (1 second)
+  button.on("longPress", function() {
+    console.log("longPress");
   });
 });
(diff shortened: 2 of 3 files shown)


dtex:
Hi @rudacs what would be the use case for this versus setting ```holdtime``` in the options and listening for the ```hold``` event?

rudacs:
Holdpress is for tasks such as increasing brightness as the user keeps pressing.

Longpress only happens only once. It comes from the touch concept, where an action occurs if the user presses for a period.
It would be a facilitator for users rather than having to implement the same function multiple times in code.

Another facilitator that I will be sending is nTabs, which also comes from the touch concept.

dtex:
First off, I forgot to thank you for contributing to Johnny-Five. It's an awesome thing to do and very much appreciated.

I must be misunderstanding something... The hold event in the button class is only emitted once, after ```holdtime``` milliseconds which is what you have described for ```longpress```. We don't have anything like ```holdpress``` in J5.

rudacs:
Thank you. But I thank you for this wonderful project.
It is good to always contribute.

I wrote wrong. It would be the hold event (not holdpress)
The hold event occurs at each holdtime until the user releases the button.

Example:

press
hold
hold
hold
release

Since the longPress it is intended to only have a "single" hold so to speak.

press
longPress
release

Only in this case is it more likely that the user ignores the press and release events and only uses the longPress.

For the user to do this job currently he would have to do something like this:

```var longPress = false;

button.holdtime = 1000;

button.on("press", function() {
  longPress = false;
});

button.on("hold", function() {
  if (!longPress) {
    longPress = true;
    console.log("longPress");
  }
});
```

Already with longPress it would only be necessary:

```button.longPress = true;

button.on("longPress", function () {
  console.log("longPress");
});
```

dtex:
Button inherits node's EventEmitter class so ```once()``` would do the same without any changes to J5.
```
button.once("hold", function () {
  console.log("longPress"); // Won't repeat
});
````
But then the listener would have to be re-registered whenever the button is released. What's your use case? Does it make sense to re-register?

If not, would it be better for the API to just add a ```repeat``` property to the button opts that defaults to true (to preserve existing functionality) instead of adding another event? 
````
let button = new Five.Button({ repeat: false, holdtime: 1000 });

button.on("hold", ()=> {
  console.log("longPress"); // Only happens once because repeat === false
});
````
We should see wait and see what @rwaldron thinks. He'll have some thoughtful input for sure.

**Comment by soundanalogous** [comment]
If you are considering repeating the hold press event, also consider whether or not to repeat at the same frequency as the hold time or a shorter frequency. For example if holdtime = 1000, once that 1000 ms threshold is reached, you could repeat every 100ms (or even a shorter duration) after that and add the current repeat count to the event.

**Comment by rudacs** [comment]
I think the repeat property looks weird.
I would not know by the name of the property what it means.

On the eleven I would have to register the event every time the press event occurs or to work just like longPress would have to be something like this:

```
button.on("press", function() {
  button.once("hold", function() {
    console.log("longPress");
  });
});
```

But I think the way I'm doing is the most transparent and easy way for the user. Which I think is the proposal.

If not, we would not even have the Led blink for example.

would be like the user doing:

```
setInterval(function() {
  led.toggle();
}, 500);
```

soundanalogous:
You could also separate the concept of press and release from the concept of gestures. In that case you could write a gesture module that captures gestures such as tap, long press, double tap, etc, and just leave the simple press and release events to the button module.

rwaldron:
Coming into this after the details have been teased out and the rationale made clear is pretty nice—thanks to all of you for ironing this out. I _really_ like @soundanalogous's idea of making a "button gesture" module, so I spent some time working on this today: https://github.com/rwaldron/j5-button-gestures

Feel free to make PRs there to fix up or add any gestures. We can work together to make sure it meets all the needs, then make sure it's documented on the site and in the `Button` API docs.

dtex:
I'm going to close this since we are moving it to a module

PR #1404, opened 2017-10-21
Add option nTaps on Button

rudacs:
Added touch concept option.This option makes it easy for the user to perform an action when N taps occur.

Example: 3 taps to restart.

I will improve after this option to be an array where it goes will be possible to detect several numbers of touches.


dtex:
I'm going to close this since we are moving it to the gestures module

Issue #1409, opened 2017-11-03
GPIO pins on Raspberry Pi Zero


zoobot:
Hi 
I want to use the moto zero motor controller with the raspberry pi zero.
Ideally I'd like to use Johnny-five but not sure if this is supported?

Are there any examples of Johnny-five Motors using GPIO pins? I'm trying to use raspi-io but getting errors:

so far looks like raspi-io is not supporting the Motor functions.
getting this error
/node_modules/raspi-io/node_modules/raspi-io-core/dist/index.js:489
        throw new Error('Unknown pin "' + pin + '"');
        ^
Error: Unknown pin "undefined"

the Johnny-five motors look like this
var motor3 = new five.Motor({
pins: {
dir: 'P1-23', // dir: 'GPIO11' ??
cdir: 'P1-16',
enable: 5,
invertPWM: true
}
});

With the python GPIO lib it looks like this:
import RPi.GPIO as GPIO
GPIO.setmode(GPIO.BCM)
Motor1A = 27
Motor1B = 24
Motor1Enable = 5

https://cdn.shopify.com/s/files/1/0176/3274/files/MotoZero_User_Guide_1.2.pdf

I just noticed raspi-io says it is fermata compatible. I am confused about that, isn't the point of using raspi-io to use the raspberry pi GPIO pins directly instead of going through an fermata arduino or something?... How does fermata come into this situation?
https://github.com/nebrius/raspi-io/wiki/Pin-Information

Thanks!

rwaldron:
> I just noticed raspi-io says it is fermata compatible. I am confused about that, isn't the point of using raspi-io to use the raspberry pi GPIO pins directly instead of going through an fermata arduino or something?..

It's interface "compatible" with https://github.com/firmata/firmata.js


> With the python GPIO lib it looks like this: ...

Which Raspberry Pi are you using? 


> https://cdn.shopify.com/s/files/1/0176/3274/files/MotoZero_User_Guide_1.2.pdf

This document includes the following: 

> All examples in this document use the GPIO.BCM mode 

We'll have to ask @nebrius what impact this has on raspi-io

nebrius:
Related: https://github.com/nebrius/raspi-io/issues/100

> All examples in this document use the GPIO.BCM mode

Oh pin numbering on the Raspberry Pi, will you always be so complicated? /sarcasm

There are a variety of ways to refer to pin numbers on the Raspberry Pi, and each library takes a slightly different approach to how to number them.

[...]

PR #1419, opened 2017-11-30
Require motor pins

dtex:
Discussed in #1409

### DIFF
--- a/lib/motor.js
+++ b/lib/motor.js
@@ -169,6 +169,14 @@ var Controllers = {
         this.setPWM(this.pin, value);
       }
     },
+    validatePins: {
+      value: function() {
+        // Check for required pins
+        if (typeof this.opts.pin === "undefined") {
+          throw new Error("Pin must be defined");
+        }
+      }
+    }
   },
 
   GROVE_I2C_MOTOR_DRIVER: {
@@ -253,6 +261,14 @@ var Controllers = {
           this.COMMANDS.NOOP,
         ]);
       }
+    },
+    validatePins: {
+      value: function() {
+        // Check for required pins
+        if (typeof this.opts.pin === "undefined") {
+          throw new Error("Pin must be defined");
+        }
+      }
     }
   }
 };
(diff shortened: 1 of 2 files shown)


lib/motor.js
```diff
@@ -497,19 +517,32 @@ function Motor(opts) {
 
   // Derive device based on pins passed
   if (typeof this.opts.device === "undefined") {
-
-    this.opts.device = (typeof this.opts.pins === "undefined" && typeof this.opts.register !== "object") ?
-      "NONDIRECTIONAL" : "DIRECTIONAL";
-
-    if (this.opts.pins && (this.opts.pins.cdir || this.opts.pins.length > 2)) {
-      this.opts.device = "CDIR";
+    this.opts.device = "DIRECTIONAL";
```
rwaldron: `Motor` is the only oddball class that does this thing with `this.opts` :|     
```js
Board.Component.call(
  this, this.opts = Board.Options(opts)
);
```

Every other class just does 

```js
Board.Component.call(
);
```
Because `Board.Component` and `Board.Options` normalize all the pins and put them directly on `this`. It works, so I wouldn't say change it all here and now, but I think we should get it consistent eventually.

test/motor.js
```diff
@@ -2082,27 +2089,34 @@ exports["Motor.Collection"] = {
...
+        pwm: 5,
+        dir: 6
+      }
     }, {
-      pwm: 9,
-      dir: 10
+      pins: {
+        pwm: 9,
+        dir: 10
+      }
```
rwaldron: I guess this should also check and make sure that the Motor objects are constructed correctly

rwaldron:
Ping me anywhere when ready (gitter, text, whatever)

lib/motor.js
dtex: I bet there are some things kept on this.opts that I refer back to outside of the constructor. I just need to move that/those specific cases to the private state, or properties on this (depending on what it is).

test/motor.js
dtex: I added tests that make sure each member of the collection is an instance of Motor.

lib/motor.js
dtex: I'm going to handle this in a separate PR. I can just change the pins getter/setter on each device type to a normalizePins method that gets called by the constructor, receives opts as a param and returns the normalized pins object. The constructor will assign the returned value to this.pins.

PR #1426, opened 2017-12-22
Accelerometer: add LIS3DH fractional digits

dnlup:
Add device sensitivity to `toGravity.value` and test
case.


rwaldron:
Excellent!

PR #1427, opened 2017-12-27
Fix relay to use analog output

### LINKED ISSUE #1277: Using pin A2 as relay control pin
I have an arduino board and would like to use pin A2 as the relay control pin, but this does not seem work. It seems like port D2 is used instead. I am wondering why that is? I think, it should be possible to use the analog pins as several tutorials are suggesting. Do I need a customized Firmata configuration for this?

bissonex:
Fixes #1277


rwaldron:
@bissonex Sorry for the delay in response, I've landed this now. I've also added tests and an example program—thanks again for the contribution!

PR #1430, opened 2017-12-30
Offset was being applied twice when to() was called with a time

### LINKED ISSUE #1429: Offset considered twice
Hello,

On the Servo class, when a time is specified on the Servo.to() method, the offset is considered twice
eg with -10 offset : 

> 
> servo.to(80)
> servo.value //70
> 
> servo.to(80,100)
> servo.value //60

Observed with ESP8266/EtherPortClient and PCA9685 as servo controller

dtex:
Move application of offset so that is only applied once.

Move update of this.value to controller update methods. A little less DRY, but having them be more proximate to the actual servoWrite makes sense.

Fixes #1429

### DIFF
--- a/lib/servo.js
+++ b/lib/servo.js
@@ -35,6 +35,7 @@ var Controllers = {
       writable: true,
       value: function(degrees) {
         var state = priv.get(this);
+        this.value = degrees;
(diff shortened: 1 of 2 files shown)


rwaldron:
Landed

PR #1434, opened 2018-01-05
[WIP] Barometer fixed point values

### LINKED ISSUE #1119: Barometer: update fixed point values to match datasheets for fractional digit accuracy

And the corresponding tests: 

dnlup:
Fix #1119


dnlup:
@rwaldron Hi, sorry to bother again. I tried to work on issue #1119 and did not find test implemented for some devices. Is it something that will be added in the future or it is better to add them now? I would be happy to help but still not super familiar with the code, should I consider the `Barometer -- MPL3115A2` as skeleton for each device? Plus I could not make the test on the `BMP180` device work because I could not trigger the spy on the `data` event handler (tried to wait the call with a `setInterval` loop as done in `Barometer -- MPL115A2`). Thank you.

rwaldron:
This is sooooo awesome :)

PR #1437, opened 2018-01-17
Servo bug fixes

### LINKED ISSUE #1436: Servo offset bugs
There are two issues:

### Servo range is improperly constrained
In this scenario:
````
let servo = new five.Servo({
  pin: 5,
  offset: 20
});

servo.to(170);
````
J5 sends ```190``` to the io-plugin

We need to constrain to the offset range [-20, 160] 

### We store the wrong value in Servo.value
In this scenario:
````
let servo = new five.Servo({
  pin: 5,
  offset: 20
});

servo.to(120);
````
J5 stores ```140``` in servo.value.

It should store the user requested value.

I'll work up a PR.

**Sidenote:** These have been around a while and were not introduced by 010d49635864e784d8f341b4d01c3eed9fe1a56a

dtex:
Closes #1436

### DIFF
--- a/lib/altimeter.js
+++ b/lib/altimeter.js
@@ -75,7 +75,7 @@ var Controllers = {
         // Resolution of output data 0.01hPa
         //
         // From paragraph 3.6, page 16 1hPa=8.43m
-        // resolution ~= 0,08m
+        // resolution ~= 0.08m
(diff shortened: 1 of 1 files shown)


dtex:
Merge Remote Tracking blargh! I'll see if I can't weed those out.

PR #1439, opened 2018-01-19
[WIP] Proximity fixed point values

dnlup:
Add proximity devices resolutions.

### DIFF
--- a/lib/proximity.js
+++ b/lib/proximity.js
@@ -130,7 +130,15 @@ var Controllers = {
       value: function(raw) {
         // From http://www.maxbotix.com/articles/032.htm
         // ADC -> inches -> cm
-        return (raw / 2) * 2.54;
+        //
+        // Datasheet available at https://www.maxbotix.com/documents/LV-MaxSonar-EZ_Datasheet.pdf
+        //
+        // From intro in page 1
+        // 'The LV-MaxSonar-EZ detects objects
+        // from 0-inches to 254-inches (6.45-meters) and provides sonar range information from 6-
+        // inches out to 254-inches with 1-inch resolution.'
+        // 1inch = 2.54cm
+        return toFixed((raw / 2) * 2.54, 2);
       }
     }
   },
--- a/lib/proximity.js
+++ b/lib/proximity.js
@@ -268,7 +268,7 @@ var Controllers = {
     },
     toCm: {
       value: function(raw) {
-        return toFixed(raw / 29.1 / 2, 3);
+        return toFixed(raw / 29.1 / 2, 1);
       }
     }
   },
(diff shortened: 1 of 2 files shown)


lib/proximity.js
```diff
@@ -163,7 +176,15 @@ var Controllers = {
...
-        return raw;
+        //
+        // Datasheet available at https://www.maxbotix.com/documents/XL-MaxSonar-EZ_Datasheet.pdf
+        //
+        // From intro on page 1
+        // 'The MB1200 and MB1300 sensor series detects objects from 0-cm1
+        //  to 765-cm (25.1 feet) or 1068cm (35 feet) (select models) and
+        // provide sonar range information from 20-cm2
+        //  out to765-cm or 1068-cm (select models) with 1-cm resolution...'
+        return toFixed(raw, 0);
```
dnlup: @rwaldron  Here I forced the call to `toFixed` even if not needed. Maybe it is too much. I can revert it if you prefer.

lib/proximity.js
```diff
@@ -276,7 +304,12 @@ var Controllers = {
     },
     toCm: {
       value: function(raw) {
-        return raw;
+        // Datasheet available at http://www.robotshop.com/media/files/pdf/operating-manual-llm20c132i500s011.pdf
+        //
+        // From `Technology` paragraph at page 11
+        // 'Our patented, high"accuracy"
+        // measurement"technique"enables"distance"measurement"accuracy down"to 1cm..'
+        return toFixed(raw, 0);
```
dnlup: @rwaldron Same as https://github.com/rwaldron/johnny-five/pull/1439/files#diff-c5fe53df19f93fb5988d5336a99e42a9R187.

dnlup:
Sorry, but I could not find valuable data on `EV3` and `SRF10` devices, plus many analog ones:
* GP2Y0A21YK
* GP2D120XJ00F
* GP2Y0A02YK0F
* GP2Y0A41SK0F
* GP2Y0A710K0F

lib/proximity.js
rwaldron: How about an optimization? Truncate the decimal digits by shifting the bits off: `raw >> 0`, this shifts right by `0` bits, but first it must convert the left hand operand to an int32—which means discarding the exponent bits :)

rwaldron:
This is excellent work! I've checked that travis failure and it appears to be unrelated to these changes (also, it passes locally)

rwaldron:
Landed!

dnlup:
@rwaldron Thanks!
Travis had problems with osX builds at that time, it happened that more than more PR failed because of it :smile: .

lib/proximity.js
dnlup: I'll do it and re open this PR, thanks!

PR #1444, opened 2018-02-02
Hygrometer fixed point values

### LINKED ISSUE #1122: Hygrometer: update fixed point values to match datasheets for fractional digit accuracy

And the corresponding tests: 

dnlup:
Add resolutions and tests (where implemented) of hygrometer class devices.

Fix #1122


rwaldron:
👏

PR #1445, opened 2018-02-11
LSM303C 6 DoF IMU (Magnetometer, Accelerometer, Temp)

sdesalas:
Hi Rick,

I got this working for one of my projects and decided to add a pull request back to main repo in case other people would like to use it.

Included:

- [x] libs/drivers
- [x] fritzing diagram
- [x] annotated example script in the eg/ folder
- [x] api documentation in the wiki
- [x] unit tests

I've tried to follow the 'flavour' of the code as much as possible (function.bind, similar examples, docs, test structure etc).

Couple things to bear in mind though:

1. The module I bought was a clone from eBay but works with the same arduino libraries and docs and the Sparkfun version, I actually used these libraries to work out the code here (registers, setup sequence, etc). If you have a Sparkfun original feel free to give it a try as well.

2. Note that I've also exposed `raw` (x/y/z) values in `compass.js`. The reason for this is because I actually needed raw values in my project rather than calculated bearing/direction and I though other people might be in the same boat. 

Let me know if I missed anything or any questions/changes you have in mind.


rwaldron:
Wow.


(Seriously, that's all I can say)

rwaldron:
@sdesalas I've gone ahead and added all of this to the docs: 

- https://github.com/rwaldron/johnny-five/wiki/IMU
- https://github.com/rwaldron/johnny-five/wiki/Accelerometer
- https://github.com/rwaldron/johnny-five/wiki/Compass
- https://github.com/rwaldron/johnny-five/wiki/Thermometer

Again—GREAT WORK!!!

sdesalas:
Awesome thanks Rick! 

I missed the extra docs, i think i'll add a PR to document the `raw` (xyz) property on `Compass` so other people can see it.

PR #1447, opened 2018-02-16
Gyro: add MPU6050 and BNO055 devices resolution

### LINKED ISSUE #1121: Gyro: update fixed point values to match datasheets for fractional digit accuracy

And the corresponding tests: 

dnlup:
Add test only for MPU6050 device.

Fix #1121 .

### DIFF
--- a/lib/gyro.js
+++ b/lib/gyro.js
@@ -76,10 +76,19 @@ var Controllers = {
       }
     },
     toDegreesPerSecond: {
+      // Datasheet available at https://www.invensense.com/wp-content/uploads/2015/02/MPU-6000-Datasheet1.pdf
+      //
+      // From paragraph 6.1 at page 12
+      // Sensitivity scale factor
+      // FS_SEL=0    131 LSB/dps -> 0,007633588 dps/LSB
+      // FS_SEL=1    65.5 LSB/dps -> 0,015267176 dps/LSB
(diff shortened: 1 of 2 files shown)


lib/gyro.js
```diff
@@ -301,9 +314,9 @@ function Gyro(opts) {
           this.toDegreesPerSecond(state.z.value, state.z.center) : 0;
 
         return {
-          x: toFixed(x, 2),
-          y: toFixed(y, 2),
-          z: toFixed(z, 2)
+          x: x,
+          y: y,
+          z: z
```
rwaldron: Nice clean up

lib/gyro.js
```diff
@@ -76,10 +76,19 @@ var Controllers = {
...
+      // FS_SEL=0    131 LSB/dps -> 0,007633588 dps/LSB
+      // FS_SEL=1    65.5 LSB/dps -> 0,015267176 dps/LSB
+      // FS_SEL=2    32.8 LSB/dps -> 0,00304878 dps/LSB
+      // FS_SEL=3    16.4 LSB/dps -> 0,06097561 dps/LSB
+      // Using 4 digits resolution
       value: function(raw, rawCenter) {
         var state = priv.get(this);
 
-        return (raw - rawCenter) / state.sensitivity;
+        return toFixed((raw - rawCenter) / state.sensitivity, 4);
```
rwaldron: 👍

PR #1449, opened 2018-02-23
Proximity fixed point values

### LINKED ISSUE #1123: Proximity: update fixed point values to match datasheets for fractional digit accuracy

And the corresponding tests: 

dnlup:
Fix #1123


dnlup:
Sorry, I did't see this has already been released in `0.14.1`. I'll close it.

PR #1453, opened 2018-03-15
fix disable to  enable state true

chacalhell:
Hello, this is my first upload to a repository in a large project, I was using sensor.enable and sensor.disable continuously and I had problems when it was in a disabled state when I came back to enable, this arrangement is the solution I could find. Sorry for my English. And if there is any other way to do it, I apologize. Regards!

### DIFF
--- a/lib/sensor.js
+++ b/lib/sensor.js
@@ -313,6 +313,7 @@ Sensor.prototype.enable = function() {
   /* istanbul ignore else */
   if (!state.enabled) {
     this.freq = state.freq || state.previousFreq;
+    state.enabled = true;
   }
 
   return this;


dtex:
This seems straightforward but I have a request... It would be helpful if we had a test that would catch the condition you were seeing before your fix. Would you be interested in writing a test for that?

chacalhell:
Yes.
TEST:
--------

var S1 = new five.Sensor("A5");

sensor.on("data", function(value) {
    console.log(value);
  });

S1.disable();
S1.enable();
S1.disable(); // here not function disable because the state.enabled =
false

2018-03-15 20:35 GMT-03:00 Donovan Buck <notifications@github.com>:

dtex:
Johnny-Five has a series of unit tests that get run on demand or anytime new code is submitted via a pull request. [Here are a couple of the unit tests for sensor](https://github.com/rwaldron/johnny-five/blob/master/test/sensor.js#L1303-L1353). You can also run them locally by running ```grunt nodeunit```. I was hoping we could add a test here for the situation you have uncovered. As I understand it if you call disable and then enable and then disable again, it will not disable the second time and you keep getting data.

dtex:
Keeping this alive. I'll write the tests.

Issue #1406, opened 2017-10-25
Board.detect should be throttled


mrmiller:
If there are no valid serial ports available, Board.detect will spin with repeated calls to itself and use 100% CPU. Detect should be called on a timer instead so that the calls are throttled to a reasonable rate.

dtex:
Sorry nobody has responded before now. 

This seems like a reasonable thing to do. There is also an attempt counter that is not being used (as far as I can tell). Would you like to submit a PR?

@reconbot Do you have any words of guidance on this?

[Relevant code block](https://github.com/rwaldron/johnny-five/blob/master/lib/board.js#L32-L102)
[Offending line](https://github.com/rwaldron/johnny-five/blob/master/lib/board.js#L91)

reconbot:
No events to hook into or anything. Polling every 200ms after the first scan is probably more than enough.

mrmiller:
Easiest solution would probably be to add a time.sleep call before calling Serial.Detect again. As @reconbot said, 200ms is more than frequent enough.

Using the attempt counter could have a similar effect but we would still want the sleep call, otherwise the client code would have to implement the throttling. I imagine Serial.detect would raise an exception if the optional max attempt count was exceeded. On the plus side, the client would have an indication that the device wasn't found rather than busy-waiting in the object constructor. It's currently pretty opaque, but this is a separate issue.

Should I put together a PR for the sleep call? Should be a one-liner.

dtex:
@rwaldron will have thoughts (and probably a better solution) so hold off for now but I think it can be non-blocking (setTimeout), give up after 10 attempts, and then call [board.fail()](http://johnny-five.io/api/board/#logging).

mrmiller:
Non-blocking seems like the best design decision overall. My only worry is that change would  break existing code that may be implicitly relying on the current blocking behavior.

reconbot:
The serial port is listed async, no way to be sync about it.

mrmiller:
You're absolutely right—serialport.list is at least structured as async (though I couldn't find a guarantee that any given platform implementation is async). In that case, it's already (likely) non-blocking and we should probably just throttle it with sleep.

As for finishing implementing the attempt counter, limiting the number of retries by default would certainly change existing behavior. As of now, the behavior is to just use an excessive amount of CPU while the device isn't connected but it will detect it if and when it shows up. With a default attempt limit, that would no longer be the case.

[...]

PR #1456, opened 2018-03-23
Serial port list throttling and repl error with multiple boards

dnlup:
Reading issue #1406 I started playing with Board.detect and indeed in case
 of no connection the function gets called with no throttling.

I found a couple of other things too:

* if a connected board(s) is(are) disconnected while running, when terminating the process the SIGINT event gets sometimes ignored (it has happened running a board with `repl: true` and `repl: false`)
 
* if multiple boards are connected with `repl: true`, when exiting the process an error is thrown because `state.board.io.pending` is not defined

In this PR a simple solution to the first problem follows what @dtex suggests in the issue discussion, while for the other two I add a fail safe timer of 1 second to force exit and fix the code in the `repl.js` module to handle the case of multiple boards.

This is not definitive and I don't know how much of this is needed, I would like to know what do you think (@rwaldron @dtex).

### DIFF
--- a/lib/board.js
+++ b/lib/board.js
@@ -1195,6 +1195,11 @@ function Boards(opts) {
       this.emit("error", error);
     }.bind(this));
 
+    // echo the fail event from the boards
+    this[index].on("fail", function (event) {
+      this.emit("fail", event);
+    }.bind(this));
+
     return new Promise(function(resolve) {
       this[index].on("ready", function() {
         resolve(initialized[portOpts.id]);
--- a/lib/board.js
+++ b/lib/board.js
@@ -25,11 +25,14 @@ var rport = /usb|acm|^com/i;
 // This string appears over 20 times in this file.
 var UNDEFINED = "undefined";
 
-
 var Serial = {
   used: [],
   attempts: [],
   detect: function(callback) {
+    // Max number of times listing serial conntections can fail
+    var maxAttempts = 10;
+    // Delay (ms) before trying again to list serial connections
+    var retryDelay = 400;
     var serialport;
 
     /* istanbul ignore if */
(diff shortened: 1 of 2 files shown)


lib/board.js
```diff
@@ -88,7 +91,14 @@ var Serial = {
         Serial.attempts[Serial.used.length]++;
 
         // Retry Serial connection
-        Serial.detect.call(this, callback);
+        if (Serial.attempts[Serial.used.length] > maxAttempts) {
+          this.fail("Board", "No connected device found");
```
dnlup: Here I would have logged the id of the board but I haven't because when initializing multiple boards without specifing ids they all end up having the same generated uid. Is this how it is supposed to be?

rwaldron:
I'm going to need a bit of time to test and fully understand these changes. I appreciate your patience.

dnlup:
@rwaldron Absolutely, thanks for reviewing this PR. If you need anything (a basic code sample that triggers the issue or anything else) I am here to help.

rwaldron:
@dtex @reconbot @monteslu Any thoughts on this?

lib/board.js
```diff
@@ -25,11 +25,14 @@ var rport = /usb|acm|^com/i;
...
 
-
 var Serial = {
   used: [],
   attempts: [],
   detect: function(callback) {
+    // Max number of times listing serial conntections can fail
+    var maxAttempts = 10;
+    // Delay (ms) before trying again to list serial connections
+    var retryDelay = 400;
```
reconbot: My gut says this is a fine timeout.

rwaldron:
I've been in the middle of moving my office for about a month now :( All of my hardware is in boxes and I'm slowing unpacking things. I'd like to try this with real hardware before I land it—I appreciate your patience

dtex:
Is anybody here able to test with X-Bee's? It's probably fine, but that's a special case that may be affected.

rwaldron:
I've tested this against hardware and I'm satisfied with the changes—thanks and great work!!

dnlup:
@rwaldron Happy this PR was useful :-). I have one question if you don't mind.
While testing this code I saw that if I initialize multiple boards
without passing ids

```
var boards = new five.Boards([
   { port: '/dev/ttyACM0' },
   { port: '/dev/ttyACM1' },
   { port: '/dev/ttyACM2' }
]);

```

all the boards will have the same generated id. Is this how it is supposed to be
or the ids should be different?
(See comment https://github.com/rwaldron/johnny-five/pull/1456/files/0c70eee531d70bd90924d77d77c6b5feced29f7c#diff-54ffaa39b9e68a102065e0d7a79b6ea8R94)

PR #1387, opened 2017-09-12
Multi turn servo

dtex:
* Adds support for 90, 360 and multi-turn servos.
* Send servoWrites with values in microseconds for pulse width instead of degrees.
* Closes #1090 

This is for testing and cannot be merged until we confirm that all the IO plug-ins support passing position in degrees or pulse width in microseconds.

 - [ ] [imp-io/tyrion PR #8](https://github.com/rwaldron/tyrion/pull/8) [50hz, 600us - 2000us , 3% - 10%]
- [ ] [tessel-io PR #21](https://github.com/rwaldron/tessel-io/pull/21) [50hz, 600us - 2400us , 3% - 12%]
- [ ] [raspi-io/rasp-io-core Issue #4](https://github.com/nebrius/raspi-io-core/issues/4) [50hz, 1000us - 2000us, 5% - 10%]
- [ ] [galileo-io/edison-io](https://github.com/rwaldron/galileo-io/blob/b8f5f9ed9c2866f5557317fb1436a9c0a21e4f08/lib/index.js#L566-L576) [125.5hz, 600us - 2400us, 7.5% - 30.12%]
- [ ] [particle-io/voodoospark](https://github.com/voodootikigod/voodoospark/blob/master/firmware/voodoospark.cpp#L1009-L1019) [50hz, 1000us - 2000us, 5% - 10%]
- [ ] [nino-io](https://github.com/rwaldron/nino-io/blob/a676fda7331b988f9f0685218a7580b06d1988ae/lib/nino-io.js#L472-L482) [50z, 1000us - 2000us, 5% - 10%]
 - [ ] [beaglebone-io](https://github.com/julianduque/beaglebone-io) calls linux-io [50hz, 600us - 2400us, 3% - 12%]
 - [ ] [linux-io](https://github.com/fivdi/linux-io/blob/v0.8.0/lib/linux-io.js#L229-L247) [50hz, 600us - 2400us, 3% - 12%]
 - [x] [firmata.js]() [50hz, 544us - 2400us, 2.72% - 12%] Confirmed w/ scope
 - [x] [blend-micro-io](https://github.com/noopkat/blend-micro-io) servoWrite not implemented
 - [x] [board-io](https://github.com/achingbrain/board-io/blob/cf1c608500261c74bbf1d7a05aa3bff58dd70a62/lib/board-io.js#L136-L138) Not an issue
 - [x] [chip-io](https://github.com/sandeepmistry/node-chip-io)  servoWrite not implemented
[...]

PR #1461, opened 2018-04-16
Add support for multi-turn, winch, 90 and 360 degree servos

dtex:
Previously #1387 

Servo values are now sent to io plug-ins in microseconds instead of degrees. This allows for better accuracy and different device ranges.

All servos now default to a PWM range of 600-2400 uS regardless of io-plugin being used (was a complete cluster before and couldn't be overridden on most platforms).

Adds new ```deviceRange``` property to servo opts that allows the user to specify the range of the servo. Defaults to 0-180°.

Changes the servo startAt value to be the center of the servo regardless of deviceRange unless otherwise specified by the user.

Adds servo multi-turn example

Updates PCA-9685 ```servoWrite()``` to allow values to be written in microseconds. This required updating esc as well.

Updates test.

Requires recent versions of most io-plugins (firmata.js was already fine):

imp-io v0.3.0 / Tyrion v0.0.3
particle-io v0.15.0 / VoodooSpark v4.1.0
tessel-io v1.2.0
raspi-io v8.1.1
beaglebone-io v3.0.0
linux-io v0.9.2

### DIFF
--- a/eg/servo-multi-turn.js
+++ b/eg/servo-multi-turn.js
@@ -1,0 +1,45 @@
+var five = require("../lib/johnny-five.js");
+var board = new five.Board();
+
+board.on("ready", function() {
+  var servo = new five.Servo({
+    pin: 10,
+    // Cheapo servo has limited PWM range
+    pwmRange: [600, 2370],
+    // This multi-turn servo can turn 2430 degrees or 6.75 revolutions
+    deviceRange: [0, 2430]
+  });
+
+  // Add servo to REPL (optional)
+  this.repl.inject({
+    servo: servo
+  });
+
+
+  // Servo API
+
+  // min()
+  //
+  // move the servo to its minimum position
+  //
+  // eg. servo.min();
+
+  // max()
+  //
+  // move the servo to its maximum position of 6.75 turns from the minimum
+  //
+  // eg. servo.max();
+
+  // center()
+  //
+  // centers the servo at 3 3/8 turns from the minimum
+  //
+  // servo.center();
+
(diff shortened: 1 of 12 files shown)


dtex:
This is tested on hardware and ready for nits.

eg/servo-multi-turn.js
```diff
@@ -0,0 +1,45 @@
...
+  //
+  // centers the servo
+  //
+  // servo.center();
+
+  // to( deg )
+  //
+  // Moves the servo to position by degrees
+  //
+  // servo.to( 90 );
```
rwaldron: Let's update all of these examples to be multi-turn relevant/interesting.

lib/esc.js
```diff
@@ -397,7 +398,7 @@ ESC.prototype.stop = function() {
   var history = state.history;
   var speed = this.type === "bidirectional" ? this.neutral : 0;
 
-  this.write(this.pin, Fn.fscale(speed, 0, 100, 0, 180));
+  this.write(this.pin, Fn.fscale(speed, 0, 100, this.pwmRange[0], this.pwmRange[1]));
 
   history.push({
     timestamp: Date.now(),
```
rwaldron: This is going to drastically improve the usefulness of `ESC` objects!

dtex: Bonus! One of these days I'm going to have to learn more about ESC's

dtex: I don't think the request for review will close unless I edit line 401, but this wasn't really a change request... Not sure what will happen if I click dismiss review.

PR #1472, opened 2018-05-27
Made the example easier to read.

joshuapinter:


joshuapinter:
Copy that! I'll close this, then. :)

PR #1477, opened 2018-06-10
Add missing commas in stepper comments

meech-ward:
Added missing commas to some of the examples in the comments. For example, I changed this:

```js
five.Stepper({
  type: five.Stepper.TYPE.FOUR_WIRE
  stepsPerRev: number,
```

to this

```js
five.Stepper({
  type: five.Stepper.TYPE.FOUR_WIRE,
  stepsPerRev: number,
```

### DIFF
--- a/lib/stepper.js
+++ b/lib/stepper.js
@@ -56,7 +56,7 @@ function isSupported(io) {
  *
  *
  * five.Stepper({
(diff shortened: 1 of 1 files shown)


dtex:
Thanks for the catch @meech-ward

Issue #370, opened 2014-06-02
Migration to a 0-100% value range


rwaldron:
Some time ago, @theycallmeswift and I were discussing the simplification and alignment of the value ranges our respective projects use for writing to a board. For context, he is the author of [BreakfastSerial](https://github.com/theycallmeswift/BreakfastSerial) which is a Python cousin of Johnny-Five. In BreakfastSerial, Swift is using 0-100 (%) as the value range for any "writable" objects. Here's some examples: 
- `led.brightness(100)` is the same as writing `led.brightness(255)` in Johnny-Five today (https://github.com/theycallmeswift/BreakfastSerial#led-brightness)
- `motor.start(100)` is the same as our `motor.speed(255)` (https://github.com/theycallmeswift/BreakfastSerial#control-a-dc-motor)

Turns out this way easier for users to understand and Swift shared that it's been much easier to teach people 0-100% vs. 0-255 8-bit values. Based on this, I had outlined a phasing plan that went something like this: 
1. 0.8.x introduces 0-100% as the default value range used by anything that would previously expect 0-255 8-bit values. User code can opt-out, allowing authors to still use 8-bit values.
   - Opt-out is defined by an option property set via `Board({ use8bitRange: true })` (or something like that)
   - Omitting the opt-out will result in a big red message being logged to the console indicating that all analog write values will be scaled from 0-100% range to 0-255
2. 0.9.x eliminates the warning and the opt-out.

scottgonzalez:
I guess this needs to be bumped to 0.9.x/0.10.x?

Resseguie:
@rwaldron shall we add a 0.9 label so we don't forget this at the time?

dtex:
Is this something we want to do on 0.9.x? I think it would necessitate a 0.10 bump. I know 0.x releases don't have to follow semver, but it seems like the right thing to do in this case.

rwaldron:
Yeah, we'll need to relabel anything that didn't get done for 0.9.0

dtex:
The recent [changes in firmata.js](https://github.com/firmata/firmata.js/pull/149) highlight the need to get this done. It's like the universe is saying "ping".

rwaldron:
Everytime I think about working on this, I get overwhelmed by the amount of work it will take and then just decide to ignore it until it comes up again.

monteslu:
👍 
Separating firmata protocol from transport is exactly the type of whacky that we need :)

dtex:
[...]

PR #1484, opened 2018-07-16
ESC: deprecate speed(percent), introduce throttle(us).

rwaldron:
- Eliminates "history" memory leak
- Deprecates speed()
- Introduces throttle(us)

### DIFF
--- a/eg/esc-blobfish.js
+++ b/eg/esc-blobfish.js
@@ -1,39 +1,14 @@
-<!--remove-start-->
-
-# ESC - Keypress controlled ESCs
-
-<!--remove-end-->
-
-
-
-
-
-
-##### Breadboard for "ESC - Keypress controlled ESCs"
-
-
-
-![docs/breadboard/esc-keypress.png](breadboard/esc-keypress.png)<br>
-
-Fritzing diagram: [docs/breadboard/esc-keypress.fzz](breadboard/esc-keypress.fzz)
-
-&nbsp;
-
-
-
-
-Run this example from the command line with:
-```bash
-node eg/esc-keypress.js
-```
-
-
-```javascript
-const {Board, ESC, Fn, Led} = require("johnny-five");
+const {Board, ESC, Fn, Led} = require("../lib/johnny-five.js");
 const keypress = require("keypress");
+
 const board = new Board();
 
-board.on("ready", function() {
+board.on("error", error => {
+  console.error(error);
+  process.exit(1);
+});
+
+board.on("ready", () => {
   const led = new Led(13);
   const esc = new ESC({
     device: "FORWARD_REVERSE",
@@ -80,24 +55,3 @@ board.on("ready", function() {
   process.stdin.setRawMode(true);
   process.stdin.resume();
 });
-
-```
-
-
-
-
-
-
-
-
-&nbsp;
-
-<!--remove-start-->
-
-## License
-Copyright (c) 2012-2014 Rick Waldron <waldron.rick@gmail.com>
-Licensed under the MIT license.
-Copyright (c) 2015-2019 The Johnny-Five Contributors
-Licensed under the MIT license.
-
-<!--remove-end-->
--- a/lib/esc.js
+++ b/lib/esc.js
@@ -1,19 +1,17 @@
-var Board = require("./board");
-var Expander = require("./expander");
-var Pins = Board.Pins;
-var Collection = require("./mixins/collection");
-var Fn = require("./fn");
-var Emitter = require("events").EventEmitter;
-var util = require("util");
+const Board = require("./board");
+const Expander = require("./expander");
+const Pins = Board.Pins;
+const Collection = require("./mixins/collection");
+const {scale, fscale, constrain} = require("./fn");
 
-var priv = new Map();
+const priv = new Map();
 
 
-var Controllers = {
+const Controllers = {
   PCA9685: {
     initialize: {
       value: function(opts) {
-        var state = priv.get(this);
+        const state = priv.get(this);
 
         this.address = opts.address || 0x40;
         this.pwmRange = opts.pwmRange || [1000, 2000];
@@ -30,11 +28,11 @@ var Controllers = {
         this.pin = state.expander.normalize(opts.pin);
       }
     },
-    write: {
+    update: {
       writable: true,
-      value: function(pin, us) {
-        var state = priv.get(this);
-        state.expander.servoWrite(pin, us);
+      value: function(us) {
+        const state = priv.get(this);
+        state.expander.servoWrite(this.pin, us);
       }
     }
   },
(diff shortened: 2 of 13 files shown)


lib/esc.js
```diff
@@ -209,34 +207,22 @@ function ESC(opts) {
...
     }
-
-    this.startAt = this.neutral;
   }
 
-  // Match either null or undefined, but not 0
-  if (this.startAt !== null && this.startAt !== undefined) {
-    this.speed(this.startAt);
-  }
+  this.throttle(this.neutral);
```
rwaldron: This is a safety precaution. No program should ever just "start" an ESC without explicit instruction. This is the safest thing I can think of.

lib/esc.js
```diff
@@ -249,11 +235,11 @@ util.inherits(ESC, Emitter);
  * @param  {Float} speed 0...100 (full range)
  *
  * @return {ESC} instance
+ * @deprecated Will be deleted in version 1.0.0. Use throttle(us) instead.
  */
-
-ESC.prototype.speed = function(speed) {
+/* istanbul ignore next */
+ESC.prototype.speed = util.deprecate(function(speed) {
```
rwaldron: This whole function is trash. It did async work to speed up/down, which was really just papering over the fact that integers between 0-100 scales to ~18us for each step and that's awful, so let's get rid of this trash.

lib/esc.js
```diff
@@ -333,24 +313,29 @@ ESC.prototype.speed = function(speed) {
...
+ * Throttle the ESC's speed by setting the pulse
+ *
+ * @param  {Integer} throttle pwmRange[0]...pwmRange[1] (full usec range)
+ *
+ * @return {ESC} instance
+ */
+ESC.prototype.throttle = function(pulse) {
+  this.write(this.pin, Fn.constrain(pulse, this.pwmRange[0], this.pwmRange[1]));
+  return this;
+};
```
rwaldron: This is what it should've always been.

lib/esc.js
```diff
@@ -394,17 +379,7 @@ ESC.prototype.brake = function() {
...
-  var speed = this.type === "bidirectional" ? this.neutral : 0;
-
-  this.write(this.pin, Fn.fscale(speed, 0, 100, this.pwmRange[0], this.pwmRange[1]));
-
-  history.push({
-    timestamp: Date.now(),
-    speed: speed
-  });
-
+  this.write(this.pin, this.neutral);
```
rwaldron: `stop` and `brake` were basically the same thing... Let's figure out why

lib/esc.js
dtex: Are you sure about 544-2400 being the ideal range for ESC's? That range comes from the Arduino servo library. I understand some ESC's will automatically adjust to fit the range they see, but I don't know what typical ESC's expect.

rwaldron: > Are you sure about 544-2400

As a default? Sure. User code can and should use the appropriate range for whatever platform and hardware is in use.

rwaldron: Upon further investigation, I believe that this range is _not_ appropriate for ESCs and will update this patch accordingly.

lib/esc.js
dtex: I have a guess why it's like this. Some ESC's have the ability to brake and some don't (that's speculation, not fact). The default don't so they just replicate the behavior of ```stop```. That can be overridden in controller objects that do support it.

HipsterBrown:
Finally tested this while preparing for the Nodeboats workshop at the upcoming JS Conf US. 

I ran the following code to test this patch:

```js
const five = require('johnny-five');
const keypress = require('keypress');

const board = new five.Board({
  port: '/dev/tty.blobfish-DevB',
});

board.on('error', error => {
  console.error(error);
  process.exit(1);
});

board.on('ready', () => {
  let speed = 50;
  const led = new five.Led(13);
  const esc = new five.ESC({
    device: 'FORWARD_REVERSE',
    neutral: speed,
    pin: 11,
  });

  // just to make sure the program is running
  led.blink(500);

  function controller(_, key) {
    let isThrottle = false;

    if (key && key.shift) {
      if (key.name === 'up' && speed <= 100) {
        speed += 1;
        isThrottle = true;
      }

      if (key.name === 'down' && speed >= 0) {
        speed -= 1;
        isThrottle = true;
      }

      if (isThrottle) {
        console.log('Setting speed: ', speed);

        esc.throttle(
          five.Fn.scale(speed, 0, 100, esc.pwmRange[0], esc.pwmRange[1]),
        );
      }
    }
  }

  keypress(process.stdin);

  process.stdin.on('keypress', controller);
  process.stdin.setRawMode(true);
  process.stdin.resume();
});
```

It ran perfectly. 👌

I noticed a comment about updating the default PWM range (https://github.com/rwaldron/johnny-five/pull/1484#discussion_r204090707) but didn't notice a change in the PR. So that still may be pending.

dtex:
The light finally went on re: why throttle is a better method name than speed, but I think it's weird that Johnny-Five doesn't abstract away the need to pass values in microseconds. It seems like that would be in keeping with other classes in J5. 

There was a plan to migrate input values like this to a [0.0 - 1.0] or [-1.0, 1.0] range. Why not go ahead and do that here?

HipsterBrown:
So throttle or some other method would just accept a range? Or you would be able to set the range on the ESC class instantiation and use a value in that range when calling throttle()?

dtex:
No, it was a single set range for all output devices whose values do not correspond to some real-world measure. I misremembered the details though, it's not 0-1.0, it's 0%-100%.

Here's the issue: https://github.com/rwaldron/johnny-five/issues/370

rwaldron:
@dtex your intuition is right, but I think I landed on this for a few reasons: 

- `throttle()` is meant to mimic the trigger or thumb stick on an actual RC transmitter. Those physically default to whatever is "neutral" (or throttling no speed at all). 
- "throttling" _might_ be bidirectional, so what does 0% mean?
- Mapping 0-100% to μs would cause a massive loss in precision, noted in my comment above:
    > It did async work to speed up/down, which was really just papering over the fact that integers between 0-100 scales to ~18us for each step and that's awful, so let's get rid of this trash.

    That meant that motors attached to ESCs might jump 18-19μs when throttling up 1%, which is physically dangerous (as @buritica and I experienced first hand about 5 years ago when we tried to build a hovering "jellyfish" drone)

So it's not as "nice", but it's safe and accurate.

dtex:
> Mapping 0-100% to μs would cause a massive loss in precision

I didn't mean to suggest we were limited to integers. I should have written 0.000% - 100.000%. Values input in that range will be mapped to the nearest whole microsecond value before being sent to the IO plugin.

> That meant that motors attached to ESCs might jump 18-19μs when throttling up 1%

I'm on board with that. It's the same underlying principle for why J5 servo writes are in microseconds now. Servos can be positioned in fractions of degrees.

rwaldron:
@dtex 

This morning I was still wondering about this: 

> "throttling" might be bidirectional, so what does 0% mean?

... but I think I've convinced myself that it's safe to allow both: 

- `throttle(0-100%)`
- `throttle(1000-2000usec)`

buritica:
I remember very well

eg/esc-blobfish.js
```diff
@@ -0,0 +1,57 @@
...
+      }
+    }
+  }
+
+  keypress(process.stdin);
+
+  process.stdin.on("keypress", controller);
+  process.stdin.setRawMode(true);
+  process.stdin.resume();
+});
```
rwaldron: @HipsterBrown ^^^ I've been using this to test all day... it feels really good

rwaldron:
I've made a number of changes that meet the need set out by @dtex, achieve my goals of simplifying and are ready for @HipsterBrown to use in the NodeBoats workshop. I want to merge this, but I'd like @HipsterBrown to verify that his boat works properly

HipsterBrown:
This works for my nodeboat setup. 👍 LGTM

dtex:
LGTM!

I think someday ESC and Motor should be refactored to have the same surface API. It's weird to me that they are different.

rwaldron:
> I think someday ESC and Motor should be refactored to have the same surface API. It's weird to me that they are different.

I look forward to this potentially contentious debate <3

Issue #1280, opened 2017-01-23
Q: PWM resolution with PCA9685


mwittig:
Question: Is it possible to use 12-bit PWM for PCA9685 with J5? Looking at the Led and Expander code the PWM output seems to be restricted to 8-bit. Is it planned to support this feature in a future release?

dtex:
Yes, this is something we need to add. Right now there is work to add support for boards that support more than 8-bits to Firmata and that will inform what we need to do for Expander (the class that wraps the PCA-9685).

HipsterBrown:
There is an open issue around Plugin Resolution here -> https://github.com/rwaldron/johnny-five/issues/1276

mwittig:
Excellent, thank you very much for the feedback. Looking at this from the top-level I can imagine the following use cases:
* query the  expander or device (e.g. Led) about native PWM resolution and use the supported range, e.g. via a new pwm() method
* use 8-bit PWM and let the controller scale the value to the native range, e.g. when using Led.brightness if this remains unchanged

dtex:
Hi @mwittig , 

This issue is super important (arguably the most important issue we have open right now) and my closing this certainly doesn't mean that we don't care, we just haven't gotten to it yet. I'm worried about it getting lost in issues so rather than leave it languishing we have created a [Requested Features](../wiki/Requested-Features) page and added your the request for leveraging io.resolution there.

There are two other related issues that I am closing along with this one, so we definitely have mots of people interested in this. We just need a motivated contributor.

dtex:
This landed in #1495 and will be in the next release of Johnny-Five (v1.1.0?)

Issue #1276, opened 2017-01-16
Use IO Plugin .RESOLUTION.* if available


rwaldron:
I'm formulating a plan for how to incorporate this: https://github.com/firmata/firmata.js/commit/bb46a890dbe31ac7ae4122f53c24b824dcee17fd across all IO Plugins

hxlnt:
Awesome--looks like this will address the issue I opened a few weeks ago (#1275). Thanks!

As I mentioned in that issue, I have a Feather HUZZAH Nodebots/J5 [repo](https://github.com/hxlnt/feather-nodebot) that has J5 motor PWM working with a generic L9110 dual-channel H-bridge board (no FeatherWings) if it's helpful to you for testing purposes. The `index.js` has a bunch of other stuff for dashboard connectivity, so I made a trimmed-down version called `index-withoutdashboard.js` that just has the bot driving code. There's also a Wiki with setup instructions, wiring diagrams, etc. Feel free to tag me or find me on [twitter](http://twitter.com/partytimehxlnt) if you have questions about it or if I can help!

rwaldron:
This has shipped in Firmata.js

dtex:
Hi @rwaldron 

This issue is super important (arguably the most important issue we have open right now) and my closing this certainly doesn't mean that we don't care, we just haven't gotten to it yet. I'm worried about it getting lost in issues so rather than leave it languishing we have created a [Requested Features](../wiki/Requested-Features) page and added your the request for leveraging io.resolution there.

There are two other related issues that I am closing along with this one, so we definitely have mots of people interested in this. We just need a motivated contributor.

dtex:
This landed in #1495 and will be in the next release of Johnny-Five (v1.1.0?)

Issue #1275, opened 2017-01-14
Feather HUZZAH (ESP8266), generic H-bridge motor drivers, and PWM range


hxlnt:
I recently led a series of NodeBots workshops using J5 + StandardFirmataWifi with the Adafruit Feather HUZZAH boards and inexpensive dual H-bridge motor drivers (L9110, various manufacturers). I'd like to use the learnings from preparing and hosting this workshop to help improve J5's Feather support. 

In particular, I initially ran into issues with adjusting the speed and direction of DC motors due to the [ESP8266's default PWM range of 0 to 1023](http://arduino.esp8266.com/versions/1.6.5-1160-gef26c5f/doc/reference.html#analog-output) instead of 0 to 255. Passing a speed variable to the motors resulted in erratic results, and motor.rev() functions did not work regardless of what speed was passed in. In our workshop, the issue was corrected by manually changing the PWM range on the Firmata side (`analogWriteRange(255)`), but I would be interested to see if this would be worth addressing in J5.

I am happy to help write the code, but this seems like something that might have broader implications than just the Feather board + L9110 motor drivers (_e.g.,_ possible implementation of PWM range parameter in all ESP8266 boards?), and I don't have a good handle on J5's overall philosophy on changes like this one.

If it's useful to anyone in the meantime, the source code and step-by-step instructions for the Feather NodeBots workshop are [on GitHub](https://github.com/hxlnt/feather-nodebot/wiki). I'm working on making a more generic version, as this was for a specific event and has elements (competition schedule, connection to a bot motion-path visualization dashboard, etc.) that aren't broadly applicable.

Tagging @BrianGenisio, as we talked about this issue yesterday and he kindly offered his help.

monteslu:
Keep your eyes on this PR:  https://github.com/firmata/firmata.js/pull/149

Soon as it lands we can get to having j5 make use of it to get PWM resolution working.

monteslu:
also more info about it in the conversation here that could be helpful:  https://github.com/firmata/firmata.js/pull/144

rwaldron:
Hi! Sorry for the delay, I was on parental leave from January 20th until last week and I'm still trying to catch up with everything. I appreciate your patience ❤️

dtex:
Hi @hxlnt 

[...]

PR #1495, opened 2018-08-26
[WIP] Add support for PWM resolutions other than 8-bit

dtex:
Required for ESP32. Addresses #1275, #1276, and #1280

### DIFF
--- a/lib/motor.js
+++ b/lib/motor.js
@@ -646,7 +646,7 @@ Motor.prototype.setPin = function(pin, value) {
 };
 
 Motor.prototype.setPWM = function(pin, value) {
-  this.io.analogWrite(pin, value);
+  this.io.analogWrite(pin, __.map(value, 0, 255, 0, this.board.RESOLUTION.PWM));
 };
 
 Motor.prototype.speed = function(opts) {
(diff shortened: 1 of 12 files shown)


lib/motor.js
```diff
@@ -646,7 +646,7 @@ Motor.prototype.setPin = function(pin, value) {
 };
 
 Motor.prototype.setPWM = function(pin, value) {
-  this.io.analogWrite(pin, value);
+  this.io.analogWrite(pin, __.map(value, 0, 255, 0, this.board.RESOLUTION.PWM));
```
rwaldron: Can we do a pass on this file and change `__` to `Fn`? Probably in another patch?

DanHartman:
I do not understand how to access the 12 bit resolution of the PCA9685.
I suspect that I have misinterpreted what `pwmRange` does.
Given the following code, I expected to see a duty cycle on my oscilloscope that is a fraction of full on.
What I see is near full scale output that would imply that my value of 254 is almost max 8 bit resolution.
What have I done wrong?
```
const five = require('johnny-five')

five.Board().on('ready', function() {
  var led = new five.Led({
    pin: 0,
    controller: 'PCA9685',
    pwmRange: [0, 4095]
  })
  led.brightness(254)
})
```

Issue #1499, opened 2018-09-02
TypeError: keyFrames.map is not a function


arindamghoshal:
Not sure what I am doing. I am getting this error while animating servos based on the code in eg/phoenix.js. I am trying to animate 4 legs. Attached is the code file. The error gets triggered while executing bot.sleep(); [line 109 in the attached file]

TypeError: keyFrames.map is not a function
    at Servo.(anonymous function) [as @@normalize] (/home/pi/projects/quadBot/node_modules/johnny-five/lib/servo.js:346:20)
    at Servos.<anonymous> (/home/pi/projects/quadBot/node_modules/johnny-five/lib/servo.js:646:46)
    at Array.map (<anonymous>)
    at Servos.(anonymous function) [as @@normalize] (/home/pi/projects/quadBot/node_modules/johnny-five/lib/servo.js:620:22)
>> 

[index.js.zip](https://github.com/rwaldron/johnny-five/files/2343521/index.js.zip)

dtex:
Hmmm, it looks like you've properly written your code to handle just four legs so the problem doesn't jump out at me. I'll have to some hardware from the lab this afternoon to run the code (and console.log a lot).

One thing does jump out at me, though it has nothing to do with your current issue... I've never successfully controlled multiple servos at a high rate of speed over I2C. I think there is some limit as to how fast those messages can be sent, and even though the PCA9685 can receive instructions for multiple servos in one message, we are only sending one at a time from Johnny-Five.

Also, there is a newer, better way to control walking robots. It's not as well documented (there is no book to go with it), but https://github.com/dtex/tharp uses an inverse kinematics layer on top of Johnny-Five to make controlling the robot much easier (and dynamic).

I guess I really only need an Arduino to run the code with little modification. I'll see if I can replicate the error.

arindamghoshal:
Thanks @dtex. I will look into https://github.com/dtex/tharp library.

dtex:
I was able to replicate. Still investigating.

dtex:
We have a collection of a collection of servos. Your code (like mine) expects the coxa, femur, tibia collections to all be passed the same value so we should be normalizing them using the Servo normalize method instead of the Servos normalize method. For some reason they are being normalized as Servos even though we have insufficient keyFrames (which is why the error is being thrown). 

I changed the ```sleep``` object to this dirty hack and it got past the error:

````js
const sleep = {
        target: bot.joints,
        duration: 500,
[...]

PR #1503, opened 2018-09-11
Fix for #1499

dtex:
If an array of keyFrames are passed to Servos, those keyFrames should be used as values for the whole collection and should not be normalized as a keyFrameSet.


dtex:
Also fixed a bug where we were not converting ```degrees``` to ```value``` on collections that shared keyFrames.

PR #1507, opened 2018-09-30
Inverts intervals in Sensor.scale when 'low' paramater is larger than 'high'

### LINKED ISSUE #878: When the lower value of the range is greater than the upper scale does not work
When I try to use the scale with the lower value greater than the upper scale does not work, this.value is returning 0

```
rotary.scale(180,0).on("change",function(){
    console.log(this.value);
});
```

HaraldNordgren:
Fixes #878.

### DIFF
--- a/lib/sensor.js
+++ b/lib/sensor.js
@@ -351,6 +351,12 @@ Sensor.prototype.disable = function() {
 Sensor.prototype.scale = function(low, high) {
   this.isScaled = true;
 
+  if (low > high) {
+    var tmp = low;
+    low = high;
+    high = tmp;
+  }
+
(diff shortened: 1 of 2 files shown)


HaraldNordgren:
> Imagine a photoresistor and an LED hooked up to a board. As the ambient light gets dimmer, we want the LED to get brighter so we use scaleTo

If this is the case, then I guess https://github.com/rwaldron/johnny-five/issues/878 should be closed just on the basis of that.

dtex:
Agreed. Thanks so much for raising this so we can get that issue closed. Sometimes we forget to revisit things that have been open for so long.

I hope this doesn't discourage you from making contributions to J5 in the future.

PR #1510, opened 2018-10-12
Changed import of nunchuck library to classic-controller

KatieK2:
*Probably* just a c&p error from when docs were created. I don't have the ability to test this right now.


dtex:
Oh, and you know about [generating the docs](https://github.com/rwaldron/johnny-five/pull/1510) right?

PR #1514, opened 2018-10-28
Create john.c

jihadijohn2:


dtex:
This appears to be a mistake

PR #1515, opened 2018-10-28
Do not require SERIAL_PORT_IDs

dtex:
Addresses rwaldron/tessel-io#23

### DIFF
--- a/lib/gps.js
+++ b/lib/gps.js
@@ -142,8 +142,6 @@ function GPS(opts) {
     this, opts = Board.Options(opts)
   );
 
-
-
   // Get user values for breakout, receiver and chip
   breakout = opts.breakout || {};
   receiver = opts.receiver;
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "johnny-five",
   "description": "The JavaScript Robotics and Hardware Programming Framework. Use with: Arduino (all models), Electric Imp, Beagle Bone, Intel Galileo & Edison, Linino One, Pinoccio, pcDuino3, Raspberry Pi, Particle/Spark Core & Photon, Tessel 2, TI Launchpad and more!",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
rwaldron: Can you remove this from the commit?

dtex: I just removed the commit altogether. That was the only change.

Issue #1490, opened 2018-08-05
Add pre-commit


dtex:
This has been helpful for me in some other projects:

https://github.com/observing/pre-commit

It allows you to specify that tests should be run anytime git commit is called and before the code is actually committed. Pretty slick. I imagine we could specify the full test suite be run.

Any interest in having this added to the Johnny-Five repo?

AlbertHambardzumyan:
Perhaps, would be nice to have a lint check script in the pre-commit list. 
Any thoughts?

PR #1541, opened 2018-12-20
Added pre-commit. This is for issue #1490

matt611:


matt611:
See pre-commit docs here: https://github.com/observing/pre-commit

dtex:
Any idea why the operator on the version ranges in package-lock.json changed?

dtex:
Here's the scoop https://github.com/npm/npm/issues/20434

There will be a lot of churn in the lock file on commits between users running npm v5 and v6.

matt611:
Yeah, I'm not sure what the correct answer for the large package-lock.json diff is.  It seems like this will be the way going forward, but I'm happy to redo this commit with npm5 instead if that makes things easier.

Issue #831, opened 2015-06-18
PCF8574 does not work as input


rodrigoadachi:
``` js
board.on("ready", function() {
  var virtual = new five.Board.Virtual(
    new five.Expander({ controller: "PCF8574",  address: 0x20 })
  );
  var button = new five.Button({
    board: virtual,
    pin: 7, 
  });
  button.on("press", function() {
    console.log('button on');
  });
});
```

Did tests with outputs and everything works ok, however when declare inputs with outputs and inputs do not work only.

rwaldron:
Thanks for the report, I will take a look at this today

rwaldron:
Can you post a picture of your exact wiring? Thanks!

rwaldron:
@rodrigoadachi can you try this with the johnny-five @ master?

rodrigoadachi:
![pcf8574_bb](https://cloud.githubusercontent.com/assets/6013579/8237640/38722062-15c7-11e5-9eb1-c23ef19ced98.png)

rodrigoadachi:
Como assim johnny-five @ master?

rodrigoadachi:
var   , five = require("johnny-five");

  var board = new five.Board({ port: "/dev/ttyUSB0" });

  board.on("ready", function() {
    var virtual = new five.Board.Virtual(
      new five.Expander({
        controller: "PCF8575",
        address: 0x20
      })
    );

    var led = new five.Led({
      board: virtual,
      pin: 0, 
    });

    var button = new five.Button({
      board: virtual,
      pin: 7, 
    });

  button.on("press", function() {
    console.log('pulsador1 on');
    led.on();
  });

  button.on("release", function() {
    console.log('pulsador1 off');
    led.off();
  });

});

rwaldron:
I just pushed the latest changes to npm: johnny-five@0.8.80

I confirmed the bug you experienced, but will also try your code above

rwaldron:
The code you have above uses `PCF8575`, but this was filed about `PCF8574` and that's the chip in the diagram—these are not interchangeable. Which one should I test with?

rodrigoadachi:
Ohh, sorry, 'cause I tried using the IC PCF8575 PCF8574, but the fact is the PCF8574.

rwaldron:
Ok, thanks. I'm still seeing the issue with your code example (updated to use `PCF8574`) so I'll continue working on this.

rodrigoadachi:
[...]

PR #1544, opened 2019-01-06
Feature proposal: Possibility to set default pin modes and pin values to expanders

derdanielmoin:
Hi Johnny-five team,

for a project I needed a possibility to change the initial pin mode and pin value for the PCF8574. The PCF8574 sets default all pins to INPUT and HIGH (from data sheet). In johnny five all pins are set in the initialisation to OUTPUT and LOW. Additionally the setMode function also sets the pin to LOW. Because the IO Pins from the PCF8574 are connected to another microcontroller these short changes cause problems.

So my idea is to add default parameters to the Expander function to add the possibility to handle the initial problems. But I also see that this problem is not as easy because keeping things consistent is related (in the broadest sense) to the arduino firmata issue https://github.com/firmata/arduino/issues/308 . 

For the second problem one possible solution is modifying the setMode function, but this is also a big consistency issue. Another idea is to add additional setMode-functions or an additional parameter to setMode to handle this.

For a better understanding I opened this as a pull request with my hacked variant of johnny five.

best regards,

Daniel

### DIFF
--- a/lib/expander.js
+++ b/lib/expander.js
@@ -382,8 +382,10 @@ var Controllers = {
             analogChannel: 127
           });
 
-          this.pinMode(i, this.MODES.OUTPUT);
-          this.digitalWrite(i, this.LOW);
+          var mode = opts.defaultPinMode || this.MODES.OUTPUT;
+          var value = opts.defaultPinValue || this.LOW;
+
+          this.initializePin(i, mode, value);
         }
 
         this.name = "PCF8574";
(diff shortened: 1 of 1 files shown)


dtex:
1. Possibly related #831

2. Should we set the default pin state in J5 to match the default pin state per the data-sheet and expect users to override that? Not sure if this would cause any problems in existing clients of the J5 PCF8574 Expander but having J5 state match the real world by default seems like the right thing to do.

3. You bring up a good point by mentioning firmata/arduino#308. Whatever solution is designed for this, should be a pattern match for what needs to be done to solve the Aduino issue. That's a much more difficult problem.

PR #1564, opened 2019-04-01
Update example after raspi-io v10.0.0

whoan:
https://github.com/nebrius/raspi-io/blob/master/CHANGELOG.md#1000-2019-2-21


rwaldron:
This file is generated. Please update the JS file: https://github.com/rwaldron/johnny-five/blob/master/eg/raspi-io.js

Thanks!!

whoan:
Done @rwaldron.

PR #1594, opened 2019-08-13
Update ESC examples to new API methods

HipsterBrown:
Found some example programs that I missed when reviewing #1484.


rwaldron:
Thanks for updating, I had actually just removed them from the main list because I was too lazy to figure out how to make them work the same way. Thanks!!

Issue #1588, opened 2019-07-24
mini summit 2019 ?


monteslu:
I'll be getting to the conf Aug 11.  Could be there a few hours before the opening party.

Any interest in a get together beforehand to go over last years goals and progress?

@rwaldron, @lyzadanger, @nebrius, @noopkat, @dtex, @ajfisher, @lynnaloo, @sandeepmistry, @HipsterBrown, @reconbot @soundanalogous @julianduque @KatieK2 @ghtomcat @nodebotanist @rockbot

dtex:
I'm not getting in until late on the 11th, but I could look at changing my flight.

dtex:
For those of y'all that weren't there last year: 
* [Day 1](https://github.com/rwaldron/johnny-five/issues/1463#issuecomment-414544943)
* [Day 2](https://github.com/rwaldron/johnny-five/issues/1463#issuecomment-415898310)

monteslu:
brunch on the way out like last time works too.  I'm not in a hurry to leave San Diego to head back to Phoenix :)

ajfisher:
I'm flying in on the Sunday but will likely need to do NodeBoats prep as we
have to make up the kits etc ready for Day 2. I am pretty flexi on the way
out as just heading back to LA before going back to AU on Friday night.




--
Andrew Fisher
http://about.me/ajfisher/bio

reconbot:
Late on the Aug 11th (Sunday after 3pm PST) might work for me but Sara has boats and I'll have my kid (which should be fine). We leave at noon on Thursday the 15th so that's tough. Happy to meet during the week. Lets collect a little more input and then start a doodle!

Excited to see everyone! 🙌

lyzadanger:
I'm free on Sunday afternoon/evening, so open to various options here...

noopkat:
I'm still unsure if I'll be able to make it to JsConf, but I'll attend if I can!

nebrius:
I'm flying in Sunday and land at 1pm PDT, getting to the hotel late afternoon-ish, and I'll be there for brunch on Thursday. I'm happy to meet once I get there/before I leave.

reconbot:
4pm break tomorrow?

noopkat:
I’m down

reconbot:
How about lunch? We'll have more time.

(grab lunch and then) 12:10PM at the office/storage room

nebrius:
[...]

PR #1595, opened 2019-08-15
Docs: update ESC examples with newer javascript syntax

HipsterBrown:
Updating the ESC examples that I've been looking at a bunch while preparing for Nodeboats at JSConf. Related to task discussed at recent collaborators mini-summit: https://github.com/rwaldron/johnny-five/issues/1588#issuecomment-521467213

The double-quote -> single-quote conversions were automated by my editor, so happy to change those back if needed.

### DIFF
--- a/docs/esc-bidirectional-forward-reverse.md
+++ b/docs/esc-bidirectional-forward-reverse.md
@@ -29,26 +29,24 @@ node eg/esc-bidirectional-forward-reverse.js
 
 
 ```javascript
-var five = require("../");
-var board = new five.Board();
+const {Board, Button, ESC, Sensor} = require('johnny-five');
+const board = new Board();
 
-board.on("ready", function() {
-  var start = Date.now();
-  var esc = new five.ESC({
-    device: "FORWARD_REVERSE",
+board.on('ready', () => {
+  const start = Date.now();
+  const esc = new ESC({
+    device: 'FORWARD_REVERSE',
     neutral: 50,
-    pin: 11
+    pin: 11,
   });
-  var throttle = new five.Sensor("A0");
-  var brake = new five.Button(4);
+  const throttle = new Sensor('A0');
+  const brake = new Button(4);
 
-  brake.on("press", function() {
-    esc.brake();
-  });
+  brake.on('press', esc.brake);
 
-  throttle.scale(0, 100).on("change", function() {
+  throttle.scale(0, 100).on('change', () => {
     // 2 Seconds for arming.
-    if (Date.now() - start < 2e3) {
+    if (Date.now() - start < 2000) {
       return;
     }
 
--- a/docs/esc-bidirectional.md
+++ b/docs/esc-bidirectional.md
@@ -29,10 +29,10 @@ node eg/esc-bidirectional.js
 
 
 ```javascript
-const {Board, Button, ESC, Sensor} = require("../");
+const { Board, Button, ESC, Sensor } = require("johnny-five");
 const board = new Board();
 
-board.on("ready", function() {
+board.on("ready", () => {
   const esc = new ESC({
     device: "FORWARD_REVERSE",
     pin: 11
@@ -40,9 +40,7 @@ board.on("ready", function() {
   const throttle = new Sensor("A0");
   const brake = new Button(4);
 
-  brake.on("press", () => {
-    esc.brake();
-  });
+  brake.on("press", esc.brake);
 
   throttle.on("change", () => {
     esc.throttle(throttle.scaleTo(esc.pwmRange));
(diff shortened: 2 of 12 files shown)


dtex:
Nick,

The linter checks fail with the double quote so those will have to be fixed. Also, the doc files don't seem to match what was expected. Did you run ```grunt examples``` after editing the files in ```/docs``` or did you manually edit the files in ```/docs```?

HipsterBrown:
@dtex I'll fix that up tonight. :+1:

HipsterBrown:
@dtex All updated and passing CI. Thanks for the review. 😃

docs/esc-bidirectional-forward-reverse.md
```diff
@@ -29,26 +29,24 @@ node eg/esc-bidirectional-forward-reverse.js
...
-var five = require("../");
-var board = new five.Board();
+const {Board, Button, ESC, Sensor} = require('johnny-five');
+const board = new Board();
 
-board.on("ready", function() {
-  var start = Date.now();
-  var esc = new five.ESC({
-    device: "FORWARD_REVERSE",
+board.on('ready', () => {
```
reconbot: nit: quotes should/could match the codebase?

docs/esc-bidirectional.md
```diff
@@ -29,20 +29,18 @@ node eg/esc-bidirectional.js
...
     device: "FORWARD_REVERSE",
     pin: 11
   });
   const throttle = new Sensor("A0");
   const brake = new Button(4);
 
-  brake.on("press", () => {
-    esc.brake();
-  });
+  brake.on("press", esc.brake);
```
reconbot: confirmed `esc.brake()` ignores all arguments and `button!press` has no data

docs/esc-dualshock.md
```diff
@@ -29,53 +29,52 @@ node eg/esc-dualshock.js
 
 
 ```javascript
-var five = require("../lib/johnny-five.js");
-var dualShock = require("dualshock-controller");
+const {Board, ESC, Fn} = require('johnny-five');
```
reconbot: What should we be doing here? It's not consistent.

dtex: Destructuring for sure.

reconbot: Yeah but the relative path or the package?

dtex: Oh, gotchya. So either works, but without the relative path does it seem too magical?

Was there ever a time when node wouldn't go up the directory structure looking for node_modules? Maybe that's why it is like this in all the old examples?

reconbot: Well the package name wont work if you try to run it in a clone of j5, but if you're copying and pasting the example into your own project then the relative path wont work because you've installed johnny-five as a dep.

dtex: Oh right, it's not in node modules if you've cloned the repo. It should be the relative path for sure.

HipsterBrown: So relative paths for all example code?

docs/esc-bidirectional-forward-reverse.md
HipsterBrown: Based on @dtex's recent comment (https://github.com/rwaldron/johnny-five/pull/1601#issuecomment-522651218), I'm going to keep this as is.

docs/esc-dualshock.md
HipsterBrown: So `grunt examples` automatically converts the relative imports in the example code (`const {Board} = require('../lib/johnny-five')`) to the package import (`const {Board} = require('johnny-five')`). I fixed the example ESC code that didn't use the relative import so it's consistent in the examples.

reconbot: @rwaldron always two steps ahead

rwaldron: Please keep the relative path so that `node eg/foo.js` continues to work as expected. johnny-five.io's build process will fix all them for you.

dtex: Witchcraft!

HipsterBrown:
Is this good to merge now?

rwaldron:
Excellent work

rwaldron:
> The double-quote -> single-quote conversions 

I will fix these

rwaldron:
> I will fix these

Looks like you already did?

docs/esc-bidirectional.md
rwaldron: If only JavaScript was the simple ;) https://github.com/rwaldron/johnny-five/pull/1602#discussion_r316386618

reconbot: The comment won’t load for me on mobile but we lose the context for `this` so this won’t work.

PR #1598, opened 2019-08-17
Examples: update led syntax

ryanjgill:


dtex:
Cool! A couple of comments based on the esc examples Rick created last week:

Use destructuring on the J5 import to grab just the classes you need rather than grabbing the parent ```five``` object.

Let the board ready handler be bound to board. It is the most likely target for ```this``` throughout a J5 program (the REPL, accessing the io object, etc ). In other words, no fat arrow on ```board.on("ready", function() { ... });```. 

I was about to say something about ```grunt examples``` but I see you just updated your PR so I'm guessing you saw the error.

````js
const {Board, LED} = require("../lib/johnny-five.js");
const board = new Board();

  const led = new Led(13);

  // This will grant access to the led instance
  // from within the REPL that's created when
  // running this program.
  this.repl.inject({
    led
  });

  led.blink();
});
````

dtex:
LGTM!

rwaldron:
> Let the board ready handler be bound to board. It is the most likely target for this throughout a J5 program (the REPL, accessing the io object, etc ). In other words, no fat arrow on board.on("ready", function() { ... });.

If someone wants to do the work of replacing all the occurrences of `this` with `board` (where appropriate), we could move the examples to use an arrow in the `"ready"` handler.

PR #1600, opened 2019-08-17
Examples: update led syntax

ryanjgill:


dtex:
FYI, Sometimes the tests that are run over a period of time fail on Travis and you just have to kick them off again. Nothing you did. Let's see how it does on this next try.

ryanjgill:
I'm going to push out the rest of the led examples to this branch so please hold until I give it the thumbs up.

docs/led-array-controller.md
```diff
@@ -31,20 +31,20 @@ node eg/led-array-controller.js
...
 
 board.on("ready", function() {
-  var leds = new five.Leds([2, 3, 4, 5, 6]);
-  var pot = new five.Sensor("A0");
+  const leds = new Leds([2, 3, 4, 5, 6]);
+  const pot = new Sensor("A0");
 
-  pot.scale([-1, 4]).on("change", function() {
-    var lastIndex = Math.round(this.value);
+  pot.scale([-1, 4]).on("change", () => {
```
reconbot: do we want to use the data passed into the change event or read off the objects themselves? I've always preferred the former (data passed in) because then I need to know less about my callback's context. I don't know if it's easier to learn but it's probably 6 one 1/2 dozen the other.

ryanjgill: So I was thinking that originally but in this case I noticed we only pass the `rawValue` of the Sensor so it confused me as I wasn't able to get the scaled value as far as I could tell without using the object itself.

dtex: I think this is one of the things driving @ajfisher's audit of parameters passed to callbacks. For now, I think we just roll with it and update the docs as best we can. We will hammer out the callback parameters in another effort, especially since no matter what we do it will be a breaking change.

eg/led-demo-sequence.js
```diff
@@ -76,7 +76,7 @@ function execute(step) {
   });
 }
 
-board.on("ready", () => {
+board.on("ready", function () {
```
rwaldron: Go ahead and change `this` to `board`, then you can use arrows throughout :)

rwaldron: Go ahead and change this to board, then you can use arrows throughout :)

docs/led-array-controller.md
```diff
@@ -31,20 +31,20 @@ node eg/led-array-controller.js
...
+    const lastIndex = Math.round(pot.value);
```
rwaldron: All uses of `sensor.scale(...)` should be removed and the code updated to use the `scaleTo()` method. So for example, this code would become: 

```js
pot.on("change", () => {
   const lastIndex = Math.round(pot.scaleTo(-1, 4));
   // ...
```

docs/led-demo-sequence.md
```diff
@@ -82,15 +82,15 @@ var demoSequence = [{
...
   // Grab everything we need for this step
-  var method = demoSequence[step].method;
-  var args = demoSequence[step].args;
-  var duration = demoSequence[step].duration || 3000;
+  const method = demoSequence[step].method;
+  const args = demoSequence[step].args;
+  const duration = demoSequence[step].duration || 3000;
 
   // Just print out what we're executing
   console.log("led." + method + "(" + (args ? args.join() : "") + ")");
```
rwaldron: Eventually I want to clean up this stuff as well, but you've already done a ton of work on these examples and I don't want to keep pushing back (so you can move on to another section)

Issue #1596, opened 2019-08-15
Update examples / documentation


dtex:
**Destructure modules requires, using relative paths**
````js
const {Board, GPS} = require("../lib/johnny-five.js");
````

**Switch to const and let where appropriate (consider scope and whether the values changes)**
````js
const board = new Board();
````

**Use arrow functions on callbacks, replacing uses of ```this``` with ```board``` or the device instance wherever necessary***
````js
gps.on("change", position => {
    console.log("position");
    console.log("  latitude   : ", position.latitude);
    console.log("  longitude  : ", position.longitude);
    console.log("--------------------------------------");
  });
````

**If an arrow function is a single statement**
````js
gps.on("change", () => console.log("Data Received"));
````

**If an arrow function is a function call with no parameters, and is NOT a callback for an event listener**
````js
board.wait(1000, motor.stop);
````

**Shorthand Properties**
````js
const bar = "bar";
const foo = {
   read() {
      //
   },
   bar
}
````

Here is a recent example:

https://github.com/rwaldron/johnny-five/blob/master/eg/esc-keypress.js

- [x] accelerometer.js (@rwaldron)
- [x] altimeter.js
- [ ] barometer.js
- [ ] board.js
- [ ] board.options.js
- [ ] board.pins.js
- [ ] button.js (@monteslu)
- [ ] color.js
- [x] compass.js
- [x] esc.js (@HipsterBrown)
- [ ] evshield.js
- [ ] expander.js
- [ ] fn.js
- [x] gps.js (@dtex)
- [ ] gripper.js
- [ ] gyro.js
- [ ] hygrometer.js
- [ ] imu.js
- [ ] johnny-five.js
- [ ] joystick.js
- [ ] keypad.js
- [ ] lcd-chars.js
- [ ] lcd.js
- [x] light.js (@reconbot)
- [ ] motion.js
- [x] motor.js (@dtex)
- [ ] orientation.js
- [ ] piezo.js
- [ ] pin.js
- [ ] ping.js
- [x] proximity.js
- [ ] reflectancearray.js
- [ ] relay.js
- [ ] repl.js
- [ ] sensor.js (@ajfisher)
- [x] servo.js (@dtex)
- [x] shiftregister.js
- [x] sleep.js
- [x] sonar.js (@dtex)
- [x] stepper.js (@dtex)
- [x] switch.js (@dtex)
- [x] thermometer.js (@dtex)
- [x] digits.js (@dtex)
[...]

PR #1601, opened 2019-08-18
Examples: update light syntax

reconbot:
This updates the `light`  examples to es6. I used eslint to do most of the work and manually removed binds and a few other things.

Works #1596

### DIFF
--- a/docs/grove-light-sensor-edison.md
+++ b/docs/grove-light-sensor-edison.md
@@ -20,17 +20,17 @@ node eg/grove-light-sensor-edison.js
 
 
 ```javascript
-var five = require("johnny-five");
-var Edison = require("edison-io");
-var board = new five.Board({
+const five = require("johnny-five");
+const Edison = require("edison-io");
+const board = new five.Board({
   io: new Edison()
 });
 
-board.on("ready", function() {
+board.on("ready", () => {
 
   // Plug the Grove TSL2561 Light sensor module
   // into an I2C jack
-  var light = new five.Light({
+  const light = new five.Light({
     controller: "TSL2561"
   });
 
--- a/docs/light-ambient-BH1750.md
+++ b/docs/light-ambient-BH1750.md
@@ -29,16 +29,16 @@ node eg/light-ambient-BH1750.js
 
 
 ```javascript
-var five = require("johnny-five");
-var board = new five.Board();
+const { Board, Light } = require("johnny-five");
+const board = new Board();
 
-board.on("ready", function() {
-  var light = new five.Light({
+board.on("ready", () => {
+  const light = new Light({
     controller: "BH1750",
   });
 
-  light.on("data", function() {
-    console.log("Lux: ", this.lux);
+  light.on("data", (data) => {
+    console.log("Lux: ", data.lux);
   });
 });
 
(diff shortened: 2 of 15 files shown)


HipsterBrown:
@reconbot Can you also destructure the `five` object to take the classes needed by the example? 

```js
const {Board, Led} = require('johnny-five');
```

dtex:
I think we should stash the changes to lib/light.js and test/light.js. Those should probably accompany a functional change or bug fix.

Also, using an arrow function for the ```board.on("ready")``` handler leaves us in the global scope. I suggested to @HipsterBrown that we scope to board because I believe that is the most likely target of ```this``` in a J5 program.

Either way, let's get input from @rwaldron before we change anything else.

dtex:
@rwaldron is okay with arrow functions for the board.on("ready") handler. We just have to replace ```this``` with ```board``` wherever it's used.

lib/light.js
```diff
@@ -1,32 +1,32 @@
...
+const Board = require("./board");
+const EVS = require("./evshield");
+const within = require("./mixins/within");
+const Fn = require("./fn");
+const Emitter = require("events").EventEmitter;
+const util = require("util");
+const priv = new Map();
+// const int16 = Fn.int16;
+const uint16 = Fn.uint16;
+const toFixed = Fn.toFixed;
```
dtex: Could we destructure these (and fn.scale) in the fn require above?

lib/light.js
```diff
@@ -1,32 +1,32 @@
...
-var toFixed = Fn.toFixed;
-
-var Controllers = {
+const Board = require("./board");
+const EVS = require("./evshield");
+const within = require("./mixins/within");
+const Fn = require("./fn");
+const Emitter = require("events").EventEmitter;
+const util = require("util");
+const priv = new Map();
```
dtex: Can a Map be a const?

lib/light.js
```diff
@@ -451,11 +451,11 @@ function Light(opts) {
     return new Light(opts);
   }
 
-  var controller = null;
-  var state = {};
-  var raw = 0;
-  var last = 0;
-  var freq = opts.freq || 25;
+  let controller = null;
+  const state = {};
```
dtex: Could we just pass an empty object in the call to priv.set below?

lib/light.js
```diff
@@ -535,7 +535,7 @@ Object.assign(Light.prototype, within);
 
 
 /* istanbul ignore else */
-if (!!process.env.IS_TEST_MODE) {
+if (process.env.IS_TEST_MODE) {
```
dtex: I'm guessing there is some coercion reasoning behind the !! Do you know what it is/was?

lib/light.js
HipsterBrown: Yes, because you're not re-assigning the variable when updating the Map, same with Arrays and using `push` or `pop`.

dtex: TIL ⬆️

dtex:
> Should I pull out the code upgrades in the light class?

I think they should at the very least be in a different commit.

lib/light.js
reconbot: Nice to give the If statement a Boolean? This was an auto fix eslint trying to show us it’s exactly the same.

rwaldron:
I'm with @dtex on this: 

> I think we should stash the changes to lib/light.js and test/light.js.

lib/light.js
rwaldron: Seems like something I might do for a reason, but I can't remember, so throw it out.

lib/light.js
reconbot: Addressed in #1604

PR #1603, opened 2019-08-20
Examples: update Accelerometer programs syntax

rwaldron:


docs/accelerometer-adxl335.md
```diff
@@ -29,25 +29,25 @@ node eg/accelerometer-adxl335.js
...
-    console.log("  y            : ", this.y);
-    console.log("  z            : ", this.z);
-    console.log("  pitch        : ", this.pitch);
-    console.log("  roll         : ", this.roll);
-    console.log("  acceleration : ", this.acceleration);
-    console.log("  inclination  : ", this.inclination);
-    console.log("  orientation  : ", this.orientation);
+    console.log("  x            : ", accelerometer.x);
+    console.log("  y            : ", accelerometer.y);
+    console.log("  z            : ", accelerometer.z);
```
dtex: Change events pass a single object {x, y, z} as a parameter. Could we illustrate that here with:

````js
accelerometer.on("change", raw => {
    console.log(`  x            : ${raw.x}`);
    console.log(`  y            : ${raw.y}`);
    console.log(`  z            : ${raw.z}`);
    // And then switch to accessing properties on the instance
    console.log(`  x            : ${accelerometer.x}`);
    ... etc ...
});
````
?

Comment (rwaldron) [reply docs/accelerometer-adxl335.md]: The purpose of these baseline examples is to be able to see the consistent output across all instances, of all instance properties, regardless of controller/driver. If you want to make alternate versions of these, which looked like: 

```js
const {Accelerometer, Board} = require("../lib/johnny-five.js");
const board = new Board();

board.on("ready", () => {
  const accelerometer = new Accelerometer({
    controller: "ADXL335",
    pins: ["A0", "A1", "A2"]
  });

  accelerometer.on("change", data => {
    console.log("accelerometer");
    console.log("  x            : ", data.x);
    console.log("  y            : ", data.y);
    console.log("  z            : ", data.z);
  });
});
```

...And are named `accelerometer-adxl335-change-event-data.js` and co-exist with the examples we currently have, then by all means!

PR #1604, opened 2019-08-20
Update light class to es6

reconbot:
This updates the `light` class to es6. I used eslint to do most of the work and manually removed binds and a few other things.

This doesn’t move to actually using ES6 Classes. Should we do that?

(moved from #1601)

### DIFF
--- a/lib/light.js
+++ b/lib/light.js
@@ -1,32 +1,29 @@
-var Board = require("./board");
-var EVS = require("./evshield");
-var within = require("./mixins/within");
-var Fn = require("./fn");
-var Emitter = require("events").EventEmitter;
-var util = require("util");
-var priv = new Map();
-// var int16 = Fn.int16;
-var uint16 = Fn.uint16;
-var toFixed = Fn.toFixed;
+const Board = require("./board");
+const EVS = require("./evshield");
+const within = require("./mixins/within");
+const { uint16, toFixed, scale } = require("./fn");
+const Emitter = require("events");
+const util = require("util");
+const priv = new Map();
 
-var Controllers = {
+const Controllers = {
   DEFAULT: {
     initialize: {
-      value: function(opts, dataHandler) {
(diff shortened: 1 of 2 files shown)


rwaldron:
> This doesn’t move to actually using ES6 Classes. Should we do that?

I don't think that's a straight forward move, because our classes essentially have multiple super classes and I want to think through a strategy instead of rushing into upgrading when there isn't anything necessarily broken.

lib/light.js
```diff
@@ -1,32 +1,29 @@
...
-// var int16 = Fn.int16;
-var uint16 = Fn.uint16;
-var toFixed = Fn.toFixed;
-
-var Controllers = {
+const Board = require("./board");
+const EVS = require("./evshield");
+const within = require("./mixins/within");
+const { uint16, toFixed, scale } = require("./fn");
+const { EventEmitter: Emitter } = require("events");
```
rwaldron: This isn't necessary, `require("events") === require("events").EventEmitter`, so you can write `const Emitter = require("events");`

lib/light.js
```diff
@@ -94,9 +91,9 @@ var Controllers = {
...
     initialize: {
-      value: function(opts, dataHandler) {
-        var address = opts.address || 0x39;
-        var command = function(byte) {
+      value(opts, dataHandler) {
+        const address = opts.address || 0x39;
+        const command = function(byte) {
           // byte | 0b10000000;
           return byte | 0x80;
         };
```
rwaldron:
```js
const command = byte => byte | 0b10000000;
```

lib/light.js
reconbot: Sure. This is closer to the es modules syntax if we're eventually going that way.

lib/light.js
reconbot: 👍

rwaldron:
Merged

PR #1607, opened 2019-08-22
Update Switch examples

dtex:


dtex:
I may need to remove switch.js from programs.json. This runs counter to 08ba5b7f376db1af1fec8ea4fbc6fe4104fd3862

PR #1609, opened 2019-08-23
Sonar examples

dtex:
I'm assuming these are removed from programs.json because Sonar has been dropped in favor of proximity.

### DIFF
--- a/eg/sonar.js
+++ b/eg/sonar.js
@@ -1,12 +1,10 @@
-var five = require("../lib/johnny-five.js"),
-  board, sonar;
+const {Board, Sonar} = require("../lib/johnny-five.js");
+const board = new Board();
 
-board = new five.Board();
-
-board.on("ready", function() {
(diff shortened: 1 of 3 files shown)


rwaldron:
> I'm assuming these are removed from programs.json because Sonar has been dropped in favor of proximity.

That's correct.

eg/sonar.js
```diff
@@ -30,21 +28,21 @@ board.on("ready", function() {
...
-    console.log("data", "Object is " + this.inches + "inches away");
+    console.log("data", "Object is " + sonar.inches + "inches away");
   });
 
   //
   // "change" fired when distance reading changes
   //
   sonar.on("change", function() {
-    console.log("change", "Object is " + this.inches + "inches away");
+    console.log("change", "Object is " + sonar.inches + "inches away");
```
rwaldron: I'll fix these

PR #1611, opened 2019-08-23
Update Thermometer examples

dtex:


tpl/programs.json
```diff
@@ -1222,6 +1222,10 @@
         "file": "temperature-ds18b20.js",
         "title": "Thermometer - DS18B20"
       },
+      {
+        "file": "temperature-dual-ds18b20.js",
+        "title": "Thermometer - Dual DS18B20"
```
rwaldron: Did you build this circuit and test the program? (I should've been asking this for all the ones you've added)

eg/temperature-mpu6050.js
```diff
@@ -1,17 +1,17 @@
...
-    console.log("  celsius      : ", this.celsius);
-    console.log("  fahrenheit   : ", this.fahrenheit);
-    console.log("  kelvin       : ", this.kelvin);
-    console.log("--------------------------------------");
+  thermometer.on("change", () => {
+    console.log(`Thermometer
+  celsius      : ${thermometer.celsius}
+  fahrenheit   : ${thermometer.fahrenheit}
+  kelvin       : ${thermometer.kelvin}
+--------------------------------------`);
```
rwaldron: Next update should include using `common-tags` module so these can be: 

```js
  thermometer.on("change", () => {
    console.log(
      st.stripIndent`Thermometer
        celsius      : ${thermometer.celsius}
        fahrenheit   : ${thermometer.fahrenheit}
        kelvin       : ${thermometer.kelvin}
        --------------------------------------`
    );
  });

```
for its `ct.stripIndent&#96;...&#96;` template tag

rwaldron: But now we have some that were left using `console.log(multi, arg, style)` and some with template strings :|

rwaldron:
For the actual "change" event handlers, the first big change I'd like to do is make them all identical: 


```js
  thermometer.on("change", () => {
    console.log("Thermometer");
    console.log("  celsius      : ", this.celsius);
    console.log("  fahrenheit   : ", this.fahrenheit);
    console.log("  kelvin       : ", this.kelvin);
  });
```

Once they are all exactly the same in this regard, then we can talk about what to do with this part: 

```js
    console.log("Thermometer");
    console.log("  celsius      : ", this.celsius);
    console.log("  fahrenheit   : ", this.fahrenheit);
    console.log("  kelvin       : ", this.kelvin);
```

eg/temperature-mpu6050.js
dtex: Oh, I thought I got the multi arg ones. I'll look again and I could go ahead and add common-tags to this PR.

dtex:
I don't understand [this comment](https://github.com/rwaldron/johnny-five/pull/1611#issuecomment-524386220)

tpl/programs.json
dtex: No, but just to be clear, and I didn't add the JS file. I just added it to programs.json so the md would be generated.

rwaldron: @dtex I think the reason I had removed it from the list is because I wasn't positive that it worked correctly and reliably (but I can't remember the specifics)

rwaldron:
> I don't understand [this comment](https://github.com/rwaldron/johnny-five/pull/1611#issuecomment-524386220)

That's ok, what I will do is make the change I want for the path forward, which might be a better way to illustrate what I'd like to do. :)

rwaldron:
Merged into master!

rwaldron:
@dtex this is the "transitional" logging style I'd like to make consistent across all sensor examples: https://github.com/rwaldron/johnny-five/commit/9dccdc3ed615d1feb69b6d5ef617e87344cf888d

PR #1621, opened 2019-09-10
fix: variable name

ccarruitero:
This fix current build


dtex:
Thanks for the quick catch. The reason this failed CI checks is the docs need to be updated to match the changes you made to the example file. The best way to do this is to run ```grunt examples``` in the johnny-five directory.

Issue #1624, opened 2019-09-14
TB6612 Motors 


5ocjal:
Hi guys, got a question. If anyone tried to run this Sunfounder Picar-V car with johhny-five?
While, servos work just fine, motors don't.

I was made to run one motor with this code:

```
const {Board, Motors} = require('johnny-five');
const raspio = require('raspi-io').RaspiIO;
const invertPWM = true;

const board = new Board({
    io: new raspio()
});

board.on("ready", function() {

  var tb6612 = new Motors([
    { controller: "PCA9685", pins: { dir: 4, cdir: 5, pwm: 6 } },

  ]);
  tb6612.forward(255);
  board.wait(3000, ()=> tb6612.stop);
});
```

Unfortunately, when I execute this code, only the right motor starts and it alfo seems that this is the result of the controller works. What's more, after 3 seconds it does not stop, but it can work indefinitely, even if I stop the script with 'ctrl + c'. 

I tried to use the example that HipsterBrown linked here, but I couldn't start the motors anyway.
---> http://johnny-five.io/examples/motor-TB6612FNG/

Wiki from Sunfounder didn't help me much either.
---> http://wiki.sunfounder.cc/index.php?title=TB6612_DC_Motor_Driver_Module

I will be grateful if someone looks at it.

dtex:
Hi @5ocjal 

I haven't done anything with the Sunfounder, but I have done a lot of work with motors and motor controllers. First, looking at the image on the Sunfounder TB6612 wiki page I can see that it does not require 3 pins to drive it. It's not unheard of for a motor controller to require 3-pins and then for the breakout manufacturer to add hardware logic that controls one of those pins so you only need two pins from your microcontroller (or Pi in this case) to drive it. 

We just need to figure out which pins on the Raspberry Pi are connected to the MA, PWMA, MB and PWMB pins on the motor controller. Then use code like:

````js
var tb6612 = new Motors([
    { controller: "PCA9685", pins: [PWMA, MA] },
    { controller: "PCA9685", pins: [PWMB, MB] }
]);
````

I'm trying to digest the Picar library code so I can see how it's wired, but I'm neither a Pi guy or a Py guy so it might take me a little while.  Can you visually tell how it's wired?

dtex:
Note: it could turn out that PWMA and PWMB are connected to the PCA9685 and MA and MB are not connected to the PCA9685 (they don't need to be). If that's the case, this is going to get tricky.

[...]

PR #1625, opened 2019-09-17
Add support for PCA9685_Hybrid motors

dtex:
It is entirely reasonable to want to control a motor using a PCA9685 for speed and regular GPIO pins for direction. This PR allows for that.

Addresses #1624

### DIFF
--- a/lib/motor.js
+++ b/lib/motor.js
@@ -119,6 +119,36 @@ var Controllers = {
       }
     }
   },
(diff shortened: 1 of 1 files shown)


dtex:
Oh snap, this landed faster than I expected. Someone with hardware was testng for me. Can you roll back or should I just submit a patch PR?

PR #1626, opened 2019-09-18
Animation modernization with built in easing

rwaldron:
@dtex As I start to think about modernizing the hardware classes themselves, I've decided to do a "practice" round on Animation, because it's one of J5 components I'm least familiar with. Please review this rigorously!!

### DIFF
--- a/lib/animation.js
+++ b/lib/animation.js
@@ -1,15 +1,12 @@
+const Emitter = require("events");
+const ease = require("./easing");
+const { cloneDeep, constrain } = require("./fn");
+let temporal;
+
 // TODO list
 // Use functions as keyFrames
 // Test metronomic on real animation
-
 // Create jquery FX like queue
-
-var Emitter = require("events").EventEmitter;
-var util = require("util");
-var ease = require("ease-component");
-var Fn = require("./fn");
-var temporal;
-
 
 /**
  * The max time we want to allow a temporal animation segment to run.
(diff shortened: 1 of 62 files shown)


lib/easing.js
```diff
@@ -0,0 +1,109 @@
...
+
+Object.keys(ease).forEach(key => {
+  exportables[key.toLowerCase()] = ease[key];
+});
+
+module.exports = new Proxy({}, {
+  get(target, property) {
+    return exportables[property.replace(/([a-z])([A-Z])/g, "$1$2").toLowerCase()];
+  }
+});
```
rwaldron: I verified these functions by modifying the example.html page that comes with "ease-component" and compared the behavior side-by-side.

dtex: Cool. Is there an initiative to minimize dependencies generally speaking?

lib/animation.js
```diff
@@ -78,492 +50,499 @@ Animation.TemporalFallback.prototype.stop = function() {
...
-          };
         }
+        source[i] = keyFrame;
 
-        // Replace step values
-        if (typeof keyFrame.step !== "undefined") {
-          keyFrame.value = keyFrame.step === false ?
-            previousVal : previousVal + keyFrame.step;
-        }
+      }, this);
```
dtex: I think the "this" binding is unnecessary since we're using an arrow function on the opening forEach.

dtex:
This is exciting BTW!

lib/easing.js
rwaldron: Not specifically, I just happened to notice that this one comes from a project that was effectively abandoned. Easing functions aren't novel, so I figured we could just roll our own in there.

rwaldron:
> Were you just knocking out all the extended tests?

Oh, yes, that's correct. Sorry for any confusion there

rwaldron:
appveyor/pr failed because I canceled the build

PR #1632, opened 2019-10-06
Fix expander MCP23017 schema in fzz and png files

PeterArriaza:


dtex:
LGTM. Verified by http://ww1.microchip.com/downloads/en/DeviceDoc/20001952C.pdf

PR #1637, opened 2019-10-17
V2.0.0 alpha 1

rwaldron:
Testers, please install this with the following: 

```sh
npm install rwaldron/johnny-five#v2.0.0-alpha-1
```

Any issues are valid issues, but please do these things: 

- Use the `v2.0.0-alpha-x` label
- Include `gh-1637` in the title or `#1637` in the body

PR #1640, opened 2019-11-01
refactor(Board.js): refactor from es5 to es2015+ (WIP)

benhalverson:
changed var to const or let and switched to using arrow functions
https://github.com/rwaldron/johnny-five/issues/1596

One test is currently broken `this.emit is not a function`


rwaldron:
Thanks for the contribution, however this change has already been made in https://github.com/rwaldron/johnny-five/pull/1637/files#diff-54ffaa39b9e68a102065e0d7a79b6ea8

PR #1645, opened 2019-12-09
chore: Upgrade SerialPort to v8

reconbot:
SerialPort v8 brings support for the latest versions of node and electron. The api has had two major breaking changes. `.list()` no longer takes a callback and only returns a promise. And the `PortInfo` object now has `path` instead of `comName`.

This PR probably requires a similar upgrade to firmata https://github.com/firmata/firmata.js/pull/201

I haven't tested this PR on hardware yet.


eMarek:
I am curious if anyone had time to review this pull request?

PR #1653, opened 2020-01-10
Example annotation

ktbee:
@rwaldron here is a PR for some of the changes we've talked about during my examples auditing. I think we've either talked about all of them offline or their pretty self-explanatory, but let me know if you have any questions!

### DIFF
--- a/eg/servo-slider.js
+++ b/eg/servo-slider.js
@@ -6,7 +6,7 @@ board.on("ready", () => {
   const slider = new Sensor("A0");
   const tilt = new Servo(10);
 
-  slider.on("slide", () => {
+  slider.on("change", () => {
     tilt.to(slider.scaleTo(0, 180));
   });
 });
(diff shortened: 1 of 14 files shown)


eg/servo-slider.js
```diff
@@ -4,7 +4,7 @@ const board = new Board();
 board.on("ready", () => {
 
   const slider = new Sensor("A0");
-  const tilt = new Servo(9);
+  const tilt = new Servo(10);
```
ktbee: [The fritzing diagram for this example](https://github.com/rwaldron/johnny-five/blob/master/docs/servo-slider.md) uses pin 10, so this makes it consistent.

rwaldron:
@ktbee I was able to identify the code that was broken in `v2.0.0-alpha-1` and eliminated the problem by converting occurrences of `x ** y` to `Math.pow(x, y)`

rwaldron:
I've cherry picked all of the relevant commits from this PR/branch. Those changes are now in https://github.com/rwaldron/johnny-five/tree/v2.0.0-alpha-1

Issue #1658, opened 2020-02-14
Grove TH02 Humidity/Temperature sensor reporting incorrect (minus) values


MintyMods:
Hi,

Could someone provide me with some assistence as I seem to have hit a dead end with trying to get temperature and humidity reading via Johnny-Five.

I have tried numerous different sensors (TH02, BME280, DH11, DH22) and I can not get a valid reading out of any of them when used with Johnny-Five. I am sure this is something I am doing wrong but do not really know how to debug the issue further.

Bsaically running the following Johnny-five sample produces incorrect results for my testing with an Arduino Leonardo:-

```
 const I2C_GROVE_TEMP_HUMIDITY_ADDR = (0x40);  
 var humidity = new five.Multi({
    controller: "TH02"
    // address: I2C_GROVE_TEMP_HUMIDITY_ADDR,
  });

  humidity.on("change", function() {
    console.log("Thermometer");
    console.log("  celsius           : ", this.thermometer.celsius);
    console.log("  fahrenheit        : ", this.thermometer.fahrenheit);
    console.log("  kelvin            : ", this.thermometer.kelvin);
    console.log("--------------------------------------");
    console.log("Hygrometer");
    console.log("  relative humidity : ", this.hygrometer.relativeHumidity);
    console.log("--------------------------------------");
  });
```

The output I receive is as follows:-

```
1581616962327 Available COM7,COM1
1581616962332 Connected COM7
1581616967343 Repl Initialized
>> Thermometer
  celsius           :  -50
  fahrenheit        :  -58
  kelvin            :  223.15
--------------------------------------
Hygrometer
  relative humidity :  -24
--------------------------------------
```

The change event only fires once which I guess is because the reading are not changing (this I can confirm by using the `'ondata' `event rather than the `'onchange'`) and seeing the same incorrect readings being output.

The TH02 is connected via I2C as per the [diagram from the API](http://johnny-five.io/examples/multi-TH02/) docs. I have verified the I2C channel to be the default `(0x40)` as shipped by Grove and I have checked the J5 source code for IMU which shows it is also using address 0x40.

Checking the code against the [source code](https://github.com/Seeed-Studio/Grove_Temper_Humidity_TH02/blob/master/TH02_dev.cpp) from seed studio for the TH02 suggests it is not an issue with the Johnny-five TH02 IMU module as both contain the same formula etc. 

[...]

PR #1660, opened 2020-02-16
#1658 Fixes incorrect TH02 readings due to register readings

MintyMods:
As per the existing comments reading a single byte from `0x01`, and `0x2` was resolving to `[0,0]` so changed to read 3 bytes from `0x01` and discard the first byte. Also replaced calls to `uint16` with DataView to resolve lack of precision.

### DIFF
--- a/lib/imu.js
+++ b/lib/imu.js
@@ -2269,21 +2269,20 @@ var Drivers = {
             // 3 bytes from 0x01.
             Promise.all([
               new Promise(function(resolve) {
-                io.i2cReadOnce(address, 0x01, 1, function(data) {
-                  resolve(data[0]);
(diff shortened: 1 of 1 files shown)


MintyMods:
```
data.forEach(function (bytes, i) {
   view.setUint8(i, bytes);
```

Should be **byte** not bytes

```
data.forEach(function (byte, i) {
   view.setUint8(i, byte);
```

rwaldron:
This looks like it's now allocating `new DataView(new ArrayBuffer(...))` on every read cycle, just to read a single uint16 value. Is that really worth it?

MintyMods:
It does resolve the issue but I will take another look at that part.

Thank you for the feedback @rwaldron

PR #1661, opened 2020-02-28
Remove support for browser-serialport

djipco:
The API changes to the `serialport` module make it incompatible with the `browser-serialport` module. Until `browser-serialport` is updated to match the API of `serialport` (if ever), requiring it for NW.js no longer makes sense. In fact, leaving it there make it impossible to use Johnny-Five within NW.js.


djipco:
I created a PR that updates `browser-serialport` to align its API with version 8 of `serialport`. This will make this PR unnecessary.

PR #1666, opened 2020-03-16
Move to chrome-apps-serialport (instead of browser-serialport)

djipco:
With the changes in the API of `node-serialport` v8.x+, the `browser-serialport` is no longer usable inside Johnny-Five. For this reason, I created the `chrome-apps-serialport` module to revive support for Chrome Apps and, specifically, NW.js (which is, by definition, a Chrome App). 

This PR is pretty much harmless since `browser-serialport` is no longer working anyway. I also created a related [PR](https://github.com/firmata/firmata.js/pull/215) for Firmata.

By merging this PR, you allow users of Johnny-Five the ability to run their code inside NW.js without being forced to recompile the native `node-serialport` module, an operation which can prove to be quite complicated.

### DIFF
--- a/lib/board.js
+++ b/lib/board.js
@@ -29,8 +29,8 @@ var Serial = {
     var serialport;
 
     /* istanbul ignore next */
-    if (parseFloat(process.versions.nw) >= 0.13) {
-      serialport = require("browser-serialport");
+    if (chrome && chrome.serial) {
+      serialport = require("chrome-apps-serialport").SerialPort;
     } else {
       serialport = require("serialport");
     }
--- a/package.json
+++ b/package.json
@@ -158,7 +158,7 @@
     "temporal": "latest"
   },
   "optionalDependencies": {
-    "browser-serialport": "latest",
+    "chrome-apps-serialport": "^1.0.0",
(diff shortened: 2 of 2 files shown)


djipco:
As suggested by @monteslu [here](https://github.com/firmata/firmata.js/pull/215), there is an alternative way to use the `chrome-apps-serialport` module with Johnny-Five. The trick is to rely on `firmata-io` instead of `firmata.js`:

Here is an example: 

```js
const SerialPort = require("chrome-apps-serialport").SerialPort;
const Firmata = require("firmata-io")(SerialPort);
const five = require("johnny-five");

SerialPort.list().then(ports => {

  const device = ports.find(port => {
    return port.manufacturer && port.manufacturer.startsWith("Arduino")
  });

  const board = new five.Board({
    io: new Firmata(device.path)
  });

  board.on("ready", () => {
    console.log("Johnny-Five is ready!");
    const led = new five.Led(13);
    led.blink(500);
  });

});
```

So, I'm going to close this PR. However, we should still be removing `browser-serialport` from Johnny-Five's codebase because it no longer works.

PR #1668, opened 2020-03-18
BME-280 0x76 address: Update imu.js

schamberg97:
BME280 can also sometimes operate on 0x76. Not sure if it applies to other sensors within the family, so I haven't applied it to other chips

### DIFF
--- a/lib/imu.js
+++ b/lib/imu.js
@@ -1486,7 +1486,7 @@ var Drivers = {
 
   BME280: {
     ADDRESSES: {
-      value: [0x77]
+      value: [0x77, 0x76]
     },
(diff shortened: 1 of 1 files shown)


dtex:
Hi @schamberg97 ,

It's not clear to me where that 2nd element in the array would ever be used. To use 0x76 a user would just pass the address in the device options. We're not whitelisting addresses with this array, just providing a default.

rwaldron:
@dtex is right—you may simply pass that value here: 

```js
  let multi = new five.Multi({
    address: 0x76,
    controller: "BME280"
  });
```

PR #1679, opened 2020-04-30
feat(button): add a debounce time option

mattp94:
First, thanks a lot for this amazing library! 👏

In my project, I have buttons (with soft click) which trigger a lot of noise when pressing on it. I fixed the issue by increasing the debounce value of the button component. Would it be possible to expose an option to set this debounce value?

### DIFF
--- a/lib/button.js
+++ b/lib/button.js
@@ -183,15 +183,6 @@ function Button(opts) {
     last: null
   };
 
-  // Create a debounce boundary on event triggers
-  // this avoids button events firing on
-  // press noise and false positives
-  var trigger = Fn.debounce(function(key) {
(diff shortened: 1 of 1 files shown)


rwaldron:
If I land this on the current master, it won't be in 2.x branch. Will you be willing to write a patch for that branch as well?

rwaldron:
Can you update the base branch from `master` to `1.5.x`? Thanks!!

mattp94:
Hi @rwaldron, I changed the base branch to `1.5.x` and I created a [patch](https://github.com/rwaldron/johnny-five/pull/1690) for v2 🙂

PR #1711, opened 2020-10-28
Updated servo docs to reflect PWM problem

kula124p:
Added short warning in Servo docs about possible problems with PWM on pins 9 and 10. Hopefully this will save time...

### DIFF
--- a/docs/servo.md
+++ b/docs/servo.md
@@ -26,8 +26,9 @@ Run this example from the command line with:
 ```bash
 node eg/servo.js
 ```
-> **Warning:** Servo library may interfere with PWM on pins 9 and 10 on some boards like UNO even if pins are not used by Servo.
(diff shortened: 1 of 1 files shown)


ScreamZ:
@kula124 Are you sure about this ? Or did you missed to drop some pulldown resistance against interferences ?

dtex:
"On boards other than the Mega, use of the library disables analogWrite() (PWM) functionality on pins 9 and 10, whether or not there is a Servo on those pins." per [Arduino Servo docs](https://www.arduino.cc/reference/en/libraries/servo/)

ScreamZ:
Wow that's sick ! Thanks @rwaldron we should really merge this

kula124p:
It was a while ago so it's a bit foggy but I remember it not working. After changing the pin it was okay.
I've updated the message to include the reference to the official documentation mentioned by @dtex

PR #1732, opened 2021-06-25
Emitter: implement our own subclass of EventEmitter that includes a "pause" and "resume" method 

rwaldron:


dtex:
Holy smokes you're fast!

Two thoughts:

1) I don't think all emitters that might need to be suspendable are instances of withinable. This would create an inconsistent API across sensors.

2) I think the user who raised this will still have problems with typescript definitions. I'm not sure if that's just a warning in their dev environment or an error at build time, but it'll complain in some way (I think, I don't really do TS)

rwaldron:
> I don't think all emitters that might need to be suspendable are instances of withinable. This would create an inconsistent API across sensors.

This inconsistency already existed, but your right, we shouldn't compound it. Let me think about this a bit. 

> I think the user who raised this will still have problems with typescript definitions. I'm not sure if that's just a warning in their dev environment or an error at build time, but it'll complain in some way (I think, I don't really do TS)

```
@ts-ignore
```
?

rwaldron:
@dtex I'm going to invert it: `Withinable` becomes a subclass of `Suspendable`, `Suspendable` becomes a subclass of `Emitter`. Then we can make everything that's an `Emitter` a `Suspendable`.

dtex:
If you call it something more generic (j5Emitter) then later you could put other things in there like frequency/interval

rwaldron:
Or we can just call it `Emitter`, since the Node.js API is actually called `EventEmitter` 🤷🏻‍♂️

dtex:
That makes sense to me

rwaldron:
@dtex PTAL

lib/thermometer.js
```diff
@@ -880,7 +880,7 @@ Controllers.DEFAULT = Controllers.ANALOG;
 
 var priv = new Map();
 
-class Thermometer extends Withinable {
+class Thermometer extends Emitter {
```
dtex, line 883: I'm confused on this point. Thermometers should extend Withanble right?

dtex (changes requested):
Just that one point that I'm confused on. 

BTW, it took me a second to understand why Thermometer had `Withinable` and `eventEmitter` instead of `Withinable` and `Emitter` but I guess Driver's never need to `pause()`, right?

lib/thermometer.js
rwaldron: This is a mistake!

rwaldron: ...but I can't find this code 0_0

rwaldron: In my local it's `class Thermometer extends Withinable {`

rwaldron: Derp, I was on a different branch. Fix coming.

rwaldron:
@dtex you were not confused, I had accidentally introduced a bug there—nice catch!!

dtex (approved):
This looks good to me

PR #1782, opened 2022-03-25
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/lib/led/ledcontrol.js
+++ b/lib/led/ledcontrol.js
@@ -674,7 +674,7 @@ const Controllers = {
           throw new Error("The `row` method is only supported for Matrix devices");
         }
         if (typeof val === "number") {
-          val = (`0000000000000000${parseInt(val, 10).toString(2)}`).slice(-this.columns);
+          val = (`0000000000000000${parseInt(val, 10).toString(2)}`).slice(0 - (this.columns));
         }
         if (arguments.length === 2) {
           val = row;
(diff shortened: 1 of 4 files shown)


lib/led/ledcontrol.js
```diff
@@ -674,7 +674,7 @@ const Controllers = {
           throw new Error("The `row` method is only supported for Matrix devices");
         }
         if (typeof val === "number") {
-          val = (`0000000000000000${parseInt(val, 10).toString(2)}`).substr(0 - (this.columns), this.columns);
+          val = (`0000000000000000${parseInt(val, 10).toString(2)}`).slice(-this.columns);
```
dtex, line 677: I seem to recall that @rwaldron prefers the leading 0 before a negative variable for legibility, so I think we should bring it back here. I see this PR has been around for a while, so if you don't want to change it, then I can. 

Rick, please correct me if I remember it wrong.

dtex (changes requested):
Just one nit

dtex (approved):

PR #1799, opened 2022-11-01
Assign colors, and then assign colors to keys

fisher-alice:
@code-dot-org/johnny-five was recently updated to the latest release of its upstream in [this PR](https://github.com/code-dot-org/johnny-five/pull/12). 

@islemaster - While implementing Code.org customizations, I noticed a small bug in `lib/led/rgb.js` - `RGB.prototype[Animation.keys]` is assigned to `RGB.colors` before `RGB.colors` is assigned to an array of strings. This PR changes the order of these lines.

[jira](https://codedotorg.atlassian.net/browse/SL-358)

PR #1803, opened 2022-12-02
Add pulse method to RGB class

fisher-alice:
Currently, the [Led class](https://github.com/rwaldron/johnny-five/blob/main/lib/led/led.js) contains both pulse and blink methods, but the [Rgb class](https://github.com/rwaldron/johnny-five/blob/main/lib/led/rgb.js) only contains the blink method. 

This PR adds a `pulse` method to the Rgb class. Note that the pulse method was recently [re-added](https://github.com/code-dot-org/johnny-five/pull/15) to the forked branch [`code-dot-org/johnny-five`](https://github.com/code-dot-org/johnny-five) after being updated to the latest release of upstream. The `pulse` method was first added to the Rgb class by @islemaster in [this PR](https://github.com/code-dot-org/johnny-five/pull/10).

I added the state value `isRunning` which was already a defined property in order to keep track of whether an Animation is in progress, i.e., if `pulse` is being run.  This is similar to how `isRunning` is used in `lib/led.js`.

The update method is now writeable to be able to reassign it in the unit test for Animation.render in [`test/rgb.js`](https://github.com/rwaldron/johnny-five/blob/87d2adabdf2b0c86c7e2b43c7034992171fe3dc1/test/rgb.js#L820).

More details of implementation are included in embedded comments.

## Testing
I modified unit tests in both `rgb.collection.js` and `rgb.js` as well as added new unit tests for the RGB pulse method.

## Documentation
At the request of @dtex, I added the example file `docs/led-rgb-pulse.md` and here is [documentation](https://github.com/fisher-alice/johnny-five-wiki/pull/1) about the Rgb pulse method that can be added to the johnny-five wiki.

### DIFF
--- a/lib/led/rgb.js
+++ b/lib/led/rgb.js
@@ -207,6 +207,7 @@ class RGB {
         }
       },
       update: {
+        writable: true,
         value(colors) {
           const state = priv.get(this);
 
@@ -478,7 +479,9 @@ class RGB {
    */
 
   [Animation.render](frames) {
-    return this.update(frames[0]);
+    const state = priv.get(this);
+    state.value = frames[0];
+    return this.update();
   }
 
 
--- a/lib/led/rgb.js
+++ b/lib/led/rgb.js
@@ -282,6 +282,11 @@ class RGB {
       };
 
       this.update(colors);
+    } else if (state.isRunning) {
+      this.stop();
+      colors = state.prev;
+      this.update(colors);
+      
     }
 
     return this;
@@ -481,7 +486,7 @@ class RGB {
   [Animation.render](frames) {
     const state = priv.get(this);
     state.value = frames[0];
-    return this.update();
+    return this.update(state.value);
   }
 
 
--- a/lib/led/rgb.js
+++ b/lib/led/rgb.js
@@ -172,14 +172,14 @@ class RGB {
       intensity: 100,
       isAnode: options.isAnode || false,
       interval: null,
-      // isRunning is used to tracks whether an Animation is in progress similar to how it's used in led.js
+      // isRunning is used to track whether an Animation is in progress similar to how it's used in led.js
       isRunning: false,
       // red, green, and blue store the raw color set via .color()
       // values takes state into account, such as on/off and intensity
       values: {
         red: 255,
         green: 255,
-        blue: 255,
+        blue: 255
       }
     };
 
@@ -261,7 +261,7 @@ class RGB {
 
     this.update(update);
 
-    // store colors to state.prev for future use by on() or pulse()
+    // store colors to state.prev for future use by on()
     state.prev = update;
 
     return this;
@@ -287,13 +287,9 @@ class RGB {
   }
 
   off() {
-    const state = priv.get(this);
-
     // If it's already off, do nothing so the previous state stays intact
     /* istanbul ignore else */
     if (this.isOn) {
-      state.prev = RGB.colors.reduce((current, color) => (current[color] = state[color], current), {});
-
       this.update({
         red: 0,
         green: 0,
@@ -338,6 +334,8 @@ class RGB {
   pulse(duration, callback) {
     const state = priv.get(this);
     var currentColor = state.prev;
+
+    // Avoid traffic jams
     this.stop();
 
     const options = {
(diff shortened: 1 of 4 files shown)


lib/led/rgb.js
```diff
@@ -275,22 +281,16 @@ class RGB {
...
 
   off() {
-    const state = priv.get(this);
-
-    // If it's already off, do nothing so the pervious state stays intact
+    // If it's already off, do nothing so the previous state stays intact
     /* istanbul ignore else */
     if (this.isOn) {
-      state.prev = RGB.colors.reduce((current, color) => (current[color] = state[color], current), {});
-
```
fisher-alice, line 293: Within the `off()` method, there is no need to reassign `state.prev` if the RGB is on since `state.prev` is assigned when the `on()` method is called. This also ensures that if `off()` is called when the RGB is pulsing that the RGB values are NOT assigned lesser and lesser values until reaching (0,0,0) which is the case when `off()` is repeatedly called after pulse() is called if this line remains.

lib/led/rgb.js
```diff
@@ -429,19 +479,22 @@ class RGB {
...
-    return this.color(frames[0]);
+    const state = priv.get(this);
+    state.value = frames[0];
+    return this.update();
   }
 
 
 }
+
+RGB.colors = ["red", "green", "blue"];
```
fisher-alice, line 490: The assignment of RGB.colors was moved before Animation.keys is assigned to it. [This small PR](https://github.com/rwaldron/johnny-five/pull/1799) resolves this issue and adds a unit test for this fix.

lib/led/rgb.js
```diff
@@ -429,19 +479,22 @@ class RGB {
    */
 
   [Animation.render](frames) {
-    return this.color(frames[0]);
+    const state = priv.get(this);
+    state.value = frames[0];
+    return this.update();
```
fisher-alice, line 484: The `Animation.render` method here was calling `color()`, which does a lot of normalization and validation before itself calling `update` on line 270 above. 
Now the `render` method behaves more like `render` in the [`led class`](https://github.com/rwaldron/johnny-five/blob/main/lib/led/led.js#L334) which 'trusts' its animations and calls `update` directly .
This also bypasses the reassignment of `state.prev` to a 'tween' value when `Animation.render` is called for `pulse` function which resulted in the RGB values decreasing so that the RGB light would gradually decrease.

lib/led/rgb.js
dtex: #1799  has been merged.

dtex (changes requested):
Thank you, @fisher-alice; This is an excellent addition to the RGB class.

Everything looks good, and I think the only thing that is missing is documentation.  The wiki needs the RGB.pulse() method, and it warrants an example file too.

Can you add those things?

lib/led/rgb.js
fisher-alice: Thanks!

lib/led/rgb.js
```diff
@@ -275,22 +281,21 @@ class RGB {
         blue: 255
       };
 
-      state.prev = null;
-
       this.update(colors);
+    } else if (state.isRunning) {
```
fisher-alice, line 285: Added this else-if branch for case when `on()` is called while RGB is pulsing so that RGB displays full RGB values and not 'tween' values of `pulse` method.

fisher-alice:
> Thank you, @fisher-alice; This is an excellent addition to the RGB class.
> 
> Everything looks good, and I think the only thing that is missing is documentation. The wiki needs the RGB.pulse() method, and it warrants an example file too.
> 
> Can you add those things?

Hello @dtex , thanks for taking time to look at my PR! Yes, I can definitely add documentation to the wiki and create an example file. FYI - I also added a commit since your comment.
I see you're based in Houston as well...go Coogs!

dtex:
Oh, you're in Houston too?! Cool! I haven't been getting out much lately (for like almost three years). Have you been to any local meetups or groups on IoT or robtics stuff? I see Houston Robotics Club has been re-animated but I haven't checked it out yet.

There are a few PR's here that I think are ready. I'll ping Rick and see if I can get them merged.

Issue #1789, opened 2022-06-16
TypeError: serialport.list is not a function (board.js line 42)


manno-xx:
I'm trying to connect my macbook (year 2018, running monterey) to an arduino. When running the standard blink example as a test, I get the error:

`
/Users/manno/node_modules/johnny-five/lib/board.js:42
    serialport.list().then(results => {
               ^

TypeError: serialport.list is not a function
    at Board.detect (/Users/manno/node_modules/johnny-five/lib/board.js:42:16)
    at new Board (/Users/manno/node_modules/johnny-five/lib/board.js:291:23)
    at Object.<anonymous> (/Users/manno/[some more path]/NodeWebSocketServer/withFirmata.js:2:13)
`

Which points to the one-before-last line in:
`
let serialport = serialport;

    /* istanbul ignore next */
    if (parseFloat(process.versions.nw) >= 0.13) {
      serialport = require("browser-serialport");
    } else {
      serialport = require("serialport");
    }

    // console.log(require);
    // Request a list of available ports, from
    // the result set, filter for valid paths
    // via known path pattern match.
    serialport.list().then(results => {
      const portPaths = results.reduce((accum, result) => {

`

npm list gives me:

├── johnny-five@2.1.0
├── serialport@10.4.0
└── ws@8.7.0

Anyone an idea how to solve this?

Kind regards,
Manno

manno-xx:
Yes, sorry :)

It is just the standard blink from the homepage:

```
var five = require("johnny-five");
var board = new five.Board();

board.on("ready", function(){
    var led = new five.Led(13);
    led.blink(500);
});
```

manno-xx:
Btw: I removed some stuff to make sure there were no wrong libraries referenced or something. Then I ran the following (with the same result):

```
manno@Manno NodeWebSocketServer % node withFirmata.js 
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module '@serialport/bindings'
Require stack:
- /Users/manno/node_modules/johnny-five/node_modules/serialport/lib/index.js
- /Users/manno/node_modules/johnny-five/lib/board.js
- /Users/manno/node_modules/johnny-five/lib/accelerometer.js
- /Users/manno/node_modules/johnny-five/lib/johnny-five.js
- /Users/manno/Library/CloudStorage/OneDrive-HanzehogeschoolGroningen/P0.0 Development/Web Dev/Server/NodeWebSocketServer/withFirmata.js
[...]

Issue #1786, opened 2022-04-19
Cannot find module 'firmata' - works only at manual copy to node_modules


Amaranthusss:
Hello there,

It looks like there is a some bug at johnny-five library's dependencies (I am not sure, please verify).

A long time ago I prepared simple NodeJS script to test NodeMcu V3 board and controlling it with the firmata firmware. And that worked but this time I could not install correctly Johnny-Five. So I made a new project with only 3 dependencies for a test:
`
    "etherport-client": "^0.1.4",
    "johnny-five": "1.5.0",
    "keypress": "^0.2.1",
`
And when I start the simple script shown below I got error.
Script: 
`
const { EtherPortClient } = require("etherport-client")
const five = require("johnny-five")
const keypress = require("keypress")

const board = new five.Board({
	port: new EtherPortClient({
		host: "192.168.8.122",
		port: 3030,
	}),
	repl: false,
})
`
Error:
`
node:internal/modules/cjs/loader:936
  throw err;
  ^

Error: Cannot find module 'firmata'
Require stack:
- C:\Users\Amaranthuss\Desktop\firmata-fix\node_modules\johnny-five\lib\board.js
- C:\Users\Amaranthuss\Desktop\firmata-fix\node_modules\johnny-five\lib\accelerometer.js
- C:\Users\Amaranthuss\Desktop\firmata-fix\node_modules\johnny-five\lib\johnny-five.js
- C:\Users\Amaranthuss\Desktop\firmata-fix\index.js
`

I tried to install firmata but it is not possible.
`
C:\Users\Amaranthuss\Desktop\firmata-fix>npm i firmata
npm ERR! code 1
npm ERR! path C:\Users\Amaranthuss\Desktop\firmata-fix\node_modules\@serialport\bindings
npm ERR! command failed
npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c prebuild-install --tag-prefix @serialport/bindings@ || node-gyp rebuild
npm ERR! gyp info it worked if it ends with ok
npm ERR! gyp info using node-gyp@8.4.1
npm ERR! gyp info using node@16.14.2 | win32 | x64
npm ERR! gyp info find Python using Python version 3.10.4 found at "C:\Python310\python.exe"
npm ERR! gyp ERR! find VS
npm ERR! gyp ERR! find VS msvs_version not set from command line or npm config
npm ERR! gyp ERR! find VS VCINSTALLDIR not set, not running in VS Command Prompt
[...]

PR #1808, opened 2023-01-20
Update serialport

### LINKED ISSUE #1780: Old serialport (< 10.x.x) usage
I believe that an old version of serialport is being used. The current serialport (10.3.0) doesn't make use of @serialport/bindings, instead it uses @serialport/bindings-cpp. I'd love to use the current serialport version as it's compatible with the latest Electron version and I'm not forced to use Electron v12 or below.

Is there any way I could help updating this package so that this can be solved?

scottgonzalez:
Fixes #1780

This update seemed too simple based on how long this issue has been open, but it worked for me with a few real world tests.


dtex:
Hi @scottgonzalez I hope all is well. Do you happen to know if the Serialport issues in #1780 were limited to Windows? If so, that might explain the delay here. Also, @reconbot (sorry to rope you in) does the change in Scott's first commit seem like the right thing to do? If so, I can help with getting this merged, but I'm not sure I can publish it to npm. I'll check.

scottgonzalez:
As far as I can tell, it was sort of limited to Windows, but not really. There's also #1789 which is effectively the same bug. But there's also #1786, which is related, and I believe that one is specific to Windows.

Here's what I've gathered so far:
* Johnny-Five lists `serialport` and `firmata` as optional dependencies, so you're not guaranteed to get the expected versions within any given app.
* On OS X, installing `johnny-five` does install both `serialport` and `firmata` with expected versions and everything works fine.
* On Windows, I *think* `serialport` is automatically installed, but doesn't work, so people install `serialport` directly and end up with a newer version, which causes them to run into the `serialport.list is not a function` bug.

What I know for certain is that I've never had any issues on OS X, but I've had multiple people on Windows report bugs when trying to use my projects that rely on Johnny-Five. Based on the comments in that issue, I've told the Windows users to downgrade to Node 16 and that has helped.

scottgonzalez:
I had someone test on Windows with a fresh clone of one of my repos. They got the same results using Node 16 and 18, so I'm not sure what steps the other people are taking to end up with a broken serialport.

```
$ npm ls serialport
blue-owl@0.0.0 C:\Dev\git\blue-owl-new
└─┬ johnny-five@2.1.0
  ├─┬ firmata@2.3.0
  │ └── serialport@8.0.8 deduped
  └── serialport@8.0.8
```

scottgonzalez:
Another update: Apparently you can install Node with or without the Windows build tools. Without the build tools, `serialport` won't install correctly. So that must be why people are explicitly adding it as a direct dependency. That seems to match up with the comments in #1786.

dtex:
Oh man, thank you for the detective work on this. I've wondered for a long time why Serialport "just worked" for some Windows users and not others. The only catch I see is that normally the `/docs` files are all [auto-generated](https://github.com/rwaldron/johnny-five/blob/main/CONTRIBUTING.md#writing-docs) from the `/eg` files with a call to `grunt examples` so the copyright date changes will get lost the next time that gets run.

scottgonzalez:
The example updates were a result of running `grunt examples`.

scottgonzalez:
The funny thing is that I only ran `grunt examples` because my fork was super old and still running Travis and the [Travis build failed](https://github.com/scottgonzalez/johnny-five/runs/10780555922) because the generated docs didn't match the expected outcome due to the copyright updates. Since it seems like that logic has changed, I'll revert that commit to keep this cleaner.

scottgonzalez:
Thanks @dtex. Were you able to find out about publishing to npm? Looks like only @rwaldron has access right now.

dtex:
Rick knows I'm merging this and a few other PR's so he's expecting me to hit him up for a publish soon.

Issue #1834, opened 2023-10-10
Unable to use Led.Digits with Multple Boards


chrisl8:
I am running Johnny-Five on a Raspberry Pi connected to two Arduino Mega 3's using the Firmata.

LEDs, Buttons, Potentiometers and switches all work fine following the instructions here:
https://johnny-five.io/examples/board-multi/

However, HT16K33 `Led.Digits` do not. I can set up multiple boards, but only the first `Led.Digits` instance that I add will work, the others remain blank/off.

I am following the instructions here:
https://johnny-five.io/examples/led-digits-clock-HT16K33/
and then adding the  `board` option, and I can use EITHER board, but if I try to add both, only the first one added works.

Here is the code I am using to test:
```
const { Boards, Led } = require("johnny-five");

const displays = [
  {
    id: "one",
    port: "/dev/ttyACM3",
    repl: false,
  },
  {
    id: "two",
    port: "/dev/ttyACM0",
    repl: false,
  },
];

const boards = new Boards(displays);

boards.on("ready", () => {
  // This is only here to demonstrate that both boards work
  const boardOneLED = new Led({ board: boards.byId("one"), pin: 13 });
  boardOneLED.blink(500);

  // This is only here to demonstrate that both boards work
  const boardTwoLED = new Led({ board: boards.byId("two"), pin: 13 });
  boardTwoLED.blink(500);

  // Only the first one will work, the second one added will not.
  // Reverse their order to demonstrate.
  const boardOneDigits = new Led.Digits({
    controller: "HT16K33",
    board: boards.byId("one"),
  });
  const boardTwoDigits = new Led.Digits({
    controller: "HT16K33",
    board: boards.byId("two"),
  });

  boardOneDigits.on();
  boardOneDigits.print("1111");
  boardTwoDigits.on();
  boardTwoDigits.print("2222");

  process.on("SIGINT", () => {
    console.log("Turning off displays.");
    boardOneDigits.off();
    boardTwoDigits.off();
  });
});
```

dtex:
This is interesting. Here's what I think is happening...

The HT16K33 is an addressable I2C device. Both of yours start at 0x70, which I believe is the default value. If no address or array of addresses passed, a built-in list of available addresses is used. That list is not built with consideration of multi-board environments, so 0x70 gets shifted off for the first one instantiated and J5 thinks the second one is at the next available address.

I think if you explicitly pass the address or addresses, it will not use the built-in list.

```js
[...]

PR #1835, opened 2023-10-15
Allow bypassing address validation in HT16K33 code to address Issue #1834

chrisl8:
Allow adding an option called `skipAddressValidation` when calling the `Led.Digits` function to skip validating the given address and deleting them.

This will address Issue #1834 

This allows using multiple HT16K33 devices on different external boards that might all be on the same address.

This is the only option I could come up with that I feel confident won't affect any existing code that may depend on the current functionality.

I've also added documentation for the option in the place that made the most sense to me.

Please let me know if you feel this option is reasonable, and if so, if I need to add any tests for this before acceptance.

### DIFF
--- a/docs/led-digits-clock-HT16K33.md
+++ b/docs/led-digits-clock-HT16K33.md
@@ -60,6 +60,39 @@ function time(showColon) {
 
 
 
+## Additional Options
+### Specifying Addresses
+You can specify the addresses to use for the controller:
+```
+  const digits = new Led.Digits({
+    controller: "HT16K33",
+    address: 0x70,
+  });
+```
+
+### Bypassing Address Validation
+Led.Digits will ensure that you do not use the same address twice,  
+but if you use multiple Arduino boards you want to reuse the same address.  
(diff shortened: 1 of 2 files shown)


dtex:
My only concern is that this doesn't solve things for the auto-addressing of multiple HT16K33s across multiple boards. 

What if we stored addresses in a WeakMap with boards as keys? It wouldn't require changes to the API, would still protect addresses from reuse, and support multiple boards.

chrisl8:
> My only concern is that this doesn't solve things for the auto-addressing of multiple HT16K33s across multiple boards.

I was just taking the "easy way out" with the assumption that I'm the only person on the planet even doing this, but my way definitely leaves users like me without that feature.

That feature seems a little odd to me, as I would expect anyone using multiple HT16K33s to already know their addresses, and want to address each one specifically rather than letting the code "guess" for them.

> What if we stored addresses in a WeakMap with boards as keys? It wouldn't require changes to the API, would still protect addresses from reuse, and support multiple boards.

That sounds like a good idea to me. I can try to revisit this when I have free time again.

